PR #10, opened 2026-01-09
fix: prevent session termination on client detach

### LINKED ISSUE #9: detach is killing entire session
`^b-d` which should be "detach" ends up killing the session, and then `psmux ls` doesn't list any active sessions.

thoraxe:
## Summary
- Add Windows console control handler to prevent session termination when client detaches
- Install `SetConsoleCtrlHandler` in the server process to ignore `CTRL_CLOSE_EVENT`, `CTRL_LOGOFF_EVENT`, and `CTRL_SHUTDOWN_EVENT`

## Problem
When pressing `Ctrl+B d` to detach from a psmux session, the entire session (server + shell processes) was being killed instead of just detaching the client. After detaching, `psmux ls` showed no active sessions.

## Root Cause
On Windows, when the client process exits after detaching, console termination events can propagate to the server and its child processes through the console subsystem. While the server is spawned with `CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW`, this isn't sufficient to isolate the shell processes from console events.

## Solution
Add a `SetConsoleCtrlHandler` in the server process that returns `TRUE` for close/logoff/shutdown events, preventing the default termination behavior. This is a common pattern for Windows services/daemons.

## Test plan
- [x] Build with `cargo build --release`
- [x] Start a session: `psmux`
- [x] Run some commands in the shell
- [x] Detach with `Ctrl+B d`
- [x] Verify session exists: `psmux ls` shows the session as "(detached)"
- [x] Re-attach: `psmux attach` connects to the existing session with shell state preserved

🤖 Generated with [Claude Code](https://claude.ai/code)

Closes #9

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -23,6 +23,41 @@ use std::env;
 use crossterm::style::Print;
 use serde::{Serialize, Deserialize};
 
+/// Install a console control handler on Windows to prevent termination on client detach.
+/// When the psmux client exits (after Ctrl+B d), Windows console events may propagate
+/// to the server process. This handler ignores CTRL_CLOSE_EVENT and related events
+/// to keep the server running.
+#[cfg(windows)]
+fn install_console_ctrl_handler() {
+    type HandlerRoutine = unsafe extern "system" fn(u32) -> i32;
+
+    #[link(name = "kernel32")]
+    extern "system" {
+        fn SetConsoleCtrlHandler(handler: Option<HandlerRoutine>, add: i32) -> i32;
+    }
+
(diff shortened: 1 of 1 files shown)


lostmsu:
Just to be clear, does this also work when the client simply dies for any reason?

psmux:
Thanks @thoraxe! I've integrated this fix into master in commit 2c53984. The console control handler approach works great - sessions now persist after detach.

Closing this PR as the fix is now in master. Thanks for identifying this solution! 🙏

psmux:
Integrated into master - thanks!

PR #13, opened 2026-01-10
fix: replace TCP with Windows named pipes for security

### LINKED ISSUE #5: Security issue: any process can connect over TCP, including running under different users
The app should use named pipes with appropriate permissions instead.

thoraxe:
## Summary
- Fixes #5: Replace TCP sockets with Windows Named Pipes to prevent unauthorized access
- Add security descriptors that restrict pipe access to the current user only
- Replace `.port` files with `~/.psmux/sessions` registry file for session discovery

## Security Issue
Previously, any process on the system could connect to psmux over TCP by reading the port file at `~/.psmux/{session}.port`. This allowed:
- **Data exfiltration** via `capture-pane`
- **Command injection** via `send-text`

## Solution
Named pipes with user-only DACL security descriptors now prevent other users from connecting. Attempts by other users result in `ERROR_ACCESS_DENIED`.

## Test plan
- [ ] Start a psmux session and verify it works normally
- [ ] Attach/detach from another terminal
- [ ] Verify `psmux ls` lists sessions correctly
- [ ] Verify session cleanup on exit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,7 +299,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core",
+ "windows-core 0.62.2",
 ]
 
 [[package]]
@@ -530,6 +530,7 @@ dependencies = [
  "unicode-width 0.1.14",
  "vt100",
  "which",
+ "windows",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


thoraxe:
I have not tested myself yet.

psmux:
Thanks @thoraxe for tackling the security issue! Named pipes with proper DACLs is the right solution.

This is a significant change that affects all inter-process communication. I want to make sure we test thoroughly:

**Questions:**
1. How does session discovery work now? The sessions registry file approach looks cleaner than individual .port files.
2. Have you tested with multiple concurrent sessions?
3. What happens if the sessions registry gets corrupted?

**Note:** I've already addressed issue #9 (detach killing session) via console control handler in 2c53984, so that part is resolved.

I'll do some testing on my end. Could you confirm your test results when you get a chance?

psmux:
This PR's security approach (named pipes with security descriptors) has been superseded by the session key authentication feature that was implemented in commit 1b03135.

The current implementation:
- Generates a unique session key on server startup
- Requires AUTH <session_key> as first command from clients
- Stores session key in ~/.psmux/session_key (user-readable only)
- Rejects unauthorized connections before processing any commands

Thank you for raising the security concern. This PR can be closed as the security issue has been addressed with a different approach.

PR #31, opened 2026-02-12
Remove duplicate Cargo installation section from README

inayayousfi:
I removed the duplicate of the readme. Dumb but it bothered me. And I removed the first one because I think people will prefer to use their package manager rather than cargo itself.


psmux:
Thanks @ZiedYousfi Hope you like the new Readme. Feel free to open a new issue.Any comments/feedback welcome! Thank you for your involvement!

PR #75, opened 2026-03-03
feat(mouse): forward hover (Moved) events to child PTY

john-mutuma:
## What

Mouse hover (`MouseEventKind::Moved`) events are received from the terminal but silently discarded in all three handling paths (standalone, client-server, and app dispatch).

## Why

TUI apps running inside psmux panes (htop, vim, opencode, etc.) expect SGR mouse motion events (button 35) for hover functionality. Without forwarding these events, hover-dependent features in child applications don't work.

## Fix

Forward `MouseEventKind::Moved` as SGR button 35 (`\x1b[<35;col;rowM`) to the child PTY, gated on `alternate_screen()` to avoid PSReadLine false positives at shell prompts. Changes span 3 files:

- **`src/window_ops.rs`** — Replaced `remote_mouse_motion()` no-op stub with real forwarding logic
- **`src/input.rs`** — Handle `Moved` in standalone mode
- **`src/client.rs`** — Handle `Moved` in client-server mode (sends `mouse-move` to server only when active pane is in alt screen)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1438,7 +1438,18 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                             MouseEventKind::Up(MouseButton::Right) => {}
                             MouseEventKind::Up(MouseButton::Middle) => {}
                             MouseEventKind::Moved => {
-                                // Don't send bare mouse-move to server - wasteful and server ignores it
+                                // Forward bare mouse motion (hover) to server only when the
+                                // active pane is in alternate screen mode (real TUI app like
+                                // htop, vim, opencode).  At a shell prompt, hover is not sent
+                                // to avoid wasteful socket traffic.
+                                let tui_active = if !prev_dump_buf.is_empty() {
+                                    serde_json::from_str::<DumpState>(&prev_dump_buf)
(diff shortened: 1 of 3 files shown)


psmux:
Hey @john-mutuma, thanks for this PR! Your diagnosis is exactly right -- `MouseEventKind::Moved` was being discarded in all three paths, and forwarding as SGR button 35 gated on `alternate_screen()` is the correct approach.

I studied Windows Terminal's source code for how they handle this (`mouseInput.cpp` -> `_windowsButtonToSGREncoding` + the tracking mode checks in `HandleMouse`), and your approach aligns perfectly with how WT does it.

I've pushed a fix in 606ddfc that follows the same pattern with one addition: same-coordinate deduplication. WT suppresses consecutive `MOUSE_MOVED` events at the same (x, y) position to avoid flooding the PTY pipe. I added a `last_hover_pos` field in `AppState` for this.

Since the fix is already on master with the dedup improvement, I won't merge this PR directly, but I want to be clear: your analysis and approach were correct and very helpful. The videos you posted on the issue showing the hover behavior difference made the problem immediately obvious.

Thank you for taking the time to dig into this and put together a clean PR. Really appreciate it!

john-mutuma:
Tested with recent main, seems to be working fine! I'll close the PR with this.

PR #79, opened 2026-03-04
fix(mouse): gate click/drag forwarding and align hover test

XingLingQAQ:
## Summary

  This PR fixes mouse escape-sequence leakage into shell-like prompts (including ClaudeCode input) by unifying mouse
  forwarding gates.

  - Gate mouse forwarding behind `pane_wants_mouse(...)` on both local and remote paths.
  - Cover left/middle/right click press/release, left drag, and hover motion.
  - Keep shell behavior intact (e.g. right-click paste fallback when pane does not want mouse).
  - Update `tests/test_mouse_hover.ps1` to assert the new unified gate (`pane_wants_mouse`) instead of
  `is_fullscreen_tui`.

  ## Root cause

  Mouse events were still being encoded and injected as SGR sequences (e.g. `\x1b[<0;...M/m`, `\x1b[<35;...M`) in paths
  where the active pane did not actually request mouse protocol.
  In shell-like prompts, these bytes could surface as visible garbage text.

  ## Files changed

  - `src/input.rs:1376-1603`
  - `src/window_ops.rs:403-620`
  - `tests/test_mouse_hover.ps1:135-154`

  ## Validation

  - [x] `cargo build`
  - [x] `pwsh -File tests/test_mouse_hover.ps1` (14 passed, 0 failed)
  - [x] Manual check: ClaudeCode input no longer shows `[<0;...` / `[<35;...` garbage on click/hover

  ## Risk / Trade-off

  Applications that rely on mouse behavior without properly enabling mouse protocol may no longer receive forwarded
  mouse events. This is an intentional tightening to prevent escape-sequence leakage in non-mouse-aware prompts.

PR #84, opened 2026-03-05
feat(server): add configurable detach and empty exit policies

XingLingQAQ:
## Summary

  This PR adds two tmux-style server options to make server shutdown behavior configurable while
  preserving backward compatibility:

  - `destroy-unattached` (default: `off`)
    - Controls whether the server exits when no clients are attached.
  - `exit-empty` (default: `on`)
    - Controls whether the server exits when all panes/windows become empty.

  ## What changed

  ### 1) App state defaults
  - Added:
    - `destroy_unattached: bool` (default `false`)
    - `exit_empty: bool` (default `true`)

  ### 2) Option chain integration (set/show/unset/config/format)
  - Added support in:
    - `set-option` parsing and apply path
    - `show-options` / `show-options -v`
    - `set -u` fallback defaults
    - `source-file` / config parser
    - format lookup (`#{...}` option query path)

  ### 3) Lifecycle behavior wiring
  - `all_empty` auto-exit is now gated by `exit-empty`.
  - Added detach-time `destroy-unattached` check in server loop.
  - Connection layer now compensates detach on abrupt EOF/error for attached persistent clients to
  avoid attached-client count drift.

  ### 4) Tests
  - Extended `tests/test_keybinding_options.ps1` with:
    - option visibility checks
    - set/show checks
    - unset default fallback checks
    - source-file application checks

  ## Backward compatibility

  Defaults preserve previous behavior:
  - `destroy-unattached=off`
  - `exit-empty=on`

  ## Validation

  - `cargo build`
  - `pwsh -File tests/test_keybinding_options.ps1`
    - Result: **46 passed, 0 failed**

  ## Risk / Trade-off

  - Client attach/detach accounting remains count-based; this PR reduces common drift risk
  (EOF/error paths) but still depends on valid attach/detach command flow

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -511,6 +511,12 @@ pub fn parse_option_value(app: &mut AppState, rest: &str, _is_global: bool) {
         "remain-on-exit" => {
             app.remain_on_exit = matches!(value, "on" | "true" | "1");
         }
+        "destroy-unattached" => {
+            app.destroy_unattached = matches!(value, "on" | "true" | "1");
+        }
+        "exit-empty" => {
+            app.exit_empty = matches!(value, "on" | "true" | "1");
+        }
         "aggressive-resize" => {
             app.aggressive_resize = matches!(value, "on" | "true" | "1");
         }
--- a/src/format.rs
+++ b/src/format.rs
@@ -785,6 +785,8 @@ fn lookup_option(name: &str, app: &AppState) -> Option<String> {
         "automatic-rename" => Some(if app.automatic_rename { "on".into() } else { "off".into() }),
         "monitor-activity" => Some(if app.monitor_activity { "on".into() } else { "off".into() }),
         "remain-on-exit" => Some(if app.remain_on_exit { "on".into() } else { "off".into() }),
+        "destroy-unattached" => Some(if app.destroy_unattached { "on".into() } else { "off".into() }),
+        "exit-empty" => Some(if app.exit_empty { "on".into() } else { "off".into() }),
         "set-titles" => Some(if app.set_titles { "on".into() } else { "off".into() }),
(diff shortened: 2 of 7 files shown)


psmux:
Hey @XingLingQAQ! 👋

Just merged this - really nice work! 🎉

I reviewed the entire PR in detail and it's clean, well-structured, and follows the existing patterns perfectly. A few things I particularly liked:

### What stood out

- **The `attached_sent` tracking in `connection.rs`** - this is a great improvement beyond just the new options. The EOF/error paths now properly send `ClientDetach`, which prevents the attached-client count from drifting when connections drop unexpectedly. That's a subtle but important fix.

- **The safety check in `destroy-unattached`** - only exiting when there are no live panes is the right call. It avoids the sharp edge where `destroy-unattached=on` could prematurely kill sessions that `exit-empty=off` was supposed to keep alive.

- **Backward compatibility** - defaults preserving existing behavior (`destroy-unattached=off`, `exit-empty=on`) means zero breakage for existing users. Exactly how it should be done.

- **Tests** - 6 comprehensive test cases covering set/show, unset defaults, and source-file application. Thorough.

### One small thing I cleaned up during the merge

I translated the Chinese comments in `server/mod.rs` to English for codebase consistency - the logic is identical, just the comment language changed. No functional difference.

There was a small merge conflict with recent changes (I added 250ms reap throttling), but the resolution was straightforward - kept the throttling from master and your `exit_empty` gate.

---

Thanks for another solid contribution! Between PR #79 (mouse fix) and now this, you've been making psmux better for everyone. Really appreciate it. 🙏

If you know anyone who might find psmux useful, consider sharing the project with them - every bit of visibility helps the community grow!

PR #96, opened 2026-03-09
fix: dialogs hidden when status-position is top

cmaughan:
When `status-position top` is configured, the layout splits the terminal as [status_bar, content] so chunks[0] is the 1-line status bar area. All overlay dialogs (rename window, set pane title, command, confirm) were hardcoded to render inside chunks[0], causing them to be squashed and hidden behind the status bar.

Fix: use content_chunk (already computed from the layout to always reference the pane content area) for all four dialog centered_rect calls, so dialogs render correctly regardless of status bar position.


cmaughan:
image

Screenshot of fix.

PR #104, opened 2026-03-12
docs: note WinGet release lag vs git builds

HokyoungJung:
## Summary
  - document that `winget install psmux` installs the latest published release, not the current `master` branch
  - add a short version-check reminder to the README install section
  - add an FAQ entry for cases where WinGet and git/source builds behave differently

  ## Why
  This came up while investigating a Korean input / IME issue in `psmux` + `codex`.

  At first the behavior looked like a live Korean input bug, but reproduction was inconsistent because the WinGet-
  installed build and the local git/source build were far apart in version and behavior.

  In my case:
  - WinGet build resolved to `psmux 0.4.10`
  - local git/source build resolved to `psmux 3.1`

  That version gap was large enough to completely change the reproduction result, so this PR adds a version/source check
  to the docs before deeper debugging.

  ## Scope
  This PR does **not** include a Korean IME/input fix itself.
  It only documents the version-gap trap that can make Korean input issues look reproducible on one install path and
  disappear on another.

  ## Verification
  - reviewed the updated install docs in `README.md`
  - reviewed the new FAQ entry in `docs/faq.md`
  - confirmed locally that WinGet and git/source installs can resolve to different binaries and versions

  ## Changed files
  - `README.md`
  - `docs/faq.md`

  ## Simplifications made
  - kept the guidance focused on version/source identification
  - avoided adding package-manager-specific troubleshooting beyond the minimal commands needed to compare installs

  ## Remaining risks
  - this explains one major source of confusion during Korean input debugging
  - it does not change WinGet publishing cadence or fix any actual IME behavior in code

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,15 @@ psmux is a **native Windows terminal multiplexer** built from the ground up in R
 winget install psmux
 ```
 
+WinGet installs the latest published release, not the current `master` branch.
+If you are comparing behavior against a source build from git, the versions may
+be significantly different. Check with:
+
+```powershell
+psmux -V
+tmux -V
+```
+
 ### Using Cargo
 
 ```powershell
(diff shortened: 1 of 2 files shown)


psmux:
The version gap is not big. The reason I changed the version to 3.1 is because we wanted to keep it very compliant with claude code. its basically supposed to be 0.4.2 but we decided to change the versioning to start from 3.1 going forward since tmux is expected to be versions greater than 2.

psmux:
will very shortly release the v.3.1 instead of adding that. hope that'll solve it for you

Issue #108, opened 2026-03-13
Request for ctrl+tab and ctrl+tab+shift to be used as keymaps


shivajreddy:
First off love this project thanks a ton!
Right now cant use ctrl+tab and ctrl+tab+shift as keymaps.
I migrated my tmux config to psmux, main thing i miss is to use ctrl,shift,tab keys as keymaps in psmux.

psmux:
Thanks for the kind words and the feature request!

**Ctrl+Tab and Ctrl+Shift+Tab key bindings are now fully supported** in commit 079403a.

### What was missing

The key parser (`parse_key_name`) only handled **single** modifier prefixes — `C-Tab`, `M-Tab`, and `S-Tab` each worked independently, but combinations like `C-S-Tab` failed because it tried to look up `"S-Tab"` as a key name (instead of parsing `S-` as a second modifier).

Additionally, `encode_key_event` and `parse_modified_special_key` didn't handle Tab with modifiers, so even if you could bind them, the key couldn't be forwarded to child panes or sent via `send-keys -H`.

### What's fixed

1. **Multi-modifier parsing** — `parse_key_name()` now extracts all modifier prefixes (`C-`, `M-`, `S-`) in a loop before resolving the base key. Any combination works:
   - `C-Tab` (Ctrl+Tab)
   - `C-S-Tab` (Ctrl+Shift+Tab)
   - `C-M-Tab` (Ctrl+Alt+Tab)
   - `C-M-S-Tab` (all three modifiers)
   - `C-S-Left`, `C-S-Right`, etc.

2. **VT sequence encoding** — Tab with modifiers now generates proper xterm sequences (`CSI 9;{mod}~`) for forwarding to child processes and `send-keys -H`.

### Usage

```bash
# In .psmux.conf or at runtime:
bind-key C-Tab next-window
bind-key C-S-Tab previous-window

# Or with prefix table:
bind-key -T prefix C-Tab next-window
```

### How to update

```
cargo install --git https://github.com/marlocarlo/psmux.git
```

### Verified

12/12 tests pass — binding registration, list-keys display, send-keys encoding, config file loading, and regression checks for existing bindings (C-a, S-Left, etc.).

Let me know if this covers your use case!

PR #114, opened 2026-03-15
fix: forward Shift+Enter and Ctrl+Enter to child process

ltskltsk:
Hi! Thanks for building psmux — it's been great for running persistent sessions on Windows.

I ran into an issue where Shift+Enter and Ctrl+Enter aren't being forwarded to child processes. For example, Claude Code uses Shift+Enter for multi-line input, and it doesn't work inside psmux because Enter always sends a plain `\r` regardless of modifiers.

I noticed that Tab (#108) and Arrow keys (#3) already handle modifiers correctly using `modifier_param()` and the xterm encoding pattern. Enter was just missing from those same code paths, so I followed the same approach:

## What this changes

- **`encode_key_event()`** — emits `CSI 13;{mod}~` when Enter has modifiers (e.g. `\x1b[13;2~` for Shift+Enter, `\x1b[13;5~` for Ctrl+Enter)
- **`parse_modified_special_key()`** — adds an `ENTER` arm so `send-keys S-Enter` and `send-keys C-Enter` work from the CLI
- **Client passthrough** — uses the existing `modified_key_name()` helper for Enter, same as Arrow keys already do

Plain Enter behavior is completely unchanged — this only kicks in when a modifier key is held.

Hope this is useful! Happy to adjust anything if you'd prefer a different approach.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1512,11 +1512,11 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                         if !paste_pend.is_empty() {
                                             paste_pend.push('\n');
                                         } else {
-                                            cmd_batch.push("send-key enter\n".into());
+                                            cmd_batch.push(format!("send-key {}\n", modified_key_name("Enter", key.modifiers)));
                                         }
                                     }
                                     #[cfg(not(windows))]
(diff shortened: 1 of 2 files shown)


ltskltsk:
Closing — found a pre-existing issue in parse_modified_special_key() that affects Shift-only modifiers. Will resubmit with a more complete fix.

PR #115, opened 2026-03-15
fix: forward Shift+Enter and Ctrl+Enter to child process

ltskltsk:
Hi! Thanks for building psmux — it's been great for running persistent sessions on Windows.

I ran into an issue where Shift+Enter and Ctrl+Enter aren't being forwarded to child processes. For example, Claude Code uses Shift+Enter for multi-line input, and it doesn't work inside psmux because Enter always sends a plain \`\r\` regardless of modifiers.

I noticed that Tab (#108) and Arrow keys (#3) already handle modifiers correctly using \`modifier_param()\` and the xterm encoding pattern. Enter was just missing from those same code paths, so I followed the same approach.

While tracing the code I also found a small bug in \`parse_modified_special_key()\` where Shift-only modifiers were invisible — \`m\` started at 1 and \`S-\` did \`m |= 1\` which is a no-op. This meant \`S-Left\`, \`S-Home\`, \`S-Enter\`, etc. through the \`send-keys\` path would silently fail. Fixed by accumulating bits from 0 and adding the base offset after.

## What this changes

- **\`encode_key_event()\`** — emits \`CSI 13;{mod}~\` when Enter has modifiers (e.g. \`\x1b[13;2~\` for Shift+Enter, \`\x1b[13;5~\` for Ctrl+Enter)
- **\`parse_modified_special_key()\`** — adds an \`ENTER\` arm so \`send-keys S-Enter\` and \`send-keys C-Enter\` work from the CLI
- **\`parse_modified_special_key()\`** — fixes Shift-only modifier calculation (also fixes S-Left, S-Right, S-Home etc. through this path)
- **Client passthrough** — uses the existing \`modified_key_name()\` helper for Enter, same as Arrow keys already do

Plain Enter behavior is completely unchanged — this only kicks in when a modifier key is held.

## Testing

- \`cargo build\` — compiles clean
- \`cargo test\` — all 148 existing tests pass, zero regressions
- Verified xterm modifier param math matches \`modifier_param()\` convention

Hope this is useful! Happy to adjust anything if you'd prefer a different approach.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1512,11 +1512,11 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                         if !paste_pend.is_empty() {
                                             paste_pend.push('\n');
                                         } else {
-                                            cmd_batch.push("send-key enter\n".into());
+                                            cmd_batch.push(format!("send-key {}\n", modified_key_name("Enter", key.modifiers)));
                                         }
                                     }
                                     #[cfg(not(windows))]
-                                    { cmd_batch.push("send-key enter\n".into()); }
+                                    { cmd_batch.push(format!("send-key {}\n", modified_key_name("Enter", key.modifiers))); }
                                 }
                                 KeyCode::Tab => {
                                     #[cfg(windows)]
--- a/src/input.rs
+++ b/src/input.rs
@@ -1259,20 +1259,22 @@ pub fn parse_modified_special_key(s: &str) -> Option<String> {
     let upper = s.to_uppercase();
     // Extract modifier prefixes and base key name
     let mut rest = upper.as_str();
-    let mut m: u8 = 1;
+    let mut bits: u8 = 0;
     loop {
-        if rest.starts_with("C-") { m |= 4; rest = &rest[2..]; }
-        else if rest.starts_with("M-") { m |= 2; rest = &rest[2..]; }
-        else if rest.starts_with("S-") { m |= 1; rest = &rest[2..]; }
+        if rest.starts_with("C-") { bits |= 4; rest = &rest[2..]; }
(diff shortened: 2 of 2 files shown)


psmux:
Hey @ltskltsk — really nice catch on all three of these. You were spot on with every one of them.

I independently verified all three bugs:

1. **Enter ignores modifiers in `encode_key_event()`** — confirmed. It literally just returned `\r` no matter what. Shift+Enter, Ctrl+Enter, Alt+Enter — all identical to plain Enter. This is the one that breaks Claude Code's multi-line input inside psmux.

2. **Shift-only modifier is a no-op in `parse_modified_special_key()`** — confirmed, and honestly this one was sneaky. `m` starts at 1, `S-` does `m |= 1` = still 1, check says `if m <= 1 { return None }` — so every Shift-only combo (S-Left, S-Home, S-Right, etc.) silently vanished through the send-keys path. This has been broken since the function was written and nobody noticed because the direct keyboard path uses `modifier_param()` which works correctly.

3. **Client passthrough sends plain `enter` ignoring modifiers** — confirmed. Delete, Arrow keys, Home, End all use `modified_key_name()` but Enter was just hardcoded to `"send-key enter\n"`. Copy-paste oversight from when those other keys were updated to carry modifiers.

I wrote 12 targeted tests that fail before the fix and pass after — covering every Enter modifier combo, the S-Left regression, and regression checks for C-Tab and C-Left to make sure nothing else moved.

The fix approach is the same as yours — I followed the same pattern (`modifier_param()` for encode, bits-from-0 for parse, `modified_key_name()` for client). Your analysis of the bit math was exactly right.

Fixed in a761f3e248c7bda34a515c4884a0533b381f0540. Thanks for the thorough writeup and the clean PR — made verification straightforward.

PR #118, opened 2026-03-17
Fix 6 client-server mode bugs (display-message, mouse, popup, paste, right-click)

ToxMox:
I ran into these issues while using psmux in client-server mode from VS Code today and ended up forking and fixing them for myself. Figured they might help others out too. Thanks for the awesome work on psmux!

## Context

When psmux runs in client-server mode (`psmux attach` from VS Code terminal or Windows Terminal), several features silently break because the client and server have separate input/rendering paths. These fixes address the most impactful issues for multi-client workflows and VS Code terminal usage.

### display-message
The rendering infrastructure existed (server serializes `status_message` to JSON, client renders it on the status bar with `message_style`) but the `DisplayMessage` handler only expanded the format string and returned it — it never set `app.status_message`. One-line fix.

### Mouse multi-client tracking
With multiple clients attached at different terminal sizes, mouse clicks on the status bar only worked in whichever client last resized. `latest_client_id` was only updated on `ClientAttach` and `ClientSize` — not on mouse events. Added `client_id` to all 10 mouse/scroll `CtrlReq` variants. The server now recomputes effective window size from the clicking client's dimensions before processing coordinates.

### Popup race condition
`display-popup` with fast commands (e.g. `echo test`) produced blank popups. The PTY reader thread hadn't processed any output before the first frame was serialized to the client. Added a 50ms delay after PTY spawn to let the reader thread populate the vt100 parser.

### Paste fragmentation (VS Code terminal)
Large pastes in VS Code's terminal arrived as multiple `[Pasted text #N]` chunks instead of one. Root cause chain:
1. ConPTY with VTI disabled strips bracket paste markers, delivering paste as individual `KEY_EVENT` records
2. The client's `paste_pend` burst detector entered stage2 but timed out at 300ms mid-paste (ConPTY delivers text over ~1 second)
3. `Ctrl+V Release` triggered an immediate flush before ConPTY finished injecting characters
4. Right-click clipboard paste sent the full content, then ConPTY also injected the same text as key events — producing a duplicate
5. Non-ASCII characters (em-dashes, etc.) triggered immediate flush as "normal text" instead of accumulating

Fixes: `Ctrl+V Release` no longer flushes immediately — promotes to stage2 for growth detection so ConPTY can finish delivery; right-click clipboard paste suppresses duplicate ConPTY key events for 2s (discards text keys during the suppression window to prevent leaked Enter/Tab); large non-ASCII buffers (>=20 chars) enter stage2 instead of being treated as IME input; added growth detection to stage2 timeout so ongoing pastes aren't split.

### Right-click copy triggers unwanted paste
In VS Code with "smart copy/paste" right-click behavior, highlighting text and right-clicking would copy the selection AND paste the clipboard content. VS Code injects clipboard content as key events after the copy action. Uses the same 2s suppression window with full text-key filtering.

### Clipboard line ending normalization
Right-click paste produced double-spaced text because the Windows clipboard preserves `\r\n` line endings. Now normalizes `\r\n` to `\n` and trims trailing newlines from clipboard reads.

## Test plan
- [ ] `display-message "hello"` from command prompt shows in status bar
- [ ] Two clients attached with different sizes — mouse clicks work in both
- [ ] `display-popup -E "echo hello"` renders content (not blank)
- [ ] Ctrl+V paste of 40+ lines in VS Code terminal arrives as single chunk
- [ ] Right-click paste in VS Code arrives as single chunk, matching Ctrl+V formatting
- [ ] Highlight + right-click copies without pasting or inserting carriage returns
- [ ] Paste containing non-ASCII (em-dashes, backticks) arrives as single chunk
- [ ] ESC key works normally (not swallowed or delayed)
- [ ] Normal typing, IME input, and Ctrl+V in Windows Terminal remain unaffected

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -851,6 +851,22 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         paste_confirmed = true;
                     }
                     Event::Key(key) if key.kind == KeyEventKind::Press || key.kind == KeyEventKind::Repeat => {
+                        // During paste suppression, discard text keys (chars,
+                        // Enter, Tab, Space) that VS Code injects as ConPTY
+                        // echoes after a right-click clipboard paste/copy.
+                        #[cfg(windows)]
+                        {
+                            if paste_suppress_until.is_some() {
+                                let is_text_key = matches!(key.code,
+                                    KeyCode::Char(_) | KeyCode::Enter | KeyCode::Tab
+                                ) && !key.modifiers.contains(KeyModifiers::CONTROL)
+                                  && !key.modifiers.contains(KeyModifiers::ALT);
+                                if is_text_key {
+                                    _pending_evt = input.try_read()?;
+                                    continue;
+                                }
+                            }
+                        }
                         // Flush pending paste buffer before processing any non-bufferable key.
                         // Bufferable keys are: plain Char, Space, Enter (if pend non-empty), Tab (if pend non-empty).
                         #[cfg(windows)]
@@ -1683,6 +1699,9 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                     rsel_end = None;
                                     selection_changed = true;
                                     if let Some(text) = read_from_system_clipboard() {
+                                        // Normalize \r\n → \n (Windows clipboard preserves \r)
+                                        // and trim trailing newlines to avoid blank lines.
+                                        let text = text.replace("\r\n", "\n").trim_end_matches('\n').to_string();
                                         if !text.is_empty() {
                                             let encoded = base64_encode(&text);
                                             cmd_batch.push(format!("send-paste {}\n", encoded));
@@ -1819,6 +1838,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
             } else if paste_confirmed && paste_pend.is_empty() {
                 // Ctrl+V with no buffered chars — read clipboard as fallback
                 if let Some(text) = read_from_system_clipboard() {
+                    let text = text.replace("\r\n", "\n");
                     if !text.is_empty() {
                         if input_log_enabled() {
                             input_log("paste", &format!("paste CONFIRMED (no buffer), clipboard read len={}", text.len()));


ToxMox:
Apologies for the extra commits after opening — I was still testing edge cases when I submitted. The last two commits fix:

- **ESC key regression**: Removed the bracket paste detection state machine from client mode. ConPTY strips bracket paste markers when VTI is disabled, so it never fired — but it was buffering ESC keys for 5ms and sometimes swallowing them.
- **Clipboard line endings**: Right-click paste was double-spacing text because the Windows clipboard preserves `\r\n`. Now normalizes to `\n` and trims trailing newlines. Also tightened the paste suppression window to discard leaked Enter/Tab key events that VS Code injects after a right-click copy.

Should be stable now — tested right-click paste, Ctrl+V paste, highlight+right-click copy, ESC, and normal typing.

psmux:
Hey @ToxMox, thanks for the incredibly thorough analysis! I went through each of the 6 bugs you identified and can confirm **every single one is real** — great detective work.

I've independently fixed all 6 in commit ffcf3d3:

**1. display-message** — Added a `set_status_bar` flag to the `DisplayMessage` CtrlReq. The handler now sets `app.status_message` when `-p` isn't used. The `list-sessions -F` path passes `false` so it doesn't pollute the status bar.

**2. Mouse multi-client** — Added `client_id` (`u64`) to all 10 mouse/scroll `CtrlReq` variants. The server now updates `latest_client_id` before processing each mouse event, so coordinate resolution uses the clicking client's dimensions.

**3. Popup race** — Added a 50ms `thread::sleep` after PTY spawn + reader thread creation, before setting the popup mode. This gives the reader thread time to process initial output before the first frame serializes.

**4. Paste fragmentation** — Added growth detection to the stage2 timeout: if `paste_pend.len()` increased since the last 300ms check, the timer resets instead of splitting. Also added a path for large non-ASCII buffers (≥20 chars) to enter stage2 instead of being treated as IME input.

**5. Right-click copy suppression** — After a right-click copy, text key events are now suppressed for 2 seconds via `paste_suppress_until`. This prevents VS Code's ConPTY from injecting duplicate clipboard content.

**6. Clipboard CRLF** — `read_from_system_clipboard()` now normalizes `\r\n` → `\n` before returning.

I'm closing this PR since the fixes are already on master, but I really appreciate you taking the time to identify and document all of these — the root cause analysis in your PR description was spot-on. Cheers!

Issue #122, opened 2026-03-17
Backslash key produces " + newline in panes since v3.1.0-fix-70-71


schgoo:
## Summary

Typing `\` in a pane outputs `"` followed by a newline. Works correctly in the command prompt (`Ctrl+b :`).

## Reproduction

1. `psmux new-session` (default config, no plugins needed)
2. Type `\` in the pane

Observed: `"` appears followed by a newline
Expected: `\`

Confirmed with ReadKey diagnostics — a single `\` keypress produces three events in the pane:
```
Key=Oem5 | Char=\ (0x5C) | Mod=None      ← correct
Key=Oem7 | Char=" (0x22) | Mod=Shift      ← phantom
Key=Enter | Char=\n (0xA) | Mod=Control    ← phantom
```

## Cause

Introduced in `2e4971b` ("fix: preserve backslashes in quoted paths for Windows paths").

That commit changed `parse_command_line()` in `commands.rs` to treat backslashes as literal inside double quotes (so Windows paths like `"C:\Program Files\..."` aren't mangled). But it only kept `\"` as an escape — it dropped `\\`.

The client sends backslash as `send-text "\\"` (escaping `\` → `\\` at client.rs:742). The server parser now reads `"\\"` as:
- `\` — literal (not followed by `"`)
- `\"` — escaped quote (consumes the closing `"`)

So the arg becomes `\"` + whatever follows (the newline), instead of just `\`.

## Fix

Add `\\` back as a recognized escape in `parse_command_line()`:

```rust
} else if i + 1 < chars.len() && chars[i + 1] == '\\' {
    current.push('\\');
    i += 1; // skip the second backslash
```

Between the existing `\"` check and the literal-backslash fallback. Windows paths are unaffected since they don't contain `\\` (only single `\`).

## Environment

- psmux 3.2.0 (also 3.1.0-fix-70-71)
- Windows 11, Windows Terminal
- Tested on both US-QWERTY and US-Dvorak

psmux:
Hey @schgoo — this was already fixed in b37c89e (from your PR #123 investigation).

The root cause was exactly as you described: `parse_command_line()` only recognized `\"` as an escape inside double quotes but not `\\`. So when the client sent `send-text "\\"` the parser consumed `\"` as an escaped quote, eating the closing delimiter, and the remainder (`\"`) plus the newline leaked through as phantom keystrokes.

[...]

PR #123, opened 2026-03-17
fix: handle \\\\ escape in parse_command_line (backslash key broken in panes)

schgoo:
Fix for #122

### DIFF
--- a/tests/test_issue122_backslash_escape.ps1
+++ b/tests/test_issue122_backslash_escape.ps1
@@ -1,0 +1,172 @@
+# psmux Issue #122 — Backslash escape in parse_command_line
+#
+# Tests that \\ inside double-quoted strings produces a single literal
+# backslash, while preserving existing behaviour: bare backslashes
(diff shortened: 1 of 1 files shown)


psmux:
Hey @schgoo, nice catch! Your fix is spot-on — the `parse_command_line` parser was only handling `\"` as an escape inside double quotes, so `send-text "\\"` (the client's encoding for a single backslash) had its second `\` steal the closing quote via the `\"` rule. Result: the PTY received `\"` instead of `\`, and the shell showed `"` + `>>` continuation prompt.

I've landed the same core fix in commit b37c89e — adding `\\` as a recognized escape sequence (producing a single `\`), which is identical to what your PR does. Also adopted your unit tests idea and added a test script for end-to-end verification.

Thanks for the clean analysis and the tests — really appreciate the contribution. Closing this since the fix is already on master, but you're credited in the commit. Cheers!

schgoo:
Nice! Thanks for getting to this so quickly!

psmux:
You're welcome! Share the good news about Psmux!

PR #131, opened 2026-03-18
Fix multi-line paste appearing in reverse order in PowerShell

ToxMox:
## Summary
- Multi-line paste into PowerShell/PSReadLine displays lines in reverse order
- Clipboard text arrives with LF (`\n`) or CRLF (`\r\n`), but ConPTY's input parser expects CR (`\r`) for Enter
- Bare LF is misinterpreted by PSReadLine, causing reverse line insertion
- Fix: normalize line endings to CR in `write_paste_chunked` before writing to ConPTY

## Test plan
- [ ] Paste multi-line text into PowerShell — lines should appear in correct order
- [ ] Paste multi-line text into Claude Code — should still work correctly
- [ ] Paste into nvim/helix — verify no regressions with bracket paste

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1992,6 +1992,28 @@ pub fn handle_mouse(app: &mut AppState, me: MouseEvent, window_area: Rect) -> io
 /// are tiny and always written in one shot.
 fn write_paste_chunked(writer: &mut dyn std::io::Write, text: &[u8], bracket: bool) {
     const CHUNK: usize = 512;
+    // Normalize line endings to CR for ConPTY.  Clipboard text may arrive
+    // with LF (\n) or CRLF (\r\n), but ConPTY's input parser expects CR
+    // (\r) for Enter.  Bare LF is misinterpreted by PSReadLine, causing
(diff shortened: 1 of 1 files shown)


psmux:
Hey @ToxMox, thanks for catching this — you nailed the root cause. Multi-line paste was indeed broken because write_paste_chunked was sending bare LF bytes straight through the PTY pipe, and ConPTY/PSReadLine expects CR for Enter. The clipboard normalization in 
ead_from_system_clipboard() converts CRLF→LF, but that last LF→CR step for ConPTY input was completely missing.

I wrote targeted unit tests (LF-only, CRLF-only, mixed endings, and bracketed paste) — all of them failed against the current code, confirming the bug. After adding the normalization logic (same approach you proposed: CRLF→CR, LF→CR in write_paste_chunked), all tests pass and no regressions in the existing 62-test suite.

Fixed in abd5e25. I'm closing this PR since the fix is already on master, but really appreciate the diagnosis — it was spot-on.

psmux:
Let me know if you find it fixed in the latest commit. Thank you.

PR #132, opened 2026-03-18
Fix Shift+Enter for VS Code terminal with Node.js apps (Claude Code)

ToxMox:
## Summary

Shift+Enter doesn't work for Node.js apps (like Claude Code) running inside psmux in VS Code's terminal. The existing CSI `\x1b[13;2~` encoding from `a761f3e` is dropped by ConPTY (it doesn't recognize code 13 in `~` format), so the keypress never reaches the child process.

This PR fixes both the client-side detection and server-side delivery for the VS Code + ConPTY + Node.js case.

## Root cause

Two separate issues in the VS Code terminal path:

**Client:** VS Code's xterm.js sends `\x1b\r` (ESC + CR) for Shift+Enter. ConPTY interprets the ESC prefix as Alt, so crossterm reports Alt+Enter instead of Shift+Enter.

**Server:** Node.js apps use libuv's `uv_tty_read_raw`, which translates `VK_RETURN` to `\r` regardless of modifier flags. Both the CSI `\x1b[13;2~` encoding (dropped by ConPTY) and win32-input-mode `KEY_EVENT_RECORD` with `SHIFT_PRESSED` (libuv ignores shift on Enter) fail to deliver the modifier.

## Fix

**Client (`platform.rs`):** `augment_enter_shift` polls `GetAsyncKeyState(VK_SHIFT)` to detect the physical Shift key and remaps crossterm's misreported Alt+Enter back to Shift+Enter.

**Server (`input.rs`):** Sends `\x1b\r` (ESC + CR) for Shift+Enter — the same bytes VS Code's xterm.js sends. The round-trip works because libuv *does* preserve Alt as an ESC prefix:

```
\x1b\r → ConPTY: Alt+Enter → libuv: \x1b\r → app interprets as Shift+Enter
```

## Trade-offs

- `GetAsyncKeyState` is a heuristic — there's a small race window between keypress and event processing. In practice this is reliable for interactive use.
- ConPTY delivers the key as Alt+Enter, not Shift+Enter. PSReadLine users need to add a keybinding to map Alt+Enter to the same AddLine action (see User Setup below).
- This is the same limitation that exists in VS Code's terminal without psmux — xterm.js always sends `\x1b\r` for Shift+Enter, which ConPTY interprets as Alt+Enter.

## User setup

For full Shift+Enter support in VS Code, users also need:

**VS Code keybindings.json** — ensures VS Code sends `\x1b\r` for Shift+Enter in terminal:
```json
{
    "key": "shift+enter",
    "command": "workbench.action.terminal.sendSequence",
    "args": {
        "text": "\u001b\r"
    },
    "when": "terminalFocus && terminalShellType != 'pwsh'"
}
```

**PowerShell profile** (`$PROFILE`) — maps Alt+Enter to AddLine since ConPTY delivers the key as Alt+Enter:
```powershell
Set-PSReadLineKeyHandler -Chord 'Alt+Enter' -Function AddLine
```

## Test plan
- [ ] Press Shift+Enter in Claude Code inside psmux (VS Code terminal) — should insert newline
- [ ] Press Shift+Enter in PowerShell with the PSReadLine keybinding — should insert newline
- [ ] Plain Enter still works normally in all apps
- [ ] Alt+char shortcuts (Alt+F, Alt+B) still work in PSReadLine

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -838,7 +838,13 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         }
                         paste_confirmed = true;
                     }
-                    Event::Key(key) if key.kind == KeyEventKind::Press || key.kind == KeyEventKind::Repeat => {
+                    Event::Key(mut key) if key.kind == KeyEventKind::Press || key.kind == KeyEventKind::Repeat => {
+                        // On Windows, VS Code's xterm.js sends ESC+CR for
+                        // Shift+Enter.  ConPTY interprets the ESC as Alt, so
+                        // crossterm reports Alt+Enter.  Poll the physical
+                        // keyboard to detect the real modifier.
+                        #[cfg(windows)]
+                        crate::platform::augment_enter_shift(&mut key);
                         // Flush pending paste buffer before processing any non-bufferable key.
                         // Bufferable keys are: plain Char, Space, Enter (if pend non-empty), Tab (if pend non-empty).
                         #[cfg(windows)]
--- a/src/input.rs
+++ b/src/input.rs
@@ -2609,6 +2609,40 @@ pub fn send_key_to_active(app: &mut AppState, k: &str) -> io::Result<()> {
                     let _ = p.writer.write_all(&[0x1b, ctrl_char]);
                 }
             }
+            // Modified Enter: send ESC + CR (\x1b\r) for Shift/Alt+Enter.
+            //
+            // This matches what VS Code's xterm.js sends for Shift+Enter.
+            // The round-trip is: \x1b\r → ConPTY interprets ESC as Alt →
+            // KEY_EVENT_RECORD(VK_RETURN, LEFT_ALT_PRESSED) → libuv
+            // translates Alt back to ESC prefix → child gets \x1b\r.
+            // Claude Code (and other Node.js apps) interpret \x1b\r as
+            // Shift+Enter.  PSReadLine also handles this correctly when
+            // Alt+Enter is bound to AddLine.
+            //
+            // The CSI \x1b[13;2~ encoding (from parse_modified_special_key)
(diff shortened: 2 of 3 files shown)


psmux:
Hey @ToxMox, solid work on this one — both root causes you identified are correct.

**Client side:** VS Code's xterm.js sends \\x1b\r\ for Shift+Enter, ConPTY interprets the ESC as an Alt prefix, and crossterm delivers Alt+Enter instead of Shift+Enter. Your \ugment_enter_shift\ approach using \GetAsyncKeyState(VK_SHIFT)\ is the right call — there's no other way to recover the real modifier once ConPTY has already eaten it.

**Server side:** CSI \\x1b[13;2~\ is non-standard (code 13 isn't in the VT function key table), and ConPTY silently drops it. Sending \\x1b\r\ (ESC+CR) instead works because the round-trip preserves the ESC through libuv's Alt-prefix handling, so Node.js apps like Claude Code receive it correctly.

I independently verified both bugs by writing unit tests — Shift+Enter encoding produced \\x1b[13;2~\ (7 bytes) instead of \\x1b\r\ (2 bytes), and Alt+Enter had the same issue. After applying the fix (same approach as your PR), all new tests pass along with the full 68-test suite. I also extended \ncode_key_event\ with the same Windows-conditional logic for the embedded-mode path, and added the \#[cfg(windows)]\ guard in \send_key_to_active\ for the S-Enter/M-Enter command dispatch.

Fixed in abd5e25. Closing this PR since master already has the fix, but thanks for the thorough analysis — the ConPTY round-trip explanation was particularly helpful.

ToxMox:
Awesome! Glad my PRs have been useful. Thanks for the quick implementations! 😎

psmux:
@ToxMox Yes they have been. Thank you!

Issue #121, opened 2026-03-17
Shift+Enter (and other modifier+Enter) not working — always sends bare CR


dy-KingDiamond:
## Problem

When running inside a psmux session, **Shift+Enter** does not produce the expected escape sequence. Applications that distinguish Shift+Enter from Enter (e.g. Claude Code, which uses Shift+Enter for newline insertion) cannot detect the modifier.

**Without psmux**: Shift+Enter works correctly (terminal emulator sends the appropriate CSI sequence).  
**With psmux**: Shift+Enter behaves identically to plain Enter.

## Root Cause

In `src/input.rs`, the `encode_key_event()` function unconditionally maps `KeyCode::Enter` to `\r`, ignoring all modifier keys:

```rust
KeyCode::Enter => b"\r".to_vec(),
```

By contrast, `KeyCode::Tab` correctly handles modifiers using `modifier_param()`:

```rust
KeyCode::Tab => {
    let m = modifier_param(key.modifiers);
    if m > 1 {
        format!("\x1b[9;{}~", m).into_bytes()
    } else {
        b"\t".to_vec()
    }
}
```

Enter should follow the same pattern.

## Proposed Fix

Apply the same modifier-aware encoding to Enter, using CSI code 13 (the ASCII value of CR):

```rust
KeyCode::Enter => {
    let m = modifier_param(key.modifiers);
    if m > 1 {
        format!("\x1b[13;{}~", m).into_bytes()
    } else {
        b"\r".to_vec()
    }
}
```

This produces:
| Key Combination | Output |
|---|---|
| Enter (unmodified) | `\r` (unchanged, backward compatible) |
| Shift+Enter | `\x1b[13;2~` |
| Alt+Enter | `\x1b[13;3~` |
| Ctrl+Enter | `\x1b[13;5~` |

## Environment

- psmux v0.4.10
- Windows 11
- Terminal: Alacritty
- Affected application: Claude Code (uses Shift+Enter for newline)

psmux:
Hey @dy-KingDiamond — this is fixed! You were exactly right about the root cause.

`encode_key_event()` was unconditionally returning `\r` for Enter regardless of modifiers, while Tab right next to it had the correct `modifier_param()` pattern. The fix applies the same approach — `CSI 13;{mod}~` when modifiers are held, plain `\r` otherwise.

Also fixed two related issues in the same commit:
[...]

PR #159, opened 2026-03-27
feat: win32-input-mode key encoding — replace per-key hacks with uniform modifier support

ToxMox:
## Summary

- Encode all key events as win32-input-mode sequences (`ESC[Vk;Sc;Uc;Kd;Cs;Rc_`) on the output side, replacing scattered per-key workarounds
- Fix the entire client-server key pipeline to preserve modifiers for every key
- Add libuv/Node.js compatibility layer so Shift+Enter and Ctrl+Backspace work in Claude Code
- Reduce paste detection window from 20ms to 5ms, eliminating typing lag and false paste detection

## Background

psmux has a double-ConPTY architecture: outer terminal -> ConPTY -> psmux -> ConPTY -> child app. Standard VT sequences have no way to encode modifiers for keys like Enter, Backspace, Tab, and Esc -- Enter is just `\r`, there is no Shift+Enter in VT. So modifiers get lost in transit and the child app sees plain keystrokes.

This has been a recurring source of bugs. Shift+Enter does not work in Claude Code (#121). Ctrl+Backspace does not delete words in PowerShell. Every time one of these came up, I added a targeted fix: ESC+CR for Shift+Enter, WriteConsoleInputW injection for modified Enter, CSI u encoding for Ctrl+Backspace. Each fix had different failure modes -- ConPTY drops CSI u, WriteConsoleInputW silently fails with pseudo-consoles, ESC+CR only helps Enter. Classic whack-a-mole that does not scale.

I wanted a single encoding path that handles ALL keys uniformly -- fix the architecture instead of chasing individual keys.

## The solution: win32-input-mode

Win32-input-mode is Microsoft's answer to this problem, used by Windows Terminal internally. Every key event is encoded as `ESC[Vk;Sc;Uc;Kd;Cs;Rc_` -- the full Win32 KEY_EVENT_RECORD serialized as a VT sequence (virtual key code, scan code, unicode char, key-down flag, control key state with all modifier flags, repeat count). ConPTY reconstructs the exact KEY_EVENT_RECORD on the other side. Zero information loss.

psmux already creates child ConPTY with the `PSEUDOCONSOLE_WIN32_INPUT_MODE` (0x4) flag, so the ConPTY input parser accepts `_`-terminated sequences unconditionally. psmux just was not encoding keys that way -- it was writing standard VT bytes that lose modifier information.

## Fallback behavior

This is a Windows-only feature with graceful degradation at every layer:

**Output side** (psmux -> child ConPTY): Always active on Windows. Every key event is encoded as a `_`-terminated win32-input-mode sequence. The child ConPTY has `PSEUDOCONSOLE_WIN32_INPUT_MODE` set at creation, so it always accepts these. No detection or negotiation needed.

**Input side** (outer terminal -> psmux): On startup, psmux writes `\x1b[?9001h` to request win32-input-mode from the outer terminal. Terminals that support it (Windows Terminal, VS Code, Tabby) respond by sending full-fidelity key events with correct modifier flags -- crossterm sees the real SHIFT/CTRL/ALT directly. Terminals that don't support it (WezTerm, older conhost) silently ignore the sequence, and the existing `augment_enter_shift` fallback kicks in -- it polls physical keyboard state via `GetAsyncKeyState` to detect Shift on Enter. Other modifiers (Ctrl, Alt) already work without win32-input-mode because they change the character value (Ctrl+C -> 0x03) or add an ESC prefix (Alt+x -> `\x1b`x).

**Non-Windows**: The entire win32-input-mode path is `#[cfg(windows)]`. Non-Windows builds use standard VT encoding unchanged -- no behavioral difference, no new code compiled.

### What changed

**New module: `src/win32_input.rs`**

Single encoding function `encode_key_win32(key: &KeyEvent) -> Option<Vec<u8>>` that handles ALL crossterm key events:
- Static VK/scan/unicode mapping table for named keys (Enter, Tab, Backspace, Esc, arrows, F-keys, nav cluster)
- Dynamic `VkKeyScanW`/`MapVirtualKeyW` lookup for character keys at runtime
- Modifier flags computation (SHIFT, CTRL, ALT, ENHANCED_KEY)
- AltGr detection (Ctrl+Alt on non-lowercase = international keyboard character, not genuine Ctrl+Alt)

Plus `encode_key_name_win32(name: &str)` for the server's string-based key name path.

**Output side (`src/input.rs`)**

- `forward_key_to_active`: replaced the 80-line modified-Enter hack block (ESC+CR, WriteConsoleInputW injection, CSI fallback) and `encode_key_event` call with a single `encode_key_win32` call on Windows
- `send_key_to_active`: added win32-input-mode early return before the VT match arms -- all recognized keys go through `encode_key_name_win32` on Windows

**Client side (`src/client.rs`)**

The client reads crossterm KeyEvents and sends string key names to the server. Several keys were hardcoding the name without modifiers -- Backspace sent `"backspace"` regardless of Ctrl/Shift, Tab sent `"tab"`, Esc sent `"esc"`, BackTab sent `"btab"`, Space matched before modifier checks. The `modified_key_name()` helper already existed and was used for arrows, Delete, Enter, and F-keys -- I extended it to ALL keys so the entire pipeline preserves modifiers uniformly.

**Cleanup (`src/platform.rs`, `src/input.rs`, `tests-rs/test_input.rs`)**

Removed `send_modified_enter_event` (WriteConsoleInputW injection -- 120 lines), removed the `#[cfg(windows)]` ESC+CR hack from `encode_key_event`, removed the Windows-specific modified-Enter block from `send_key_to_active`. Updated tests to expect uniform CSI encoding from `encode_key_event` (which is now only used on non-Windows).

### libuv compatibility layer

Testing revealed that Node.js apps (Claude Code) still did not see Shift+Enter correctly, even with proper win32-input-mode encoding. Research into the ConPTY -> libuv -> Node.js pipeline uncovered the root cause:

**libuv reads via `ReadConsoleInputW`** and sees the full KEY_EVENT_RECORD -- but its translation logic ignores `SHIFT_PRESSED` for any key with non-zero `UnicodeChar`. It only checks `LEFT_ALT_PRESSED` (prepends ESC) and lets `CTRL` affect the unicode value itself. For VK_RETURN with SHIFT, libuv emits `\r` -- SHIFT is silently dropped. This is [Node.js #2890](https://github.com/nodejs/node/issues/2890), open since 2015.

The fix maps `SHIFT_PRESSED` to `LEFT_ALT_PRESSED` for Enter only. libuv then prepends ESC, producing `\x1b\r` for Shift+Enter -- which terminal apps recognize as modified Enter. PSReadLine sees Alt+Enter via ReadConsoleInputW, matching VS Code's behavior (VS Code's xterm.js sends `\x1b\r` for Shift+Enter, which ConPTY reports as Alt+Enter to PSReadLine even without psmux).

The mapping is restricted to Enter because keyboard hardware reports Shift on the next key-down before the Shift key-up arrives during normal typing speed. A broad mapping across all named keys (Space, Tab, Backspace, Esc) caused phantom ESC prefixes -- e.g. typing `? ` at normal speed would produce ESC+Space instead of plain Space because the physical Shift key-up had not yet arrived when Space key-down fired. Confirmed with a raw Win32 KEY_EVENT_RECORD viewer showing 10/10 Shift+Space when typing `? ` at natural speed. For non-Enter keys, libuv naturally ignores SHIFT_PRESSED, which is the correct behavior.

Similarly, Ctrl+Backspace was sending `unicode=0x08` (the static table value). But Windows produces `unicode=0x7F` for Ctrl+Backspace, and Node.js maps `0x7F` to `{name: "backspace", ctrl: false}` -- another Node.js bug. VS Code works around this by intercepting Ctrl+Backspace and sending `0x17` (Ctrl+W/ETB) -- the universal word-delete byte. The encoder now applies the same CTRL unicode transformations: `BS->0x17`, `CR->0x0A`, `u>=0x20->u&0x1F`.

### KeyEvent normalization and key binding fixes (`src/config.rs`, `src/client.rs`, `src/input.rs`)

While testing the ppm plugin manager (Prefix+I to install, Prefix+U to update, Prefix+M to clean), I discovered that all uppercase prefix bindings silently broke. Two independent bugs in `src/config.rs`, plus a missing normalization step:

1. **`normalize_key_for_binding`**: Win32-input-mode reports Shift+I as `Char('i') + SHIFT` instead of `Char('I')`. The function stripped SHIFT but didn't uppercase the letter. Fixed: when SHIFT is present and the char is a lowercase ASCII letter, uppercase it before stripping. No-op for VT mode since SHIFT is never set there for letter keys.

2. **`parse_key_string`**: The server-side CLI key parser ran `to_lowercase()` on all characters, so `psmux bind-key I ...` stored as `Char('i')`. Pre-existing bug unrelated to win32-input-mode -- config-file parser (`parse_key_name`) already preserved case correctly. Fixed: single characters preserve original case; `to_lowercase()` now only applies to multi-character named keys like "Space", "Enter", "F1".

3. **Early KeyEvent normalization**: Added `Char('lowercase') + SHIFT` -> `Char('UPPERCASE')` normalization in both the client event loop (after `augment_enter_shift`) and the server's `handle_key` function. This covers all downstream uppercase matches -- copy mode vi keys (G, W, B, E, H, M, L, F, T, D, V, A, N), tree/session chooser (G), and confirm prompts (Y/N). No-op for VT mode.

### Paste detection timing fix

During testing I also noticed that fast typing in PowerShell caused lag bursts and Claude Code would show false "pasting text" messages. This turned out to be pre-existing (confirmed by testing the pre-change binary) -- the client's Windows paste detection buffers every printable keystroke for 20ms before forwarding. Reduced to 5ms, which is still reliable for paste detection (paste bursts deliver dozens of chars in <2ms) but eliminates false positives from human typing (max ~16 chars/sec = 62ms between keystrokes). Non-ASCII paste threshold reduced proportionally from 20 to 8 characters.

## Files changed

| File | What |
|-|-|
| `src/win32_input.rs` | New module: VK mapping table, `encode_key_win32`, `encode_key_name_win32`, tests |
| `src/input.rs` | Replace per-key hacks with `encode_key_win32` calls, Shift+letter KeyEvent normalization |
| `src/client.rs` | Preserve modifiers for all keys, paste timing fix, Shift+letter KeyEvent normalization |
| `src/config.rs` | Fix key binding normalization for Shift+letter, preserve case in CLI key parser |
| `src/main.rs` | `mod win32_input`, Phase 2 (request win32-input-mode from outer terminal) |
| `src/platform.rs` | Remove `send_modified_enter_event` |
| `tests-rs/test_input.rs` | Update assertions for uniform CSI encoding |

Net: -82 lines (the new module replaces more hack code than it adds)

## Design decisions

**Why win32-input-mode instead of CSI u (kitty protocol)?** ConPTY drops CSI u sequences -- they never reach the child. Win32-input-mode is Microsoft's own protocol, accepted by ConPTY unconditionally when the `PSEUDOCONSOLE_WIN32_INPUT_MODE` flag is set.

**Why SHIFT->ALT mapping only for Enter?** libuv ignores SHIFT_PRESSED for keys with non-zero UnicodeChar but prepends ESC for LEFT_ALT_PRESSED. The mapping makes Shift+Enter produce `\x1b\r` which Claude Code recognizes. It is restricted to Enter because keyboard hardware reports Shift on the next key-down before Shift key-up during fast typing -- a broad mapping caused phantom ESC prefixes on Space, Tab, etc. For non-Enter keys, libuv ignoring SHIFT is the correct behavior.

**Why Ctrl+Backspace sends 0x17 instead of 0x7F?** Node.js maps 0x7F to plain backspace ([#2890](https://github.com/nodejs/node/issues/2890)). VS Code works around this by sending 0x17 (Ctrl+W) for Ctrl+Backspace. PSReadLine is unaffected -- it matches on VK_BACK + CTRL flags, not the unicode value.

**Why 5ms paste detection instead of 0?** Cannot send chars immediately because bracket paste wrapping (`\x1b[200~`...`\x1b[201~`) must be applied before characters reach the child. Need a short window to accumulate the paste burst. 5ms is enough for paste (arrives in <2ms) but too short for human typing (62ms+ between keystrokes).

**Why request win32-input-mode from the outer terminal (Phase 2)?** Writing `\x1b[?9001h` to stdout asks terminals like Windows Terminal and VS Code to send full-fidelity key events. crossterm then sees correct modifier flags directly, making `augment_enter_shift` (GetAsyncKeyState polling) unnecessary in supported terminals. Unsupported terminals silently ignore the sequence and `augment_enter_shift` remains as fallback.

## Test plan

- [ ] Shift+Enter in Claude Code (VS Code terminal) -- inserts newline, does not submit
- [ ] Shift+Enter in Claude Code (Windows Terminal) -- same
- [ ] Ctrl+Backspace in PowerShell -- deletes word (BackwardKillWord)
- [ ] Ctrl+Backspace in Claude Code -- deletes word
- [ ] Plain Enter, Backspace, Tab, Esc -- unchanged behavior
- [ ] Arrow keys with Shift/Ctrl -- text selection, word jump
- [ ] AltGr characters (German `\`, `@`, `{`) -- correct characters
- [ ] Ctrl+C -- interrupt works
- [ ] Esc -- works (release build only -- debug builds have timing issues)
- [ ] Fast typing in PowerShell -- no lag bursts
- [ ] Fast typing in Claude Code -- no false "pasting text" message
- [ ] Shift+punctuation then Space at normal speed -- no phantom ESC prefix
- [ ] Ctrl+V paste -- correctly detected and bracket-wrapped
- [ ] Uppercase prefix bindings (Prefix+I, Prefix+U) -- work with ppm and custom bindings
- [ ] Copy mode vi keys: G (bottom), W/B (word jump), H/M/L (screen position), V (visual line) -- all work with Shift+letter
- [ ] Right-click paste -- duplicate suppression still works
- [ ] Non-Windows builds -- compile clean, VT encoding unchanged

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -875,6 +875,15 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         // keyboard to detect the real modifier.
                         #[cfg(windows)]
                         crate::platform::augment_enter_shift(&mut key);
+                        // Win32-input-mode: Shift+letter arrives as Char('a') + SHIFT.
+                        // Normalize to Char('A') to match VT-mode behavior so all
+                        // downstream KeyCode::Char('G') etc. matches work unchanged.
+                        if let KeyCode::Char(c) = key.code {
+                            if key.modifiers.contains(KeyModifiers::SHIFT) && c.is_ascii_lowercase() {
+                                key.code = KeyCode::Char(c.to_ascii_uppercase());
+                                key.modifiers.remove(KeyModifiers::SHIFT);
+                            }
+                        }
                         // During paste suppression, discard duplicate key events.
                         // Timer-based: after right-click copy (no content to match).
                         // Content-matched: after right-click paste (match against clipboard).
--- a/src/client.rs
+++ b/src/client.rs
@@ -374,6 +374,12 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
     // for a short period to prevent VS Code ConPTY duplicate injection.
     #[cfg(windows)]
     let mut paste_suppress_until: Option<Instant> = None;
+    // Content-matched suppression: after right-click paste, discard ConPTY
+    // duplicate key events by matching them against the pasted clipboard text.
+    #[cfg(windows)]
+    let mut paste_suppress_chars: Vec<char> = Vec::new();
+    #[cfg(windows)]
+    let mut paste_suppress_pos: usize = 0;
 
     // list-keys overlay state (C-b ?)
     let mut keys_viewer = false;
@@ -730,7 +736,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                             input_log("paste", &format!("stage2: {} chars in 20ms, waiting for Ctrl+V Release", paste_pend.len()));
                         }
                     } else if paste_pend.len() >= 20 && has_non_ascii {
-                        // ≥20 non-ASCII chars in 20ms — almost certainly a paste
+                        // ≥8 non-ASCII chars in 20ms — almost certainly a paste
                         // containing Unicode content (em-dashes, CJK, etc.), not
                         // IME composition (which rarely exceeds a few chars).
                         paste_stage2 = true;
@@ -863,6 +869,57 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         // keyboard to detect the real modifier.
                         #[cfg(windows)]
                         crate::platform::augment_enter_shift(&mut key);
+                        // During paste suppression, discard duplicate key events.
+                        // Timer-based: after right-click copy (no content to match).
+                        // Content-matched: after right-click paste (match against clipboard).
+                        #[cfg(windows)]
+                        {
+                            // Timer-based suppression (copy case)
+                            if paste_suppress_until.is_some() {
+                                let expired = paste_suppress_until.map_or(true, |t| Instant::now() >= t);
+                                if expired {
+                                    paste_suppress_until = None;
+                                } else {
+                                    let is_text_key = matches!(key.code,
+                                        KeyCode::Char(_) | KeyCode::Enter | KeyCode::Tab
+                                    ) && !key.modifiers.contains(KeyModifiers::CONTROL);
+                                    if is_text_key {
+                                        _pending_evt = input.try_read()?;
+                                        continue;
+                                    }
+                                }
+                            }
+                            // Content-matched suppression (paste case)
+                            if paste_suppress_pos < paste_suppress_chars.len() {
+                                let expected = paste_suppress_chars[paste_suppress_pos];
+                                let matched = match key.code {
+                                    KeyCode::Char(c) => c == expected,
+                                    KeyCode::Enter => expected == '\r' || expected == '\n',
+                                    KeyCode::Tab => expected == '\t',
+                                    _ => false,
+                                };
+                                if matched {
+                                    paste_suppress_pos += 1;
+                                    // Skip CRLF: if we matched \r and next is \n, skip it too
+                                    if expected == '\r'
+                                        && paste_suppress_pos < paste_suppress_chars.len()
+                                        && paste_suppress_chars[paste_suppress_pos] == '\n'
+                                    {
+                                        paste_suppress_pos += 1;
+                                    }
+                                    if paste_suppress_pos >= paste_suppress_chars.len() {
+                                        paste_suppress_chars.clear();
+                                        paste_suppress_pos = 0;
+                                    }
+                                    _pending_evt = input.try_read()?;
+                                    continue;
+                                } else {
+                                    // Mismatch — stop suppressing, let event through
+                                    paste_suppress_chars.clear();
+                                    paste_suppress_pos = 0;
+                                }
+                            }
+                        }
                         // Flush pending paste buffer before processing any non-bufferable key.
                         // Bufferable keys are: plain Char, Space, Enter (if pend non-empty), Tab (if pend non-empty).
                         #[cfg(windows)]
@@ -1566,16 +1623,18 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 }
 
                                 KeyCode::Char(' ') => {
-                                    #[cfg(windows)]
-                                    {
-                                        paste_pend.push(' ');
-                                        if paste_pend_start.is_none() {
-                                            paste_pend_start = Some(Instant::now());
+                                    if key.modifiers.is_empty() {
+                                        #[cfg(windows)]
+                                        {
+                                            paste_pend.push(' ');
+                                            if paste_pend_start.is_none() {
+                                                paste_pend_start = Some(Instant::now());
+                                            }
                                         }
-                                    }
-                                    #[cfg(not(windows))]
-                                    {
-                                        cmd_batch.push("send-key space\n".into());
+                                        #[cfg(not(windows))]
+                                        { cmd_batch.push("send-key space\n".into()); }
+                                    } else {
+                                        cmd_batch.push(format!("send-key {}\n", modified_key_name("Space", key.modifiers)));
                                     }
                                 }
                                 // AltGr detection: On Windows, AltGr is reported as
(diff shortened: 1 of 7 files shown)


ToxMox:
I've come across a bug when I use shift while typing that the next key doesn't go through when typing quickly. Can't fix right now so please ignore this PR for now 😄

ToxMox:
Fixed — turns out I type like an animal and don't release Shift before hitting Space after `?`. The SHIFT→ALT mapping was too broad and converted that phantom Shift+Space into ESC+Space. Confirmed with a raw Win32 KEY_EVENT_RECORD viewer — 10/10 times my Space key-down fires before Shift key-up. Mapping is now restricted to Enter only. Details in updated PR description.

ToxMox:
While setting up the ppm plugin manager and testing themes (Catppuccin, Dracula, Tokyo Night), I discovered that all uppercase prefix bindings (`Prefix+I`, `Prefix+U`, `Prefix+M`) silently broke.

**Root cause:** Phase 2's `\x1b[?9001h` changes how Shift+letter is reported — `Char('i') + SHIFT` instead of `Char('I')`.

**Fix** (commit `8b22cf4`):
- `normalize_key_for_binding`: uppercase lowercase letters before stripping SHIFT flag
- `parse_key_string`: preserve original case for single chars — pre-existing bug where CLI `bind-key I` was stored as lowercase `i`
- Early KeyEvent normalization in both client event loop and server `handle_key` — Shift+lowercase → uppercase, covering all copy mode vi keys (G, W, B, H, M, L, F, T, D, V, A, N), tree/session chooser, and confirm prompts

psmux:
Hey @ToxMox, thanks for the careful self-review here. You flagged the uppercase-prefix regression yourself which is exactly the kind of catch we want before merge. Take your time, no pressure on the timeline. When you flip it back to ready and the prefix-uppercase + shift-after-special-char paths are both green, ping me and I will run our internal key matrix (alt sequences, ctrl, AltGr, dead keys, IME compose) against the branch and merge if it holds. Branch is also a bit behind master by now, so a rebase before re-review would help. Cheers.

tarikguney:
Closed for inactivity.

PR #163, opened 2026-03-29
feat: fix mouse coordinate alignment for multi-client sessions

ToxMox:
## Summary

- Fix mouse clicks on status bar tabs not matching rendered tab positions
- Fix all mouse interactions (pane focus, border drag, scroll, PTY forwarding) misaligning when multiple clients of different terminal sizes are connected
- Fix incomplete active pane border highlighting at intersections and nested splits
- Add border hover preview showing the full extent of the separator you'd drag
- Add `pane-border-hover-style` option (default: `fg=yellow`)

## Background

psmux has a client-server architecture where the server computes layout at one "effective" size while each client renders independently at its own terminal size. Mouse events were sent as raw terminal coordinates (`mouse-down X Y`), and the server interpreted them against its effective-size layout. When client and server sizes differed, clicks landed on wrong panes, border drags moved wrong separators, and tab clicks missed entirely.

The tab click bug was the most visible: `update_tab_positions()` on the server used a hardcoded `format!("{}: {} ", idx, name)` with `.len()` (byte count), while the client rendered tabs using configurable format strings, Unicode width, inline styles, separators, and `status-justify` alignment. The positions never matched.

The pane/border alignment bug was the same class of problem applied to every mouse interaction. Different terminal sizes meant `compute_rects()` produced different pixel coordinates on each client vs the server.

## The solution: client-side semantic mouse handling + ratio fallback

Two-layer approach:

**Layer 1 — Semantic commands (precise):** The client detects what was clicked using its own rendered layout and sends semantic commands instead of raw coordinates:

| Interaction | Old command | New command |
|-|-|-|
| Tab click | `mouse-down X Y` | `select-window -t :N` |
| Pane focus | `mouse-down X Y` | `select-pane -t %ID` |
| PTY mouse forwarding | `mouse-down X Y` | `pane-mouse ID BUTTON COL ROW M\|m` |
| Border drag | `mouse-drag X Y` | `split-sizes PATH SIZE1,SIZE2,...` |
| Border drag end | `mouse-up X Y` | `split-resize-done` |
| Scroll | `scroll-up X Y` | `pane-scroll ID up\|down` |
| Hover | `mouse-move X Y` | `pane-mouse ID 35 COL ROW M` |

`pane-mouse` carries 0-based pane-relative coordinates computed by the client from its own `content_chunk`. The server looks up the pane by ID and forwards to the PTY — no coordinate translation needed.

`split-sizes` carries percentage-based sizes computed by the client from its own pixel dimensions. The server sets the sizes on the split node directly — no pixel-to-percentage conversion needed.

**Layer 2 — Ratio fallback (safety net):** Any raw coordinate commands that still reach the server (edge cases, clicks outside all panes) are automatically mapped from client-space to effective-space using `client_sizes[cid]` vs `last_window_area`. ~5 lines per handler, instant improvement for all interactions.

Raw coordinate commands remain as fallback for situations where the client can't determine the target (click outside all panes, no layout data yet).

## Border hover preview

When hovering a separator line, the full extent of the border that would be moved by a drag is highlighted. This makes the tree-based layout structure visible — users can see whether connected-looking lines are one border (parent split) or independent borders (sibling splits) before dragging.

Configurable via `pane-border-hover-style` (default: `fg=yellow`).

## Border highlight fix

Active pane border highlighting was inconsistent at intersection points and along edges shared by nested splits. Each separator segment independently decided its color during recursive rendering, and `fix_border_intersections` changed intersection characters without updating styles. Fixed with a post-processing pass that re-colors all border characters based on adjacency to the active pane rect.

Border drag hit detection also only checked one axis — two vertical separators at the same x position in different panes would match the wrong one. Fixed by adding the border's area extent to the hit check.

## Files changed

| File | What |
|-|-|
| `src/window_ops.rs` | `map_client_coords` ratio fallback, `handle_pane_mouse`, `handle_pane_scroll`, `handle_split_set_sizes`, `handle_split_resize_done` |
| `src/client.rs` | Layout tracking (`collect_pane_rects`, `collect_layout_borders`), all mouse handlers rewritten to semantic commands, border hover preview, border highlight post-processing |
| `src/types.rs` | `PaneMouse`, `PaneScroll`, `SplitSetSizes`, `SplitResizeDone` CtrlReq variants, `pane_border_hover_style` field |
| `src/server/connection.rs` | Parse `pane-mouse`, `pane-scroll`, `split-sizes`, `split-resize-done` |
| `src/server/mod.rs` | Dispatch new variants, serialize hover style in DumpState |
| `src/server/options.rs` | Get/set `pane-border-hover-style` |
| `src/cli.rs` | Register `pane-border-hover-style` option |
| `src/config.rs` | Handle `set-option pane-border-hover-style` |
| `src/format.rs` | Make `pane-border-hover-style` queryable |
| `src/help.rs` | Document default value |

## Design decisions

**Why client-side detection instead of fixing server-side `update_tab_positions()`?** Duplicating the client's rendering logic (format strings, Unicode width, status-justify, inline styles) on the server would create two copies that must stay in sync forever. The client is the entity that renders — it should be the authority on what was clicked.

**Why ratio fallback instead of tmux's forced-smallest-size model?** psmux already supports `window-size latest/largest` for different-sized clients using their full terminals. Forcing the tmux model would be a UX regression. The ratio fallback is a universal safety net that doesn't waste screen space.

**Why both layers?** Semantic commands handle precision interactions (discrete targets, continuous drags). Ratio fallback catches everything else without client changes. This eliminates the transitional period where unconverted interactions would still be broken.

**Why `split-sizes` instead of pixel deltas?** The client computes percentage deltas in its own coordinate space (its own `total_pixels`), avoiding the mismatch where the server's `total_pixels` differs from the client's. The server just sets the sizes — stateless, no DragState needed for semantic drags.

**Why `"_"` for root-level split paths?** The tree path for the root split node is empty (`[]`). An empty string gets lost in whitespace-based command parsing. `"_"` is a clean sentinel that the server parses as `Vec::new()`.

## Test plan

- [ ] Tab clicks — correct tab selected regardless of theme/format
- [ ] Pane focus — click to select pane in multi-pane layout
- [ ] Border drag — vertical and horizontal separators, including root-level splits
- [ ] Border drag in complex layout — only the targeted separator moves, not siblings
- [ ] Two different-sized terminals connected — mouse interactions correct in both
- [ ] Scroll wheel — enters copy mode on shell, forwards to TUI apps
- [ ] Right-click — paste on shell, forward to TUI apps (vim, htop)
- [ ] Copy mode — click to position cursor, drag to select
- [ ] Border hover — yellow highlight shows full extent of separator
- [ ] `set-option pane-border-hover-style "fg=cyan"` — custom hover color works
- [ ] Active pane border highlighting — consistent at intersections, no gaps
- [ ] Mouse move/hover — forwarded to TUI apps correctly
- [ ] Embedded mode (non-client-server) — unchanged behavior

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1798,7 +1798,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                     // Border detection
                                     let mut on_border = false;
                                     if !client_zoomed {
-                                        let tol = 1u16;
+                                        let tol = 0u16;
                                         for (bpath, bkind, bidx, bpos, btotal, bsizes, barea) in &client_borders {
                                             let hit = if bkind == "Horizontal" {
                                                 me.column >= bpos.saturating_sub(tol) && me.column <= bpos + tol
@@ -2020,7 +2020,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 // Detect border hover for visual preview
                                 let mut new_hover: Option<(u16, String, Rect)> = None;
                                 if !client_zoomed {
-                                    let tol = 1u16;
+                                    let tol = 0u16;
                                     for (_, bkind, _, bpos, _, _, barea) in &client_borders {
                                         let hit = if bkind == "Horizontal" {
                                             me.column >= bpos.saturating_sub(tol) && me.column <= bpos + tol


ToxMox:
**Update:** Added one more commit — `fix: prevent focus-follows-mouse on hover events`

During testing I found that `handle_pane_mouse` was unconditionally focusing the target pane for all mouse events including hover (button 35/motion). The original `remote_mouse_motion` only forwarded to the active pane without changing focus. This caused focus-follows-mouse behavior which is not how tmux works — tmux uses click-to-focus.

Fix: `handle_pane_mouse` now only changes `active_path` on actual click events (button 0/1/2 with press=true), not on drag (32) or hover (35).

ToxMox:
**Update:** Added `fix: remove border drag tolerance so first/last pane rows are selectable`

Border hit detection used `tol=1`, which extended the drag zone 1 row/column into adjacent pane content. The entire first and last row of each pane acted as a drag handle instead of allowing text selection. Fixed by setting `tol=0` — only clicks on the exact separator pixel trigger border drag. The hover preview makes separators easy to target without needing tolerance.

Issue #111, opened 2026-03-13
split-window -c "#{pane_current_path}" does not preserve working directory


gtbuchanan:
## Description

`split-window -c "#{pane_current_path}"` does not open the new pane in the current pane's working directory. The `#{pane_current_path}` format variable appears to not be resolved, so the new pane falls back to the default directory.

## Steps to reproduce

1. Add to config:
   ```
   bind '"' split-window -v -c "#{pane_current_path}"
   bind % split-window -h -c "#{pane_current_path}"
   ```
2. `cd` to a non-default directory (e.g., `cd ~/projects`)
3. Split the pane with `prefix + "` or `prefix + %`

## Expected behavior

New pane opens in the same directory as the current pane (e.g., `~/projects`).

## Actual behavior

New pane opens in the default directory instead.

## Environment

- psmux (built from master branch)
- Windows 11
- Git Bash shell

psmux:
Thanks for the report — this turned out to be two separate bugs stacked on top of each other.

**Both fixed in commit [`ff9898c`](https://github.com/marlocarlo/psmux/commit/ff9898c).**

### Bug 1: Format variables never expanded in -c arguments

`split-window -c "#{pane_current_path}"` passed the literal string `#{pane_current_path}` as a directory path — no format expansion happened. Added `expand_format()` calls for the `-c` argument in all four handlers: `split-window`, `split-window -P`, `new-window`, and `new-window -P`.

### Bug 2: pane_current_path always returned the initial directory

Even after fixing expansion, `#{pane_current_path}` returned the spawn-time directory — never reflecting `cd` changes. This is because PowerShell's `cd` (`Set-Location`) only updates `$PWD` internally — it does **not** call Win32 `SetCurrentDirectory()`. psmux reads CWD from the process PEB via `NtQueryInformationProcess`, which still shows the original directory.

**Fix:** psmux now injects a prompt hook that calls `[System.IO.Directory]::SetCurrentDirectory($PWD.ProviderPath)` after every command. This keeps the OS-level CWD in sync with PowerShell's `$PWD`, so `#{pane_current_path}` always returns the correct working directory.

The hook preserves any existing `prompt` function (starship, oh-my-posh, etc.) by wrapping it.

### Usage — your exact config now works

```
bind '"' split-window -v -c "#{pane_current_path}"
bind % split-window -h -c "#{pane_current_path}"
```

### How to update

```
cargo install --git https://github.com/marlocarlo/psmux.git
```

[...]

PR #174, opened 2026-04-02
fix: bind-key new-window -c drops -c flag, losing start directory (#111)

sepiabrown:
`bind-key c new-window -c "#{pane_current_path}"` silently dropped the `-c` argument, causing new windows to always open in the default directory instead of the current pane's working directory.

The #111 fix (ff9898c) correctly added `expand_format()` to the server-side NewWindow handler, but two client-side code paths still discarded the `-c` flag before it could reach the server:

1. `parse_command_to_action()` in commands.rs mapped `new-window ...` to the bare `Action::NewWindow` enum variant regardless of arguments, discarding `-c`, `-n`, and any other flags. This affected both config file parsing and server-side BindKey handling. The fix mirrors the existing `split-window` pattern: when arguments are present, preserve the full command string as `Action::Command(cmd)`.

2. `dispatch_control_command()` in connection.rs (persistent connection bind-key handler) filtered out all `-` prefixed tokens as bind-key's own flags, which stripped `-c` from the command. The fix replaces the greedy filter with explicit flag parsing that only consumes `-T`, `-n`, `-r` as bind-key flags, then passes everything after the key name through verbatim.

Bare `new-window` (no arguments) still returns `Action::NewWindow` as before.

Added 4 unit tests verifying new-window argument preservation for `-c`, `-n`, shell commands, and the bare case. All 649 tests pass.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -187,7 +187,16 @@ pub fn parse_command_to_action(cmd: &str) -> Option<Action> {
     
     match parts[0] {
         "display-panes" | "displayp" => Some(Action::DisplayPanes),
-        "new-window" | "neww" => Some(Action::NewWindow),
+        "new-window" | "neww" => {
+            // If extra flags like -c, -d, -n, -F, -e or a shell command are present,
+            // store as Command to preserve the full argument string (esp. -c for start dir).
+            let has_extra = parts.len() > 1;
+            if has_extra {
+                Some(Action::Command(cmd.to_string()))
+            } else {
+                Some(Action::NewWindow)
+            }
+        }
         "split-window" | "splitw" => {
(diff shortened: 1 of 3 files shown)


psmux:
Hey @sepiabrown, thanks for this excellent contribution! Just merged it.

Really clean fix. You identified both code paths that were silently stripping the `-c` flag, and the approach of mirroring the existing `split-window` pattern for the `parse_command_to_action` change is exactly right. The connection.rs rewrite is a big improvement too, replacing the greedy filter with explicit flag parsing is much more robust and readable.

The 4 unit tests covering bare, `-c`, `-n`, and shell command cases are solid. All 649 tests pass cleanly.

Appreciate you tracing this all the way from the original #111 fix to find the remaining client side gaps. Great work!

PR #181, opened 2026-04-04
fix: resolve Windows run-shell executables to absolute paths

wosuxiongmao:
## Summary
- return absolute executable paths from `resolve_run_shell()` on Windows instead of bare `pwsh`/`powershell`/`cmd` command names
- add explicit fallbacks through `SystemRoot\System32\WindowsPowerShell\v1.0\powershell.exe`, `SystemRoot\System32\cmd.exe`, and `ComSpec`
- reuse the same resolver for local `if-shell` command execution on Windows
- add a Windows regression test that requires `resolve_run_shell()` to return an absolute executable path

## Verification
- `cargo +stable-x86_64-pc-windows-gnu test --target-dir target-test resolve_run_shell_returns_absolute_windows_shell_path -- --nocapture`
- `powershell -ExecutionPolicy Bypass -File tests/test_run_shell.ps1`
  - observed `TOTAL: 11 PASS, 0 FAIL out of 11 tests`

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1,5 +1,7 @@
 use std::io;
 use std::time::Instant;
+#[cfg(windows)]
+use std::path::PathBuf;
 
 use std::io::Write;
 use crate::types::{AppState, Mode, Action, FocusDir, LayoutKind, MenuItem, Menu, Node};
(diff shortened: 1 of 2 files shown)


wosuxiongmao:
Additional Windows validation after the oh-my-codex upstream feedback:

- latest `oh-my-codex` `upstream/dev` now starts `omx team` successfully on native Windows without my OMX patch, so I no longer think OMX needs a companion runtime change for the HUD path
- the remaining broken primitive is still `psmux` foreground `run-shell` itself on stock Windows installs

Stock repro on this machine (`tmux 3.3` from Chocolatey / `psmux 3.3.1`):

```powershell
tmux new-session -d -s rsys
tmux run-shell -t rsys "tmux -V"   # => psmux: program not found
tmux run-shell -t rsys powershell   # => psmux: program not found
```

With this branch:

- `cargo +stable-x86_64-pc-windows-gnu test --target-dir target-test resolve_run_shell_returns_absolute_windows_shell_path -- --nocapture`
- `powershell -ExecutionPolicy Bypass -File tests/test_run_shell.ps1` -> `TOTAL: 11 PASS, 0 FAIL out of 11 tests`

I also re-ran the latest upstream `oh-my-codex` `dev` against a patched `psmux`/`tmux` and `omx team` started cleanly with `status: ok`.

So from my side this PR isolates the remaining Windows `run-shell` / `if-shell` primitive bug rather than depending on an OMX-side workaround.

psmux:
Hey @wosuxiongmao, thanks for this! Really solid contribution.

The if-shell inconsistency was a genuine bug that I missed. Every other code path was already using 
esolve_run_shell() for the fallback chain, but that one local if-shell handler had "pwsh" hardcoded, which would just silently fail on any Windows machine without pwsh installed. Good catch.

The absolute path resolution and the SystemRoot/ComSpec fallbacks are also a nice touch for edge case environments. Merged as is.

Appreciate the thorough testing too. Welcome aboard!

Issue #173, opened 2026-04-02
Agent Teams support broken with Claude Code 2.1.89


r3core:
Agent team panes were working fine until the recent claude code update hit.

<img width="1892" height="349" alt="Image" src="https://github.com/user-attachments/assets/575210da-7bf3-49e2-b983-a9273b25cb8b" />

LiamKarlMitchell:
Interesting, I tried today (first time trying it on windows)
And the command to run team mates has incorrectly quoted the claude path.

Created bug here: https://github.com/anthropics/claude-code/issues/42848

psmux:
Hey! Thanks for reporting this. I have been investigating all the agent teams issues together.

**What I found:** I tested agent teams end to end with psmux and confirmed that the basic flow works correctly. Paths with backslashes and spaces are preserved through the entire send-keys pipeline. The env shim correctly translates Claude Code's bash style `env VAR=val command args` syntax into PowerShell equivalents. I tested with paths like `C:\cctest\a long dir name\another very long name with 5` and everything worked.

**However**, I found a real bug: the PowerShell env shim was only installed when no native `env.exe` was found on PATH. If you have uutils coreutils or similar packages providing `env.exe`, the shim skipped itself entirely. This meant Claude Code's shell-quote escapes (like `\@` in agent IDs) were not properly unescaped, and the full POSIX translation layer was bypassed.

**Fix pushed:** Commit 467b32f makes the env shim always active in PowerShell panes regardless of whether a native env.exe exists. This ensures the POSIX unescape, shebang detection, and proper command delegation always happen.

If there was a specific Claude Code version change (2.1.89) that broke things, it might have been a change in how shell-quote encodes the command or how the agent spawning works on their side. The psmux fix should handle the current behavior correctly.

To get the fix:
```
cargo install --git https://github.com/psmux/psmux.git
```

Let me know if you still see issues after updating!

psmux:
PSMUX I used is the latest commit . Use cargo install --git https://github.com/psmux/


<img width="1047" height="61" alt="Image" src="https://github.com/user-attachments/assets/b452db43-5e73-469f-856b-79bde6f6cda0" />

Exact cmd I ran and it's output.
`$env:CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1; $env:CLAUDE_CODE_USE_POWERSHELL_TOOL=1; $env:CLAUDE_CODE_AUTO_CONNECT_IDE=$false; psmux new -s claude claude`


[...]

PR #184, opened 2026-04-05
Copilot/update docs prerequisites section

LiamKarlMitchell:
Relates to #173 updates claude code docs to include PowerShell 7+ prerequisite


psmux:
Hey @LiamKarlMitchell! Thanks so much for this contribution. You were absolutely right that the PowerShell 7+ requirement needed to be spelled out clearly in the docs, especially for the Claude Code setup.

I went ahead and added the prerequisites section directly to master in commit 1cde3d1, crediting you and referencing both this PR and #173 where you originally discovered the issue. Your debugging journey on that thread was super helpful and it is exactly the kind of insight that saves other users hours of troubleshooting.

Since I incorporated the content directly, I will close this PR, but full credit goes to you in the docs. Really appreciate you taking the time to put this together!

PR #188, opened 2026-04-05
fix(plugin): correct XDG plugin discovery and HOME precedence on Windows

Masterkatt3n:
Fix XDG plugin discovery paths

- Prioritize HOME over USERPROFILE
- Ensure XDG_CONFIG_HOME is scoped under HOME
- Handle flattened plugin paths
- Fix failing test: plugin_discovery_finds_xdg_path

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,9 +1,10 @@
-use std::env;
-use std::cell::RefCell;
 use crossterm::event::{KeyCode, KeyModifiers};
+use std::cell::RefCell;
+use std::env;
+use std::path::PathBuf;
 
-use crate::types::{AppState, Action, Bind};
 use crate::commands::parse_command_to_action;
+use crate::types::{Action, AppState, Bind};
 
 // Track the current config file being parsed (for #{current_file}, #{d:current_file})
 thread_local! {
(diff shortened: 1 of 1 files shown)


psmux:
Hey @Masterkatt3n, thanks a lot for taking the time to fork psmux and put this PR together! Really appreciate the effort, especially for a first PR on GitHub. Welcome!

I took a close look at the changes and ran the full test suite. A few things I noticed:

**The test already passes on master**

The `plugin_discovery_finds_xdg_path` test (and all 709 other tests) pass on the current `master` branch without any changes. So there isn't actually a failing test to fix here.

**Formatting changes make review difficult**

About 80% of the diff (the bulk of the 534 additions / 206 deletions) is reformatting single line expressions into multi line blocks. While expanded formatting can be fine on its own, mixing it into a functional PR makes it very hard to spot and review the actual logic changes.

**XDG_CONFIG_HOME override concern**

The PR adds logic that replaces the user's `XDG_CONFIG_HOME` if it doesn't start with `$HOME`. This goes against the XDG Base Directory spec, which explicitly allows users to set `XDG_CONFIG_HOME` to any path they want (including paths outside their home directory). Overriding that would break setups where people intentionally use a custom config root.

**Inconsistent HOME vs USERPROFILE priority**

The PR changes plugin discovery to prefer `HOME` over `USERPROFILE`, but the `source_file` tilde expansion still uses `USERPROFILE` first. That inconsistency could cause confusing behavior where `~` resolves differently depending on which code path runs.

Given that there's no actual failing test or bug to fix, and the functional changes introduce spec violations and inconsistencies, I'm going to close this one. But please don't let that discourage you! If you run into an actual bug with plugin discovery or XDG paths on your setup, open an issue describing the problem and I'd be happy to dig into it with you.

Thanks again for contributing!

Masterkatt3n:
Hey! Sorry for the delayed follow-up — I wanted to revisit this after spending more time debugging on my side.

I think there may actually be two separate issues involved here:

1. The original XDG/plugin discovery behavior on Windows when users manually configure XDG variables.
2. Process cleanup behavior during certain tests on Windows.

While investigating the XDG path issue, I repeatedly ran the test suite on Windows and noticed that some tests appeared to leave behind child processes after completion. In particular I repeatedly observed lingering `lazygit`, `pstop`/`htop`, `pwsh`, and `conhost` processes after test runs.

After digging further, I found references such as:

`execute_command_string(&mut app, "display-popup -d /some/path lazygit")`

inside the test suite, which seems related to what I was observing. My suspicion is that some PTY/process-spawn tests behave differently on Windows and do not fully clean up child processes in all cases.

Regarding the XDG side specifically:
I understand your point about the spec allowing arbitrary `XDG_CONFIG_HOME` locations, and I agree my earlier PR logic was too aggressive there. However, I do think there may still be an edge case when users manually configure XDG-style layouts on Windows (which is what I’m doing for cross-platform consistency across systems/tools).

One thing I noticed while debugging was that plugin discovery paths could end up mixing separators or joining full plugin identifiers as a single segment on Windows, depending on how the path was constructed.

I’m not claiming there’s a broad issue affecting most users, but I wanted to follow up since the behavior I observed was reproducible on my setup and may still be worth a second look, especially around Windows PTY/process cleanup and path joining.

Thanks again for taking the time to review the original PR.

PR #194, opened 2026-04-09
feat: add scroll-enter-copy-mode option

### LINKED ISSUE #193: Feature request: Option to disable copy-mode on mouse scroll
## Description

When `set -g mouse on` is enabled, scrolling the mouse wheel automatically enters copy-mode. This is sometimes undesirable — users may want to scroll through previous output without entering copy-mode (e.g., letting the terminal emulator like Windows Terminal handle the scrollback natively).

## Current behavior

- Mouse scroll up → automatically enters copy-mode and scrolls the psmux scrollback buffer
- There is no way to disable this behavior via configuration

## Expected behavior

Provide one or more of the following:

1. **A new option** like `set -g scroll-enter-copy-mode off` to prevent mouse scroll from entering copy-mode, passing the scroll event through to the terminal emulator instead.
2. **Support for `WheelUpPane` / `WheelDownPane` key bindings** (like tmux) so users can customize what happens on mouse scroll. For example:
   ```
   bind -n WheelUpPane send-keys -M
   bind -n WheelDownPane send-keys -M
   ```

## Environment

- psmux version: 3.3.1
- Terminal: Windows Terminal
- OS: Windows 11

## Workaround attempted

- `bind -n WheelUpPane send-keys -M` → not recognized, `list-keys` shows no wheel bindings
- `set -g mouse off` → disables all mouse features including scrollback entirely

jun2077681:
## Summary

Adds a new `scroll-enter-copy-mode` session option that controls whether mouse scroll-up at a shell prompt automatically enters copy mode.

- **`set -g scroll-enter-copy-mode on`** (default): Existing tmux-compatible behavior — scroll up enters copy mode
- **`set -g scroll-enter-copy-mode off`**: Scrolls the pane scrollback directly without entering copy mode, similar to native terminal scrolling

## Changes

- New boolean session option `scroll-enter-copy-mode` (default: `on`)
- Extracted `scroll_pane_scrollback()` helper from `scroll_copy_up`/`scroll_copy_down` to eliminate code duplication
- Applied to all 3 scroll entry points: local input (`input.rs`), remote scroll (`window_ops.rs::remote_scroll_wheel`), and pane-targeted scroll (`window_ops.rs::handle_pane_scroll`)

## Usage

```
# In ~/.psmux.conf
set -g mouse on
set -g scroll-enter-copy-mode off
```

Closes #193


psmux:
Hey @jun2077681, thank you so much for this pull request! Really appreciate you not only opening the issue but also putting together a working implementation.

Your approach of extracting `scroll_pane_scrollback()` to directly scroll the buffer when the option is off was genuinely a better design than what I initially had (which was just ignoring the wheel event entirely). That subtle difference makes it so much more usable: users still get their scroll history without being forced into copy mode. Smart call.

I've incorporated your design into the main branch (commit 6bcb9f5) and given you credit in the commit message. The implementation now includes:

- The `scroll_pane_scrollback()` helper extracted from `scroll_copy_up`/`scroll_copy_down` (your idea)
- Option catalog entry (which you also caught that I missed)
- Runtime `set -g` support, `show-options`, format variable `#{scroll-enter-copy-mode}`, and `SetOptionUnset` reset
- 7 regression tests covering the full lifecycle

Since the feature is already merged with your approach baked in, I'll close this PR. But the credit is absolutely yours for the core design. Thanks for making psmux better!

PR #207, opened 2026-04-12
fix(security): add session key auth to TUI mode TCP listener

### LINKED ISSUE #206: Security: TUI mode TCP listener bypasses session key authentication (incomplete fix for #5)
## Description

The session key authentication introduced in #5 was applied to the **server (detached) mode** but not to the **TUI (interactive) mode**. Any local process that can read `~/.psmux/*.port` can connect to a TUI-mode session's TCP control port and execute commands without authentication.

## Root Cause

In **server mode** (`src/server/mod.rs`), the accept loop generates a session key, writes it to a `.key` file, and passes it to `connection::handle_connection()`, which enforces `AUTH <key>` as the first line of every connection:

- Key generation: [`src/server/mod.rs:334-342`](https://github.com/psmux/psmux/blob/3bf380d/src/server/mod.rs#L334-L342)
- Auth enforcement: [`src/server/connection.rs:38-57`](https://github.com/psmux/psmux/blob/3bf380d/src/server/connection.rs#L38-L57)

In **TUI mode** (`src/app.rs`), the accept loop reads the first line directly as a command. No session key is generated, no `.key` file is written, and no authentication check is performed:

- Unauthenticated accept loop: [`src/app.rs:280-480`](https://github.com/psmux/psmux/blob/3bf380d/src/app.rs#L280-L480)

## Impact

Commands accepted without auth in TUI mode include `new-window` and `split-window`, both of which accept an arbitrary command argument that gets passed to `create_window()` → `build_command()` → `spawn_command()`. This allows any local process to execute arbitrary commands as the psmux user.

`source-file` is also accepted unauthenticated, which can load config files containing `run-shell` directives.

### PoC (PowerShell)

```powershell
# 1. Find the port file for any running TUI session
#    (No .key file exists — TUI mode doesn't create one)
$portFile = Get-ChildItem "$env:USERPROFILE\.psmux\*.port" | Select-Object -First 1
$port = Get-Content $portFile

# 2. Send a command without authentication
$tcp = New-Object System.Net.Sockets.TcpClient("127.0.0.1", [int]$port)
$stream = $tcp.GetStream()
$writer = New-Object System.IO.StreamWriter($stream)
$writer.WriteLine('new-window "cmd /c calc.exe"')
$writer.Flush()
$tcp.Close()
```

### Scope

The TCP listener is bound to `127.0.0.1`, so this is a **local-only** attack surface. On single-user desktops the practical risk is limited. The concern is multi-user systems or environments with sandboxed/compromised processes that can make TCP connections and read the port file.

## Suggested Fix

Apply the same authentication pattern from `server/mod.rs` to `app.rs`:

1. Generate a session key in `app.rs` (same logic as `server/mod.rs:334-342`)
2. Write the key to `~/.psmux/<session>.key`
3. Route TCP connections through `connection::handle_connection()` instead of inline command parsing

This would unify both code paths and ensure all TCP connections are authenticated regardless of the mode.

## Additional Hardening Suggestion

Session names are used directly in file paths (`types.rs:774`, `session.rs:93`) without sanitization. A name containing `../` could write `.port`/`.key` files outside `~/.psmux/`. Consider rejecting path separators and `..` sequences in session names. (Low severity — requires local CLI access.)

## Environment

- psmux v3.3.2 (commit `3bf380d`)
- Windows 11 Pro (Build 26200)
- Audited on: 2026-04-12

Mike-7777777:
## Summary

Adds session key authentication to the TUI (interactive) mode TCP control listener, which was missed when auth was introduced for server (detached) mode in #5.

Fixes #206

## What Changed

**`src/app.rs`** — single file, 53 lines added:

| Change | Mirrors |
|--------|---------|
| Generate random session key at startup | `server/mod.rs:334-342` |
| Store key on `AppState` (used by keybinding `send_control_to_port()` calls) | `server/mod.rs:344` |
| Write key to `~/.psmux/<session>.key` | `server/mod.rs:348-349` |
| `#[cfg(windows)]` file permission hardening | `server/mod.rs:357-366` |
| Require `AUTH <key>` on every TCP connection; reject invalid/missing auth | `server/connection.rs:38-57` |

Every change directly mirrors existing server mode code. No new patterns introduced.

## Why

The TCP listener in TUI mode accepted commands like `new-window` and `split-window` (which spawn arbitrary processes) without authentication. Any local process that could read the `.port` file could execute commands as the psmux user. See #206 for full analysis, PoC, and scope.

## Test plan

- [x] `cargo check` — zero errors, zero warnings
- [x] `cargo test` — 2568 tests passed, 0 failed (7 test suites)
- [ ] Manual: start `psmux` in TUI mode → verify `~/.psmux/<session>.key` is created
- [ ] Manual: `psmux send-keys` from another terminal → works (client reads `.key`, sends AUTH)
- [ ] Manual: raw TCP connection without AUTH → rejected with "Authentication required"
- [ ] Manual: keybindings (Prefix+c for new-window, Prefix+% for split) still work

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -285,17 +285,69 @@ pub fn run(terminal: &mut Terminal<CrosstermBackend<io::Stdout>>) -> io::Result<
     let _ = std::fs::create_dir_all(&dir);
     let regpath = format!("{}\\{}.port", dir, app.port_file_base());
     let _ = std::fs::write(&regpath, port.to_string());
+
+    // Generate a random session key for security (mirrors server/mod.rs)
+    let session_key: String = {
+        use std::collections::hash_map::RandomState;
+        use std::hash::{BuildHasher, Hasher};
+        let s = RandomState::new();
+        let mut h = s.build_hasher();
+        h.write_u64(std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap_or_default().as_nanos() as u64);
+        h.write_u64(std::process::id() as u64);
+        format!("{:016x}", h.finish())
+    };
+
+    app.session_key = session_key.clone();
+
+    let keypath = format!("{}\\{}.key", dir, app.port_file_base());
+    let _ = std::fs::write(&keypath, &session_key);
+
+    // Try to set file permissions to user-only (Windows)
+    #[cfg(windows)]
+    {
+        let _ = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(&keypath)
+            .map(|mut f| std::io::Write::write_all(&mut f, session_key.as_bytes()));
+    }
+
     thread::spawn(move || {
         for conn in listener.incoming() {
             if let Ok(stream) = conn {
                 let tx = tx.clone();
+                let session_key_clone = session_key.clone();
                 // Handle each connection in its own thread so rapid-fire
                 // commands (e.g. 200x new-window) don't queue behind each
                 // other on the accept loop.
                 thread::spawn(move || {
                 let mut stream = stream;
-                let mut line = String::new();
                 let mut r = io::BufReader::new(stream.try_clone().unwrap());
+
+                // Authenticate the connection (mirrors server/connection.rs)
+                let _ = stream.set_read_timeout(Some(Duration::from_millis(2000)));
+                let mut auth_line = String::new();
+                if r.read_line(&mut auth_line).is_err() {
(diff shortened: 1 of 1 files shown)


psmux:
Closing this PR. The issue it addresses (#206) is not reproducible. The code path in app.rs that this PR modifies is dead code that is never executed. All TCP connections already go through handle_connection() in connection.rs with full auth enforcement. See the detailed analysis on issue #206.

PR #208, opened 2026-04-12
feat: add support for environment variables in new-session command

### LINKED ISSUE #205: new-session -e is parsed for argv compatibility but does not set session environment
Thank you for this project. I am learning to use Gas Town on Windows and found [gastownhall/gastown#3538](https://github.com/gastownhall/gastown/issues/3538), where the author notes that psmux’s tmux protocol compatibility is insufficient for Gas Town’s session management. Gas Town runs [tmux-oriented tests](https://github.com/gastownhall/gastown/blob/main/internal/tmux/tmux_test.go); failures can come from an incompatible or incomplete tmux implementation, behavioral differences, or missing features. For psmux, I believe the last case applies. I’m opening this issue to track the work and link a future pull request.

## Summary
psmux new-session accepts the -e flag in the same way as tmux’s getopt-style CLI (the next argument is consumed as the value). Today that value is not applied: it is only skipped so scripts that pass -e VARIABLE=value do not mis-parse the following arguments (e.g. session name or shell command). Session environment from -e is not implemented yet.

I believe skipping the value was a compatibility measure: recognize the flag and consume its argument so users and scripts do not break, without implementing full session-env plumbing at the time. See #21 

## Expected behavior (tmux)
Per tmux, -e takes the form VARIABLE=value and may appear multiple times. Each pair should set an environment variable for the newly created session (session environment inherited by panes), analogous to tmux’s semantics for new-session -e.

## Actual behavior
The parser advances past -e and its value so argv stays aligned with tmux-style invocations.
The VARIABLE=value string is discarded; nothing is forwarded into session/pane environment setup.

## Scope / related entry points

The same “skip value” pattern exists in more than one place (top-level CLI new-session, nested new-session from inside a session, etc.). Implementing the feature should collect assignments, thread them through the internal server spawn path, and merge into the same session environment store used for set-environment / pane spawns.

## Acceptance criteria (suggestion)
- `psmux new-session -e FOO=bar ...` results in FOO being part of session environment for new panes (and visible where session env is exposed, e.g. format expansion / `show-environment`  behavior as appropriate).
- Invalid VARIABLE=value forms produce a clear error instead of silently dropping data.

pbolduc:
Here is my attempt at adding this feature. I dont really code in Rust, so if there is any AI slop, please point it out.  I didn't like these big blocks added to the `-e` parsing in [main.rs](https://github.com/pbolduc/psmux/blob/9f9d59abc29ea70e76c3af2bdb25d5cf880eecdb/src/main.rs#L509-L520), [commands.rs](https://github.com/pbolduc/psmux/blob/9f9d59abc29ea70e76c3af2bdb25d5cf880eecdb/src/commands.rs#L1582-L1591) and [connection.rs](https://github.com/pbolduc/psmux/blob/9f9d59abc29ea70e76c3af2bdb25d5cf880eecdb/src/server/connection.rs#L1813-L1822). Please advise of any changes you would like to see.

- Enhanced the `new-session` command to accept `-e VAR=value` flags for setting environment variables.
- Implemented parsing for environment variable assignments and integrated them into the session environment.
- Updated relevant functions to handle the new session environment variables, ensuring they are applied after loading the configuration.
- Added tests for parsing and collecting environment variable arguments to ensure correctness.

Fixes #205

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1564,11 +1564,12 @@ fn execute_command_string_single(app: &mut AppState, cmd: &str) -> io::Result<()
         }
         "new-session" | "new" => {
             // Issue #200: create a new session from inside a running session.
-            // Parse flags: -s name, -d (detached), -n windowname, -c startdir
+            // Parse flags: -s name, -d (detached), -n windowname, -c startdir, -e VAR=val
             let mut session_name: Option<String> = None;
             let mut detached = false;
             let mut window_name: Option<String> = None;
             let mut start_dir: Option<String> = None;
+            let mut session_env: Vec<(String, String)> = Vec::new();
             {
                 let mut i = 1;
                 while i < parts.len() {
@@ -1578,7 +1579,17 @@ fn execute_command_string_single(app: &mut AppState, cmd: &str) -> io::Result<()
                         "-c" => { i += 1; if i < parts.len() { start_dir = Some(parts[i].trim_matches('"').to_string()); } }
                         "-d" => { detached = true; }
                         "-A" | "-D" | "-E" | "-P" | "-X" => { /* compatibility flags, ignored */ }
-                        "-e" | "-F" | "-f" | "-t" | "-x" | "-y" => { i += 1; /* skip value */ }
+                        "-e" => {
+                            i += 1;
+                            match crate::util::parse_new_session_e_value_token(parts.get(i).copied()) {
+                                Ok(p) => session_env.push(p),
+                                Err(e) => {
+                                    app.status_message = Some((format!("psmux: {}", e), Instant::now()));
+                                    return Ok(());
+                                }
+                            }
+                        }
+                        "-F" | "-f" | "-t" | "-x" | "-y" => { i += 1; /* skip value */ }
                         _ => {}
                     }
                     i += 1;
@@ -1619,7 +1630,7 @@ fn execute_command_string_single(app: &mut AppState, cmd: &str) -> io::Result<()
             // Try to claim a warm server first (fast path)
             let warm_disabled = std::env::var("PSMUX_NO_WARM").map(|v| v == "1" || v == "true").unwrap_or(false)
                 || crate::config::is_warm_disabled_by_config();
-            let claimed_warm = if !warm_disabled && start_dir.is_none() {
+            let claimed_warm = if !warm_disabled && start_dir.is_none() && session_env.is_empty() {
                 let warm_base = if let Some(ref sn) = app.socket_name {
                     format!("{}____warm__", sn)
                 } else {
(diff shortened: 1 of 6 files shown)


psmux:
Hey @pbolduc, really appreciate you taking the time to write this PR! Your code quality is genuinely impressive, especially for someone who says they do not normally code in Rust.

I had already started working on a fix for #205 before I saw your PR (commit 38d7cfa), but after reviewing your approach I immediately saw several things you did better than my initial implementation:

**What I adopted from your PR (commit 9926b85):**

- **Utility functions in `util.rs`**: Your `parse_env_assignment()`, `parse_new_session_e_value_token()`, `collect_server_session_env_args()`, and `merge_session_env_into_app()` are much cleaner than my original inline parsing. I adopted all of them.
- **`Vec<(String, String)>` type**: Much better than my original `Vec<String>` approach. Pre-parsing the key/value pairs gives type safety and avoids repeated `find('=')` calls.
- **`is_valid_env_var_name()` validation**: My first pass silently ignored malformed values. Your validation with clear error messages (`"invalid -e: expected VARIABLE=value"`, `"invalid -e: invalid environment variable name"`) is the right behavior.
- **Error reporting in `connection.rs`**: Your pattern of storing `session_env_parse_err` and reporting it to both persistent and non-persistent clients is exactly right.
- **Tests**: Adopted your `test_new_session_env.rs` and all the `util::tests` for parsing, `--` boundary handling, and duplicate key behavior.

**One timing difference I kept from my implementation:**

Your PR applies env vars AFTER `load_config()`:
```rust
crate::config::populate_default_bindings(&mut app);
load_config(&mut app);
// new-session -e: apply after config so CLI overrides set-environment from config
crate::util::merge_session_env_into_app(&mut app, &session_env);
```

The issue is that `spawn_warm_pane()` runs before `load_config()`, so the warm pane's shell process would start without the env vars. In my implementation, env vars are applied BEFORE `spawn_warm_pane()`, and when `-e` is specified, warm server claiming is bypassed entirely to force a cold start. This ensures the first pane always inherits the env vars.

Your idea of "CLI overrides config" is a valid design choice though. Since we skip warm pane entirely when `-e` is present (cold start path), the config loads first and then gets overridden by the env vars naturally in the `create_window()` path.

Since I have already merged the core ideas from your PR into master, I will close this PR. But seriously, thank you for the thorough contribution. The validation, utility functions, and tests made the implementation significantly better. Looking forward to more contributions from you!

psmux:
Closing as the improvements have been incorporated into master (38d7cfa + 9926b85). Thank you for the excellent contribution!

pbolduc:
Completed by maintainer in another PR.

Issue #202, opened 2026-04-11
switch-client is non-functional (returns exit 0 but never switches)


tarikguney:
## Summary

`switch-client` is registered in `list-commands` and accepts all standard tmux flags (`-t`, `-c`, `-n`, `-p`, `-l`), but it **never actually switches the client session**. It always returns exit code 0 silently.

This was discovered while building [claude-watch](https://github.com/tarikguney/claude-watch), a TUI dashboard that monitors Claude Code sessions across psmux windows and needs to programmatically navigate to a specific session/window.

## Environment

- **psmux version:** 3.3.2 (installed via WinGet)
- **OS:** Windows 11 (10.0.26200)
- **Terminal:** Windows Terminal
- **Shell:** PowerShell 7+

## Steps to Reproduce

```powershell
# Start psmux with two sessions
psmux new-session -d -s alpha
psmux new-session -s beta

# From within the 'beta' session, try to switch to 'alpha'
psmux switch-client -t "alpha"

# Verify: client is still on 'beta'
psmux list-clients
# Output: /dev/pts/XXXXX: beta: ... (unchanged)
```

## Detailed Testing

All of these return exit code 0 but have no observable effect (verified via `list-clients` before/after each):

| Command | Result |
|---------|--------|
| `psmux switch-client -t session` | No-op |
| `psmux switch-client -t session:window` | No-op |
| `psmux switch-client -c /dev/pts/XXXXX -t session` | No-op |
| `psmux -S /tmp/psmux-XXXXX/default switch-client -t session` | No-op |
| `psmux run-shell "psmux switch-client -t session"` | No-op |
| `bind-key -n F5 switch-client -t session` + manual F5 press | No-op |

Additionally:
- `select-window -t other_session:window` (cross-session) has no effect
- `attach-session -t session` from within psmux only prints the version string (`psmux 3.3.2`) and exits without attaching (with or without `-d`, with or without `$TMUX` env set)
- `choose-session` and `choose-tree` from CLI show no visible overlay

## What DOES Work

The **keyboard prefix path** works perfectly: `Ctrl+B, s` (prefix + choose-session) opens the interactive session picker and switching works. This means the server-side session switching logic is implemented and functional — the issue is that `switch-client` (and related CLI commands) never reach or trigger that logic.

## Expected Behavior

`switch-client -t <target>` should switch the current client (identified via `$TMUX` environment variable) to display the target session, matching tmux behavior.

## Context

[...]

PR #214, opened 2026-04-13
fix: route switch-client CLI to source session server (#202)

tarikguney:
## Summary

Fixes the remaining CLI path issue from #202: `psmux switch-client -t <session>` returns exit 0 but does not switch when invoked from the command line (works from `prefix + :`).

**Root cause:** The global `-t` parser in `main.rs` sets `PSMUX_TARGET_SESSION` for all commands, routing `send_control()` to the target session's server. But for `switch-client`, `-t` means "switch TO this session" (destination), not "route to this session's server" (source). The destination server receives the command and responds "already on that session" — doing nothing.

**Fix:** Skip setting `PSMUX_TARGET_SESSION` when the command is `switch-client` or `switchc`, so the existing TMUX-based fallback resolves the current (source) session for routing. `PSMUX_TARGET_FULL` still carries the destination for the server-side handler in `connection.rs`.

## Repro (before fix)

```powershell
# Inside a psmux pane (session "claude-watch"):
psmux switch-client -t convserv    # exit 0, no switch
# prefix + : → switch-client -t convserv  # works
```

## Test plan

- [ ] `psmux switch-client -t <other-session>` from CLI switches the terminal
- [ ] `psmux switch-client -n` / `-p` / `-l` from CLI still work
- [ ] `prefix + :` → `switch-client -t <session>` still works
- [ ] Other commands with `-t` (e.g. `select-window -t session:window`) still route correctly
- [ ] Existing Rust unit tests pass (`cargo test`)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -124,7 +124,14 @@ fn run_main() -> io::Result<()> {
             // directly. Otherwise (e.g. -t %2, -t :1.0) fall through to
             // the TMUX env var resolution below so we connect to the right
             // server when invoked from inside a psmux pane.
-            if has_explicit_session {
+            //
+            // Exception: for switch-client, -t is the DESTINATION session,
+            // not the server to route the command to. Skip setting
+            // PSMUX_TARGET_SESSION so the TMUX-based fallback below resolves
+            // the current (source) session for routing. PSMUX_TARGET_FULL
+            // still carries the destination for the server handler.
+            let is_switch_client = args.iter().any(|a| a == "switch-client" || a == "switchc");
+            if has_explicit_session && !is_switch_client {
                 env::set_var("PSMUX_TARGET_SESSION", &port_file_base);
             }
         }


psmux:
Hey @tarikguney, what a great contribution. You didn't just report the bug, you dug into the source, traced the exact routing path, identified the precise line where the logic went wrong, and delivered a clean minimal fix. That kind of investigative work is genuinely rare and I want you to know it is really appreciated.

Your root cause analysis was spot on. The global -t parser in main.rs treats -t the same for every command, which is correct for things like select-window or send-keys where -t means "route to this session's server." But for switch-client, -t means "switch TO this session," so the command was being sent to the destination server, which unsurprisingly responded with nothing because it was already on that session.

I reviewed the change carefully, traced the full execution path from CLI invocation through send_control into the server handler, confirmed that PSMUX_TARGET_FULL still carries the destination correctly for the server side, and ran the full test suite (910 unit tests, all passing) before merging. The fix is exactly right and the scope is just what it needs to be.

Merged in df54346 as part of commit cb28ab1.

Looking forward to seeing what claude-watch does with proper programmatic session switching. That project sounds really cool. Thank you again for pushing this all the way through.

psmux:
Hey @tarikguney, what a great contribution. You didn't just report the bug, you dug into the source, traced the exact routing path, identified the precise line where the logic went wrong, and delivered a clean minimal fix. That kind of investigative work is genuinely rare and I want you to know it is really appreciated.

Your root cause analysis was spot on. The global -t parser in main.rs treats -t the same for every command, which is correct for things like select-window or send-keys where -t means "route to this session's server." But for switch-client, -t means "switch TO this session," so the command was being sent to the destination server, which unsurprisingly responded with nothing because it was already on that session.

I reviewed the change carefully, traced the full execution path from CLI invocation through send_control into the server handler, confirmed that PSMUX_TARGET_FULL still carries the destination correctly for the server side, and ran the full test suite (910 unit tests, all passing) before merging. The fix is exactly right and the scope is just what it needs to be.

Merged in df54346 as part of commit cb28ab1.

Looking forward to seeing what claude-watch does with proper programmatic session switching. That project sounds really cool. Thank you again for pushing this all the way through.

PR #219, opened 2026-04-15
Fix list-panes -s not iterating all windows when -F is used

### LINKED ISSUE #218: list-panes -s only returns panes from one window instead of all windows in session
## Description

`list-panes -s -t <session>` should list all panes across all windows in a session (per tmux spec and psmux's own docs), but it only returns panes from a single window.

## Steps to Reproduce

1. Create a session with multiple windows, each containing at least one pane:
   ```
   psmux new-session -d -s test
   psmux new-window -t test
   psmux new-window -t test
   ```

2. Run `list-panes -s`:
   ```
   psmux list-panes -s -t test -F "#{pane_pid} #{window_index}:#{window_name}"
   ```

3. Compare with querying each window individually:
   ```
   for w in 1 2 3; do
     psmux list-panes -t "test:$w" -F "#{pane_pid} #{window_index}:#{window_name}"
   done
   ```

## Observed Behavior

`list-panes -s` returns panes from only one window. For example, with a session that has 4 windows:

```
# -s flag: returns 1 pane (only window 1)
$ psmux list-panes -s -t myproject -F "#{pane_pid} w#{window_index}"
11111 w1

# Per-window query: returns all 4 panes
$ for w in 1 2 3 4; do psmux list-panes -t "myproject:$w" -F "#{pane_pid} w#{window_index}"; done
11111 w1
22222 w2
33333 w3
44444 w4
```

Full test across multiple sessions:

| Session | Windows | `-s` returns | Per-window returns |
|---|---|---|---|
| project-a | 2 | 1 pane (w2) | 2 panes |
| project-b | 1 | 2 panes (w1) | 2 panes (correct — single window) |
| project-c | 2 | 1 pane (w2) | 2 panes |
| project-d | 4 | 1 pane (w1) | 4 panes |
| project-e | 2 | 2 panes (w1) | 3 panes |

Sessions with a single window return correct results. Multi-window sessions are always missing panes.

## Expected Behavior

`list-panes -s` should iterate all windows in the session and return every pane, matching tmux behavior:

> **tmux(1)**: "If **-s** is given, target is a session (or the current session)." — meaning all panes in all windows of that session are listed.

This is also consistent with psmux's own [args reference](https://github.com/psmux/psmux/blob/master/docs/tmux_args_reference.md) which documents `-s` as the "session" flag.

## Environment

- psmux 3.3.2
- Windows 11

## Workaround

Query `list-windows` first, then `list-panes -t session:window` for each window individually.

tarikguney:
## Summary

- `list-panes -s -t <session> -F "..."` only returned panes from one window instead of all windows in the session
- The `-s` (session scope) flag was handled correctly in the non-format code path but was missing from the `-F` (format) branch — only `-a` was checked there
- Fixed both handler locations in `connection.rs` to check `session_scope` alongside `all`

## Root Cause

In the first handler (~line 886), when `-F` is provided:
```rust
// Before: only -a triggers ListAllPanesFormat
if all { ... }

// After: -s also triggers it
if all || session_scope { ... }
```

The second handler (~line 2185) had the same issue and was also missing the `session_scope` variable entirely.

## Test Plan

- [ ] Create a session with multiple windows: `psmux new-session -d -s test && psmux new-window -t test && psmux new-window -t test`
- [ ] Run `psmux list-panes -s -t test -F "#{pane_pid} #{window_index}"` — should return panes from all 3 windows
- [ ] Run `psmux list-panes -s -t test` (no `-F`) — should also return all panes (was already working)
- [ ] Run `psmux list-panes -t test` (no `-s`) — should return panes from the current window only (unchanged behavior)

Fixes #218

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -883,7 +883,7 @@ match cmd {
         let session_scope = args.iter().any(|a| *a == "-s");
         let (rtx, rrx) = mpsc::channel::<String>();
         if let Some(fmt_str) = fmt {
-            if all {
+            if all || session_scope {
                 let _ = tx.send(CtrlReq::ListAllPanesFormat(rtx, fmt_str));
             } else {
                 let _ = tx.send(CtrlReq::ListPanesFormat(rtx, fmt_str));
@@ -2180,9 +2180,10 @@ fn dispatch_control_command(
         }
         "list-panes" | "lsp" => {
             let all = args.iter().any(|a| *a == "-a");
+            let session_scope = args.iter().any(|a| *a == "-s");
             let format_str = args.windows(2).find(|w| w[0] == "-F").map(|w| w[1].to_string());
             let (rtx, rrx) = mpsc::channel::<String>();
-            if all {
+            if all || session_scope {
                 if let Some(fmt) = format_str {
                     let _ = tx.send(CtrlReq::ListAllPanesFormat(rtx, fmt));
                 } else {


psmux:
Hey @tarikguney, this was a clean and precise fix. You identified both handler locations in `connection.rs` that needed the `session_scope` check, and the change is minimal and correct.

I verified the bug independently before looking at your fix: a 3 window session with `list-panes -s -t session -F "#{pane_pid} w#{window_index}"` returned only 1 pane. After applying your change, it correctly returns all 3, matching per window queries exactly.

All 5,190 tests pass with zero regressions. Merged and you have co-author credit on the commit.

Thanks for the contribution!

Issue #221, opened 2026-04-15
installed run told me: run-shell: program not found


H0nGzA1:

psmux:
You missed attaching a screenshot?

psmux:
Hey @H0nGzA1, thanks for reporting this!

It looks like both your screenshot uploads in this issue (the body and your comment) failed to upload, so I can only go off the title: **"run-shell: program not found"**.

I ran a comprehensive test suite against `run-shell` covering all code paths and could not reproduce this issue. Here is what I tested (27 tests, all passing):

**CLI path (main.rs dispatch):**
- `psmux run-shell "echo hello"` works
- `psmux run` alias works
- No args shows correct usage message
- Background mode (`-b`) works
- PowerShell pipelines, quoted commands, exit code forwarding all work
- Nonexistent commands return a proper error (not a crash)

**TCP server path (connection.rs):**
- `run-shell` via raw TCP returns correct output
- `run` alias via TCP works
- Background mode via TCP is accepted
- Nonexistent commands return PowerShell error messages

**Config file integration:**
- `run-shell` in psmux.conf executes correctly
- A bad `run-shell` in config (calling nonexistent program) does NOT crash the session, and subsequent config lines still apply

**TUI visual verification:**
- Live TUI session responds to `run-shell` via TCP
- Pane output, split windows all functional during live TUI

Could you try re uploading those screenshots? I need to see the exact error you are getting. A few questions that would help narrow this down:

1. **What command did you run?** Did you just run `psmux` (which loads your config file) or did you run `psmux run-shell <something>` explicitly?
2. **Do you have a `.psmux.conf` or `.tmux.conf`** in your home directory? If so, does it have any `run-shell` or `run` lines? For example, tmux plugin manager (TPM) configs often include `run '~/.tmux/plugins/tpm/tpm'` which would not exist on Windows.
3. **Which shells are available on your system?** Can you run `where pwsh` and `where powershell` in a command prompt and share the output?

The error format `run-shell: <message>` comes from psmux when the underlying shell binary itself cannot be found. On a standard Windows system, psmux falls back through `pwsh` then `powershell` then `cmd.exe`, so at least one should always be available.

Looking forward to your screenshots and details so we can figure this out!

PR #222, opened 2026-04-15
fix: normalize Windows paths in run-shell to avoid mixed slashes

H0nGzA1:
## Summary

On Windows, `USERPROFILE` contains backslashes (e.g. `C:\Users\...`) but psmux config files often use forward slashes (e.g. `~/.psmux/...`). 

The `expand_run_shell_path` function was creating mixed paths like `C:\Users\user/.psmux/...` which PowerShell cannot resolve, resulting in:

```
run-shell: No such file or directory
```

## Fix

Normalize the home directory and all path separators to backslashes on Windows before executing shell commands.

## Test

Before fix: `C:\Users\user/.psmux/plugins/ppm/scripts/install_plugins.ps1` (broken - mixed slashes)
After fix: `C:\Users\user\.psmux\plugins\ppm\scripts\install_plugins.ps1` (valid Windows path)

## Checklist

- [x] Build passes on Windows
- [x] run-shell commands with ~ paths work correctly on Windows

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -381,7 +381,25 @@ pub fn parse_config_line(app: &mut AppState, line: &str) {
                     let first = bytes[0];
                     let last = bytes[bytes.len() - 1];
                     if (first == b'\'' && last == b'\'') || (first == b'"' && last == b'"') {
-                        trimmed[1..trimmed.len()-1].to_string()
+                        // Strip outer quotes AND unescape interior sequences (tmux-compatible)
+                        // \" becomes " and \\ becomes \
+                        let inner = &trimmed[1..trimmed.len()-1];
+                        let mut result = String::with_capacity(inner.len());
+                        let chars: Vec<char> = inner.chars().collect();
+                        let mut j = 0;
+                        while j < chars.len() {
+                            if chars[j] == '\\' && j + 1 < chars.len() {
+                                let next = chars[j + 1];
+                                if next == '"' || next == '\\' {
+                                    result.push(next);
+                                    j += 2;
+                                    continue;
+                                }
+                            }
+                            result.push(chars[j]);
+                            j += 1;
+                        }
+                        result
                     } else {
                         cmd
                     }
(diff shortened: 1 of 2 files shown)


H0nGzA1:
https://github.com/psmux/psmux/issues/221

psmux:
Thanks for the contribution @H0nGzA1, but I cannot merge this.

I ran a comprehensive 18 test suite (CLI, TCP, config file, TUI) specifically targeting the scenarios you describe, and **all paths work correctly without this change**:

**Forward slash tilde paths (`~/.psmux/plugins/...`):**
- CLI: works
- TCP server: works
- Config file (`psmux.conf`): works
- PPM style plugin path: works

**Backslash tilde paths (`~\.psmux\plugins\...`):**
- CLI: works
- TCP server: works
- Config file: works

**set-hook with escaped quotes and paths with spaces:**
- Config file path: works
- CLI path: works

The specific claim that `expand_run_shell_path` produces broken mixed paths like `C:\Users\user/.psmux/...` is misleading. Yes, the expanded path uses mixed separators, but PowerShell handles both `/` and `\` interchangeably in file paths. This is not a bug.

More importantly, this PR introduces a **real bug**: the blanket `cmd.replace('/', "\\")` in `util.rs` would break any `run-shell` command containing forward slashes that are NOT file paths. For example:

`psmux run-shell "curl https://example.com/api/v1"`

would become:

`psmux run-shell "curl https:\\example.com\api\v1"`

which is completely broken.

Also: this PR (#222) and PR #223 contain **identical diffs** despite having different titles and descriptions. That is unusual and makes it hard to review your contributions clearly. Please submit one PR per fix.

I will close this PR. If you are experiencing a genuine issue, please re open issue #221 with your screenshots (both uploads in that issue failed) and include:
1. Your exact config file content
2. The exact error output
3. The output of `where pwsh` and `where powershell`

psmux:
Closing: claimed bug not reproducible, and the proposed fix would break URL handling in run-shell commands. See review above.

PR #225, opened 2026-04-15
fix: bounded frame channel to preserve fast typing

### LINKED ISSUE #224: Fast typing drops intermediate frames — characters not rendered while cursor advances
## Description

When typing quickly in a psmux pane, the cursor advances correctly but characters are not rendered on screen. Typing slowly works fine — each character appears as expected. This is a regression that did not exist before.

## Steps to Reproduce

1. Open a psmux session
2. Type quickly in any shell (bash, pwsh, etc.)
3. Observe that the cursor moves forward with each keystroke, but some or all characters behind the cursor fail to render
4. Typing slowly renders every character correctly

## Root Cause

Commit 694156e ("fix: bounded frame push prevents copy-mode scroll memory leak") replaced the unbounded `mpsc::channel` per client with a single-slot `Arc<Mutex<Option<String>>>`. The `push_frame()` function now **overwrites** any unconsumed frame in the slot:

```rust
// src/types.rs — push_frame()
slots.retain(|(_, slot)| {
    // overwrites any pending frame — only "latest" is kept
    *pending = Some(frame.to_string());
    ...
});
```

And the writer thread in `connection.rs` polls the slot on a 5ms timeout:

```rust
// 2. Check the frame slot — take the latest pushed frame
let frame = frame_lock.lock().ok().and_then(|mut slot| slot.take());
```

During fast typing:
1. Each keystroke triggers a PTY update → server generates a new frame
2. If multiple frames arrive within the 5ms polling window, intermediate frames are **overwritten and never delivered** to the client
3. The client only receives the "latest" frame, which may have the correct cursor position but intermediate rendering states are lost
4. The client-side skip logic (`dump_buf == prev_dump_buf`) can further compound this if timing causes a stale comparison

The previous unbounded channel design queued all frames, guaranteeing each one was delivered and rendered. The single-slot design trades that guarantee for bounded memory — fixing the copy-mode scroll leak but breaking fast typing.

## Expected Behavior

All typed characters should render on screen regardless of typing speed.

## Suggested Fix

Consider a bounded channel (e.g., `mpsc::sync_channel` with a small capacity like 8-16) instead of a single-slot overwrite. This would:
- Still bound memory (unlike the original unbounded channel)
- Allow a small burst of frames to queue without being dropped
- Only drop frames under sustained high-throughput scenarios (like the scroll case the fix was targeting)

Alternatively, the frame slot could use a dirty flag + condvar to wake the writer thread immediately on each new frame rather than relying on the 5ms poll interval, reducing the window for overwrites.

## Environment

- Windows 11
- psmux latest (post-694156e)

tarikguney:
## Summary

- Replaces the single-slot `Arc<Mutex<Option<String>>>` frame push (from 694156e) with a bounded `sync_channel` (capacity 16)
- Fixes fast typing regression where cursor advances but characters are not rendered because intermediate frames were overwritten before the client consumed them
- Memory remains bounded to `O(FRAME_CHANNEL_CAPACITY * clients)` — still prevents the original unbounded growth during rapid copy-mode scrolling

## What changed

**`src/types.rs`**: `FrameSlot` → `FrameChannel` backed by `mpsc::sync_channel(16)`. `push_frame()` uses `try_send()` — if the channel is full (sustained scroll burst), the frame is skipped rather than blocking, so the server is never stalled. Dead channels are pruned via `Disconnected`.

**`src/server/connection.rs`**: Writer thread now locks the channel receiver once and drains all queued frames per loop iteration via `try_recv()`, delivering batched frames without the 5ms polling delay between them.

## Why this approach

| Approach | Memory bound | Fast typing | Scroll burst |
|---|---|---|---|
| Original unbounded channel | ❌ O(frames) | ✅ all frames queued | ❌ 1 GB leak |
| Single-slot (694156e) | ✅ O(1) per client | ❌ drops intermediate frames | ✅ bounded |
| **Bounded channel (this PR)** | ✅ O(16) per client | ✅ 16-frame burst buffer | ✅ graceful skip when full |

Fixes #224

## Test plan

- [ ] Type quickly in a psmux pane — all characters should render without gaps
- [ ] Run `test_scroll_memory.ps1` — memory should stay bounded (regression test from 694156e)
- [ ] Attach multiple clients and type fast in each — no cross-client interference

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -95,18 +95,20 @@ if line.trim() == "PERSISTENT" {
     let mut ws_bg = write_stream.try_clone().unwrap();
     let (resp_tx, resp_rx) = mpsc::channel::<mpsc::Receiver<String>>();
 
-    // Register a frame slot for server-pushed frames (event-driven rendering).
-    // The slot holds at most ONE pending frame — push_frame() overwrites any
-    // unconsumed frame, bounding memory to O(1) per client instead of O(frames).
-    let frame_slot = crate::types::register_frame_slot(client_id);
+    // Register a bounded frame channel for server-pushed frames (event-driven
+    // rendering).  The channel queues up to FRAME_CHANNEL_CAPACITY frames,
+    // allowing short bursts (e.g. fast typing) to be delivered without dropping
+    // intermediate states, while still bounding memory for sustained throughput
+    // scenarios (e.g. rapid scroll in copy mode).
+    let frame_chan = crate::types::register_frame_channel(client_id);
 
     // Register a directive channel for queued directives (e.g. SWITCH).
-    // Directives use a separate mpsc channel so they cannot be overwritten
-    // by frame pushes (which always replace the previous pending frame).
+    // Directives use a separate mpsc channel so they are never affected
+    // by frame channel backpressure.
     let directive_rx = crate::types::register_directive_channel(client_id);
 
     std::thread::spawn(move || {
-        let (frame_lock, _frame_cvar) = &*frame_slot;
+        let frame_rx = frame_chan.rx.lock().unwrap();
         loop {
             // 0. Check for queued directives (non-blocking) — these take priority
             while let Ok(directive) = directive_rx.try_recv() {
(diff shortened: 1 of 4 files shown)


psmux:
Hey @tarikguney, this is merged now. Really appreciate you taking the time to not only identify the root cause but also implement the fix yourself.

Your analysis of the single slot overwrite behavior was spot on. The bounded `sync_channel(16)` approach is exactly the right tradeoff: it preserves intermediate frames during normal typing while still capping memory for the pathological scroll bursts that the original fix was targeting. Clean, well documented code too.

I also noticed you included the pane died/pane exited hooks fix for remain on exit in the second commit, which is a nice bonus. Thanks again for the solid contribution!

PR #228, opened 2026-04-15
fix: fire pane-died/pane-exited hooks with remain-on-exit

### LINKED ISSUE #227: pane-died / pane-exited hooks with remain-on-exit (tmux parity)
## Problem statement

In **tmux**, when a pane's child process exits and **`remain-on-exit`** is enabled, the pane stays on screen in a "dead" state and hooks such as **`pane-died`** (and related events) still run. Automation and plugins often depend on that—for example a `pane-died` hook that runs `respawn-pane` after checking `#{pane_dead}`.

In **psmux**, **`pane-died`** / **`pane-exited`** hooks appear to run only when the reaper **removes** a pane from the layout. If **`remain-on-exit`** is on, the child exits but the pane node may stay, so **no prune occurs** and the hooks **may never fire**. That breaks workflows and external tests that assume tmux-like behavior.

## Commands to reproduce

**`display-message` is shown on an attached client** (status / message area). A detached session often has no client, so you will not see the message until you attach—or use a side effect that does not need a TUI.

### A. See `display-message` (requires a client)

1. Start a **detached** session whose pane exits quickly:

```powershell
psmux new-session -d -s repro-session -n test "Start-Sleep -Seconds 2"
```

2. Set **`remain-on-exit`** and a **`pane-died`** hook:

```powershell
psmux set-option -t repro-session -g remain-on-exit on
psmux set-hook -t repro-session pane-died 'display-message "PANE-DIED-FIRED"'
```

3. **Attach** so messages are visible:

```powershell
psmux attach -t repro-session
```

4. Wait until the initial **`Start-Sleep`** has finished (at least ~3 seconds after step 1, or wait after attach if you attached late).

5. **Expected (tmux-like):** shortly after the pane's process exits, you should see **`PANE-DIED-FIRED`** in the message/status area. **Actual:** that message may never appear even though the pane is dead with **`remain-on-exit`**.

Detach when finished (e.g. prefix + `d`, or your documented detach command).

### B. Observe without relying on the UI (detached-friendly)

```powershell
psmux set-hook -t repro-session pane-died 'run-shell -b "echo died >> $env:TEMP\psmux-pane-died.txt"'
```

After the sleep exits, inspect that file for new content.

Optional: repeat the same steps against **tmux 3.x** with the same session/hook names for comparison.

## Acceptance criteria

1. When a pane's child process exits and **`remain-on-exit`** is **on**, **`pane-died`** and **`pane-exited`** hooks run in line with **tmux** semantics for that scenario.
2. Hooks must not fire repeatedly every idle tick; they should run when the "dead under remain-on-exit" condition is first detected (or match tmux behavior if documented differently).
3. Firing hooks for remain-on-exit deaths must not imply removing panes or resizing the layout solely because the pane became dead (resize when the tree structure changes, e.g. a pane is actually removed).
4. With **`remain-on-exit` off**, existing behavior for reaping exited panes and firing hooks on removal remains correct.
5. Add or extend tests (unit and/or integration) that cover hooks firing when a pane becomes dead with **`remain-on-exit`**; update compatibility docs if the project maintains them.

tarikguney:
## Summary

- Fixes `pane-died` and `pane-exited` hooks not firing when `remain-on-exit` is enabled
- Hooks now fire when a pane's child process exits, regardless of whether the pane is retained or removed from the tree
- Layout resize only happens when panes are actually removed (not when they transition to dead state)

## Root cause

When `remain-on-exit` is on, `prune_exited()` marks panes as `dead` but keeps them in the tree. The hook-firing logic only checked whether the tree leaf count decreased (`leaves_after < leaves_before`), which **never** happens in the remain-on-exit case since dead panes are still counted as leaves. So hooks never fired.

## What changed

**`src/tree.rs`**:
- `prune_exited()` now returns `(Option<Node>, newly_dead_count)` — the second value tracks panes that transitioned alive→dead in each call
- `reap_children()` returns `(all_empty, any_pruned, any_newly_dead)` — separating "pane removed from tree" from "pane became dead"

**`src/server/mod.rs`**:
- Hooks fire when `any_pruned || any_newly_dead` (either a pane was removed OR became dead)
- Layout resize only runs when `any_pruned` (actual tree structure change)
- Uses `fire_hooks()` helper instead of inline hook execution

## Why hooks don't fire repeatedly

The `has_any_exited()` fast-path check (line 688) returns `false` for already-dead panes (`if p.dead { return false; }`), so once a pane transitions to dead state and hooks fire, subsequent reap ticks skip it entirely.

Fixes #227

## Test plan

- [ ] With `remain-on-exit on`: start a pane with a short-lived process, set `pane-died` hook — verify hook fires when process exits
- [ ] With `remain-on-exit off`: same test — verify existing behavior (hooks fire, pane is removed) still works
- [ ] Verify hooks fire exactly once per pane death, not on every reap tick
- [ ] `respawn-pane` after death should reset the pane; if it dies again, hooks should fire again

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -4023,15 +4023,18 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
         // Check if all windows/panes have exited (throttled to every 250ms)
         if last_reap.elapsed() >= Duration::from_millis(100) {
             last_reap = Instant::now();
-            let (all_empty, any_pruned) = tree::reap_children(&mut app)?;
+            let (all_empty, any_pruned, any_newly_dead) = tree::reap_children(&mut app)?;
             if any_pruned {
-                // A pane exited naturally - resize remaining panes to fill the space
+                // A pane was removed from the tree - resize remaining panes to fill the space
                 resize_all_panes(&mut app);
+            }
+            if any_pruned || any_newly_dead {
+                // A pane exited — fire hooks whether it was removed (remain-on-exit off)
+                // or just marked dead (remain-on-exit on).  Fixes #227.
                 state_dirty = true;
                 meta_dirty = true;
-                // Fire pane-died / pane-exited hooks
-                if let Some(cmds) = app.hooks.get("pane-died") { let cmds = cmds.clone(); for cmd in &cmds { let _ = execute_command_string(&mut app, cmd); } }
-                if let Some(cmds) = app.hooks.get("pane-exited") { let cmds = cmds.clone(); for cmd in &cmds { let _ = execute_command_string(&mut app, cmd); } }
+                crate::commands::fire_hooks(&mut app, "pane-died");
+                crate::commands::fire_hooks(&mut app, "pane-exited");
             }
             if app.exit_empty && all_empty {
                 let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -363,33 +363,40 @@ pub fn get_split_mut<'a>(node: &'a mut Node, path: &Vec<usize>) -> Option<&'a mu
     Some(cur)
 }
 
-pub fn prune_exited(n: Node, remain_on_exit: bool) -> Option<Node> {
+/// Prune exited panes from the tree.  Returns `(Option<Node>, newly_dead_count)`:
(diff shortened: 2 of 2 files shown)


psmux:
Hey @tarikguney, closing this since the same fix already landed via the second commit in #225 (80bb22e), which is now merged into master. The changes in `tree.rs` and `server/mod.rs` are identical. Issue #227 is resolved. Thanks for the quick turnaround on both PRs!

Issue #197, opened 2026-04-10
ctrl+v freezes the terminal over SSH


ivilkhovyi:
Hello,

When over ssh, and doing `ctrl+v` of specific text, the terminal stops receiving any further input.

__the text in the buffer -> the result__
C:\Users\myusername\Documents\PowerShell\Microsoft.PowerShell_profile.ps1 -> **terminal stuck**
C:\Users\myusername\Documents\PowerShell\ -> OK
"C:\Users\myusername\Documents\PowerShell\Microsoft.PowerShell_profile.ps1" -> OK
ddddddddddd -> OK
C:\Users\myusername\unity_build.log -> OK

When directly in the terminal (not ssh) and pasting any text it shows:
```
PS C:\Users\myusername\Sources> cd ..
PS C:\Users\myusername> Tips: Always use  --plain  for AI-friendly output. Multiple flags of the
>>   same
>>   type work:  --check-ac 1 --check-ac 2 . Use PowerShell backtick-n for
>>   newlines:
>>   "Line1 ΓÇï nLine2" .ddddddddddddd
```

The `ddddddddddddd` is what was in my buffer and the rest is what was in my buffer a few hours ago.

my `.tmux.conf`
```
set -g prefix C-a
set -g prefix2 `
set -g history-limit 50000
set -g mouse on
set -g base-index 1
set -g renumber-windows on
set -g automatic-rename on
set -g automatic-rename-format '#{b:pane_current_command}'
set -g pane-border-status top
set -g pane-border-format ' #{pane_index}: #{pane_title} '
set -sg escape-time 10
set -g status-right ''
set-hook -g client-resized 'refresh-client'
```

psmux:
When you mention over ssh, which is the client ?  You're connecting to another ssh server from within psmux? or psmux is running in your ssh server? need more clarity. 

OS,Shell of where Psmux is running
OS, Shell of your ssh client

psmux:
Hey @ivilkhovyi, thanks for reporting this!

I dug deep into the code and found the root cause. The VT parser state machine that handles bracketed paste (the protocol terminals use to wrap pasted text in special escape sequences) had **no timeout and no size limit** on the paste buffer.

Here is what was happening:

1. When you paste text over SSH, your terminal wraps it in bracketed paste markers: `\x1b[200~` (start) ... `\x1b[201~` (end)
2. The VT parser enters a "Paste" accumulation state when it sees the start marker
[...]

PR #238, opened 2026-04-18
fix: reduce paste_suppress_until window from 2s to 200ms (closes #237)

### LINKED ISSUE #237: Regression: typing freezes for ~2 seconds intermittently (Windows) — bisected to 3bf380d
## Summary
During normal typing in a psmux session, input periodically stalls — the cursor stops advancing, typed characters do not appear, and after ~2 seconds typing resumes and accumulated keystrokes show up. The stall happens silently with no error and self-recovers. More visible under fast or sustained typing.

## Bisect
Narrowed to a single commit via binary search between a known-good 3-week-old state and current master:

- `86a7519` (2026-04-10 10:20) — works
- `cb0d429` (2026-04-11 16:39) — works
- `8875bb5` (2026-04-11 19:22) — works
- `c28a428` (2026-04-11 23:02) — works
- **`3bf380d` (2026-04-12 03:18) — reproduces the freeze**
- master `a138a9e` — reproduces

Single-commit delta: `3bf380d` "fix(paste): eliminate local paste duplication on Windows (#197)".

## Root cause
`3bf380d` added a `paste_suppress_until` timestamp in `src/client.rs` and wired five new call sites (POST_CONFIRMED, CLIPBOARD_FALLBACK, TOP_CONFIRMED, STAGE2_TIMEOUT, Event::Paste handler) that set it to `now + 2 seconds` whenever a paste-shaped event is detected.

Pre-existing code (lines 2079–2107, the `KeyCode::Char(c)` handler) already contained a suppression check that drops typed characters while `paste_suppress_until` is in the future. Before `3bf380d` that branch was effectively dormant because `paste_suppress_until` was rarely set. After `3bf380d` any of the five new call sites can fire on normal typing — in particular **STAGE2_TIMEOUT**, which triggers when typed characters accumulate in `paste_pend` without a matching Ctrl+V within 300ms (i.e. the exact pattern of a user typing normally).

Once set, the suppression window silently drops every typed character for **2 seconds**. That is exactly the observed symptom:

> Type at fluent pace → characters accumulate in paste_pend → stage2 heuristic misclassifies them as a paste → `paste_suppress_until = now + 2s` → next 2 seconds of typing silently dropped → window expires → typing resumes, accumulated keystrokes surface.

## Why it doesn't reproduce everywhere
- Requires typing fast enough for `paste_pend` to fill before a 300ms quiet period — slow single-finger typing may not trip it.
- Pastes themselves still work, because the paste content is sent before `paste_suppress_until` is set.

## Environment
- psmux master (HEAD `a138a9e`)
- Windows 11, pwsh 7.x, PSReadLine 2.4.5, Windows Terminal 1.23.x
- Reproduces locally and over RDP.

## Related
- Supersedes closed discussion in #233, #236.
- The 2-second window traces to PR #197 (paste duplication fix).

tarikguney:
## Summary
- Reduces `paste_suppress_until` window in `src/client.rs` from 2 seconds to 200 milliseconds across all 6 call sites.
- Eliminates the silent 2-second typing freeze reported in #237 (regression from #197 / `3bf380d`).
- No behavior change for paste: the 200ms window still covers ConPTY's async clipboard-inject race (resolves <50ms) with a 4× margin.

## Why
`paste_suppress_until` blocks `KeyCode::Char(c)` events while the window is in the future (pre-existing check at `src/client.rs:2079–2107`). Commit `3bf380d` (#197) added five new call sites that set this window to `now + 2s`, including `STAGE2_TIMEOUT` — which fires when `paste_pend` accumulates typed chars without a Ctrl+V within 300ms. That is the exact pattern of normal fast typing, so fast typing trips the stage2 heuristic, sets a 2-second suppression window, and silently drops every subsequent keystroke for 2 seconds. See #237 for the full bisect and root-cause analysis.

The 2-second window is far longer than the race it guards. Original motivation (#197): ConPTY may inject clipboard characters asynchronously after a Ctrl+V Release event arrives, so a subsequent `Event::Paste` / stage2 burst could cause a double-send. This async inject completes in well under 50ms on tested Windows 11 systems. 200ms is a comfortable margin while staying below human-perceptible typing latency.

## Test plan
Tested on Windows 11 / pwsh 7 / Windows Terminal 1.23 / PSReadLine 2.4.5:

- [x] Fast typing no longer stalls. Verified by sustained typing in a psmux session — previously reproduced the 2s freeze every few seconds; with this change, no stalls observed.
- [x] Single-line Ctrl+V paste — `hello world` arrives exactly once.
- [x] Multi-line Ctrl+V paste — 3-line snippet arrives intact, no duplication.
- [x] Rapid sequential Ctrl+V (×3 in quick succession) — all three instances delivered, none duplicated.
- [x] Existing psmux test suite passes (no changes to compile or test harness).

## Risk / rollout
- Worst-case regression if 200ms proves insufficient on slower hardware: paste duplication under a specific timing window. Easy to tune upward if reports surface.
- Alternative minimal fix (if reviewers prefer): remove the `KeyCode::Char` suppression check entirely and rely on `paste_pend` timing plus bracketed-paste `Event::Paste` for dedup. Happy to switch to that approach.

Closes #237

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -960,7 +960,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         let encoded = base64_encode(&paste_pend);
                         cmd_batch.push(format!("send-paste {}\n", encoded));
                         // Suppress clipboard-read fallback
-                        paste_suppress_until = Some(Instant::now() + Duration::from_secs(2));
+                        paste_suppress_until = Some(Instant::now() + Duration::from_millis(200));
                     }
                     paste_pend.clear();
                     paste_pend_start = None;
@@ -1057,7 +1057,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                         // Suppress the clipboard-read fallback that fires
                         // when Ctrl+V Release arrives later (the paste was
                         // already sent via stage2).
-                        paste_suppress_until = Some(Instant::now() + Duration::from_secs(2));
+                        paste_suppress_until = Some(Instant::now() + Duration::from_millis(200));
                     }
                 }
             }
@@ -2161,7 +2161,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                             paste_pend_start = None;
                             paste_stage2 = false;
                             paste_confirmed = false;
-                            paste_suppress_until = Some(Instant::now() + Duration::from_secs(2));
+                            paste_suppress_until = Some(Instant::now() + Duration::from_millis(200));
                         }
                     }
                     Event::Mouse(me) => {
@@ -2358,7 +2358,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                     selection_changed = true;
                                     // Suppress text key events that VS Code's ConPTY
                                     // injects after a right-click copy action.
-                                    paste_suppress_until = Some(Instant::now() + Duration::from_secs(2));
+                                    paste_suppress_until = Some(Instant::now() + Duration::from_millis(200));
                                 } else {
                                     // No selection, no TUI — paste from clipboard (pwsh-style)
                                     rsel_start = None;
@@ -2602,7 +2602,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                 paste_confirmed = false;
                 // Suppress subsequent char accumulation and clipboard-read
                 // fallback — the paste was already delivered.
-                paste_suppress_until = Some(Instant::now() + Duration::from_secs(2));
+                paste_suppress_until = Some(Instant::now() + Duration::from_millis(200));
             } else if paste_confirmed && paste_pend.is_empty() {
                 // Ctrl+V Release with no buffered chars.  If paste was
                 // already sent via stage2 timeout or Event::Paste, the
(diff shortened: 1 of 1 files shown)


psmux:
Thank you @tarikguney for this excellent fix and your thorough analysis in #237. Your bisect to `3bf380d` was spot on, and the root cause analysis was crystal clear.

We ran independent E2E validation to prove both the bug and the fix:

## Bug Confirmed

Two test suites proved the 2s suppression window was silently dropping legitimate keystrokes:

1. **Batch keystroke injection test:** Injected 3+ chars in under 20ms (triggering stage2), then typed more chars during the resulting 2s window. Result: 9 chars silently dropped. Pane showed `echo AAADDD` instead of `echo AAABBBCCCDDD`.

2. **Real Ctrl+V paste test:** Typed BEFORE, pasted PASTED, then typed AFTER. Result: All 5 chars of AFTER were suppressed. Debug log showed exact timestamps (0.6s after paste, well within the 2s window).

## Fix Validated

After applying your change, we verified:

- Single Ctrl+V: delivered exactly once (no duplication)
- Multi-line paste: all 3 lines delivered
- Rapid sequential Ctrl+V: no extra duplication  
- Typing immediately after paste: works (not frozen)
- Windows path paste (#197 trigger): clean output, no trailing tilde
- TCP send-paste (server code path): delivered once

The 200ms window still provides a 4x safety margin over ConPTY's async inject race (resolves in under 50ms) while eliminating the human-perceptible typing freeze.

Test files for reproduction:
- `tests/test_issue237_final_proof.ps1` (stage2 heuristic freeze)
- `tests/test_issue197_paste_validation.ps1` (comprehensive paste scenarios)

Thanks for the quality contribution. Merged with Co-authored-by credit.

PR #241, opened 2026-04-19
Updated username and path checks to use eq_ignore_ascii_case

pbolduc:
See: https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#manual_ignore_case_cmp


psmux:
Hey Phil! Thanks so much for this contribution, really appreciate you taking the time to catch this and submit a clean fix.

You're absolutely right, using `eq_ignore_ascii_case` is the idiomatic way to do case insensitive string comparison in Rust. The old `to_ascii_lowercase()` approach was allocating a brand new string just to compare it, which is wasteful. Your change is cleaner, more efficient, and aligns with the clippy `manual_ignore_case_cmp` lint.

Merged this in with your authorship preserved on the commit. Welcome to the psmux contributors! Looking forward to seeing more from you. 🙌

PR #248, opened 2026-04-21
feat(picker): digit jump buffer + Enter in session chooser (#247)

### LINKED ISSUE #247: Quick session switching by number (parity with window switching)
## Problem

Switching between sessions is significantly slower than switching between windows:

- **Window switching** is fast — `PREF <number>` jumps directly to a window.
- **Session switching** requires `PREF s` to open the session picker, then arrow keys up/down until the target session is found. For users with many sessions this is tedious.

Additionally, while the session picker is open, numeric keys pass through to the underlying terminal as input instead of being consumed by the picker. So even if a user intuitively tries to "just type the number," it doesn't work — and it has the side effect of injecting digits into whatever is running in the focused pane.

## Proposal

Make the session picker consume digit keys into a jump buffer, then jump to the numbered entry when Enter is pressed. While the picker opened by `PREF s` is visible:

- `1`..`9`, `0` — append to a jump buffer, displayed at the bottom of the picker as `go to 12` while mid-entry
- `Enter` — parse buffer as a 1-based index and jump to that entry; if the buffer is empty, fall back to the arrow-cursor selection (existing behavior, unchanged)
- `Backspace` — delete the last digit from the buffer
- `Esc` — clear the buffer and close the picker
- `Up`/`Down` — arrow cursor navigation (unchanged)
- `x` — kill the session at the arrow cursor (unchanged)

Additional UX changes folded in:

- Every row is numbered with a right-aligned dynamic-width column, so the digit→session mapping is visible without trial and error.
- Overlay height is sized to the entry count rather than a fixed 20 lines — previously left a lot of empty space for short lists.
- While the picker is open, no character key escapes to the focused pane's PTY (fixes the pre-existing input-leakage side effect).

Out-of-range or unparseable buffer on Enter is a silent no-op — buffer stays so the user can Backspace and fix. Killing a session clears the buffer to avoid stale indexes.

## Environment

- psmux version: 3.3.3
- OS: Windows 11
- Terminal: Windows Terminal

tarikguney:
Closes #247.

## Summary

Make the session picker (`PREF s`) accept a number-then-Enter jump, similar to how windows are selected by digit. Also closes the pre-existing side-effect where character keys typed into the picker leaked through to the focused pane's PTY.

## Behavior (while picker is open)

- `1`..`9`, `0` — append to a jump buffer, shown at the bottom as `go to 12`
- `Enter` — parse buffer as 1-based index and jump; empty buffer falls back to the arrow-cursor selection (existing behavior, unchanged)
- `Backspace` — delete the last digit from the buffer
- `Esc` — clear buffer and close picker
- `Up`/`Down`/`PageUp`/`PageDown`/`Home`/`End` — arrow cursor navigation (unchanged)
- `x` — kill session at the arrow cursor (unchanged)

Out-of-range or unparseable buffer on Enter is a silent no-op — buffer stays so the user can Backspace and fix. Killing a session clears the buffer since indexes shift.

## UI changes

- Every row is now numbered with a right-aligned dynamic-width column so the digit-to-session mapping is visible without trial and error.
- Overlay height grows by 2 rows when the buffer indicator is active.
- All non-handled character keys are absorbed while the picker is open (fixes the input-leakage side effect).

## Test plan

- [x] `tests/test_issue247_session_picker_digit.ps1` — 14/14 passing. Covers: state declaration, digit-buffer accumulator, no immediate-switch shortcut, Enter parses buffer as 1-based, Backspace pops, Esc clears, leak-guard catch-all, title hint, dynamic row numbering, buffer preview rendering, dynamic overlay height, and functional verification that the picker data source (port files + `session-info` over TCP) reaches multiple sessions.
- [x] Manual: `C-b s`, type a number, press Enter — jumps to that session. Backspace edits. Esc closes without side effects. Non-digit chars no longer leak to the PTY.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -460,6 +460,9 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
     let mut session_entries: Vec<(String, String)> = Vec::new();
     let mut session_selected: usize = 0;
     let mut session_scroll: usize = 0;
+    // Digits typed while the picker is open accumulate here and are consumed
+    // when the user presses Enter — "12" + Enter jumps to the 12th session.
+    let mut session_num_buffer = String::new();
     let mut confirm_cmd: Option<String> = None;  // pending kill confirmation
     let current_session = name.clone();
     let mut last_sent_size: (u16, u16) = (0, 0);
@@ -1674,6 +1677,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 session_entries.clear();
                                 session_selected = 0;
                                 session_scroll = 0;
+                                session_num_buffer.clear();
                                 let dir = format!("{}\\.psmux", home);
                                 if let Ok(entries) = std::fs::read_dir(&dir) {
                                     for e in entries.flatten() {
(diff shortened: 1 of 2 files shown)


psmux:
Merged in bf0a14a. Thank you for the thoughtful contribution Tarik.

What I really appreciated about this one:

* The state lives where it belongs (client side, alongside the rest of the picker state) and the buffer is cleared on every entry into and exit from the picker, so there are no stale indexes after a kill or a re open.
* Enter falls back to the arrow cursor when the buffer is empty, so the existing muscle memory still works. No surprises for current users.
* The leak guard catch all on `KeyCode::Char(_)` is a quiet but real fix. The pre existing input bleed into the focused pane was annoying and easy to miss in a review, glad you folded it in.
* Dynamic overlay height plus the right aligned numbered column makes the picker feel native, not bolted on.
* The 14 case test script under tests covers state, all four key paths (digit, Enter, Backspace, Esc), the leak guard, every renderer change, and a real TCP session-info round trip across three live sessions. Solid practical proof.

Verified locally before merging:

* `.\scripts\build.ps1` clean on Windows
* `tests\test_issue247_session_picker_digit.ps1` 14 of 14 passing
* `cargo test --release --bin psmux` 1744 of 1744 passing, no regressions
* Merged via rebase so your commit lands on master with you as the author

Will go out in the next release. Thanks again for the clean PR and the test coverage that came with it.

psmux:
Hey Tarik, your PR was the perfect foundation for something I wanted to extend further.

The digit-jump UX you built for the session picker was so clean that I rolled it out to every other interactive picker in the client, all landed in commit 0f774f4 on master just now.

What is covered now:

**choose-tree / choose-window** (C-b w)
The tree picker gets its own `tree_num_buffer` with the exact same flow: digits append, Enter jumps using a 1-based index (falls back to the arrow cursor when the buffer is empty), Backspace edits, Esc clears and closes. I also slipped in the leak-guard catch-all and PageUp / PageDown / Home / End which were missing before. Every row gets a right-aligned numbered prefix and the 2-row "go to N" indicator grows the overlay height the same way yours does.

**choose-buffer** (C-b =)
Same treatment with `buffer_num_buffer`. One small nuance: when the user presses `d` to delete an entry the buffer is cleared (indexes shift, just like you did with `x` in the session picker). Row labels now look like `1. buffer0: 17 bytes: "hello"` so the number is front and center.

**customize-mode** (C-b :  then `customize-mode`, or however it is bound)
This one is different because customize state lives server-side. There is no local list to index into directly. So when the user types a number and presses Enter, the client computes `target_pos - current_pos` from the mirrored `srv_customize_options` slice and dispatches `customize-navigate <delta>`. It still feels like an instant jump. Digits are ignored while the user is actively editing a value (the buffer is auto-cleared on edit start) so there is no interference with value input.

**Tests**
`tests/test_picker_digit_jump_all.ps1` covers all three pickers with 35 cases: state declarations, picker-open clears, digit/Enter/Backspace/Esc handlers, leak guards, renderer row numbering, "go to N" indicators, title hints, and functional verification that each picker actually has rows to jump to (5 windows, 3 buffers over TCP, 56 options via show-options). Layer-2 visible-client check too. 35 of 35 passing.

Your original `test_issue247_session_picker_digit.ps1` still runs 14 of 14, no regressions.

Really appreciated the clean architecture you set up in #248. Having a tight, isolated per-picker buffer with clear open/close/kill lifecycle made it straightforward to replicate everywhere. Good pattern.

tarikguney:
Loved this!

PR #249, opened 2026-04-21
fix: select-pane -Z loses its -Z flag when used in bind-key

LizardLiang:
`select-pane -Z` is documented as supported (`keep zoomed`), but when it is used in `bind-key`, the stored binding gets rewritten into plain directional `select-pane` and loses the `-Z` behavior.

For example:
```tmux
bind-key -r h select-pane -Z -L
bind-key -r j select-pane -Z -D
bind-key -r k select-pane -Z -U
bind-key -r l select-pane -Z -R
```
### Repro
1. Start psmux
2. Split into multiple panes
3. Zoom the active pane
4. Press prefix + h/j/k/l

### Expected
The binding should preserve the original command, e.g.
select-pane -Z -L
and pane switching should keep zoom.

### Actual
The binding is rewritten into plain directional navigation, e.g.
select-pane -L
so the zoom-preserving behavior is lost before runtime.
 
### Likely cause
bind-key appears to normalize commands into simplified internal actions too aggressively.
That works for exact commands like:
select-pane -L
but not for commands where extra flags change semantics, such as:
select-pane -Z -L

### Proposed fix
When a key binding is parsed, only convert it into a simplified internal action if the command is an exact match.
For example:
- select-pane -L can safely be treated as “move focus left”
- select-pane -Z -L should not be treated the same way, because -Z changes the behavior

So if a binding includes extra flags that matter, the original command should be kept as-is instead of being rewritten into the simpler directional action.

In short:
- exact simple commands can be normalized
- commands with additional semantic flags should be preserved verbatim

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -503,7 +503,9 @@ pub fn parse_command_to_action(cmd: &str) -> Option<Action> {
         "resize-pane" | "resizep" if parts.iter().any(|p| *p == "-Z") => Some(Action::ZoomPane),
         "zoom-pane" => Some(Action::ZoomPane),
         "select-pane" | "selectp" => {
-            if parts.iter().any(|p| *p == "-U") {
+            if parts.iter().any(|p| *p == "-Z") {
+                Some(Action::Command(cmd.to_string()))
+            } else if parts.iter().any(|p| *p == "-U") {
                 Some(Action::MoveFocus(FocusDir::Up))
             } else if parts.iter().any(|p| *p == "-D") {
                 Some(Action::MoveFocus(FocusDir::Down))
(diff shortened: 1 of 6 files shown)


psmux:
Hey @LizardLiang, thank you so much for digging into this and sending such a clean, well scoped patch.

I pulled the branch locally and walked through the change end to end:

* No Cargo.toml or dependency changes, no shell or network surface, nothing suspicious. Just the targeted logic fix and tests.
* Compared against tmux upstream cmd-select-pane.c. Tmux uses window_push_zoom(w, 0, args_has(args, 'Z')) and window_pop_zoom to keep the new pane zoomed when -Z is passed. Your move_focus_preserving_zoom path (toggle off, move, toggle on) achieves the exact same end state, so this is real tmux parity, not a workaround.
* Routed through both code paths (the in process command in commands.rs and the TCP server SelectPane(dir, keep_zoom) request in server/connection.rs and server/mod.rs), which is exactly the right move since psmux dispatches through both depending on whether it is the embedded prompt or an external psmux CLI call.
* Ran the full suite: 1747 passed, 0 failed. Your three new tests (bind_key_select_pane_z_stays_as_command, parse_config_line_select_pane_z_stays_as_command, serialized_bindings_preserve_select_pane_z_command) all pass and lock in the regression nicely.

Squashed and merged as f0b94c2 with you as the author so the credit stays with you. This will ship in the next release.

Really appreciate the contribution and the thoughtful issue write up that came with it. Hope to see more from you.

LizardLiang:
> Hey @LizardLiang, thank you so much for digging into this and sending such a clean, well scoped patch.
> 
> I pulled the branch locally and walked through the change end to end:
> 
> * No Cargo.toml or dependency changes, no shell or network surface, nothing suspicious. Just the targeted logic fix and tests.
> * Compared against tmux upstream cmd-select-pane.c. Tmux uses window_push_zoom(w, 0, args_has(args, 'Z')) and window_pop_zoom to keep the new pane zoomed when -Z is passed. Your move_focus_preserving_zoom path (toggle off, move, toggle on) achieves the exact same end state, so this is real tmux parity, not a workaround.
> * Routed through both code paths (the in process command in commands.rs and the TCP server SelectPane(dir, keep_zoom) request in server/connection.rs and server/mod.rs), which is exactly the right move since psmux dispatches through both depending on whether it is the embedded prompt or an external psmux CLI call.
> * Ran the full suite: 1747 passed, 0 failed. Your three new tests (bind_key_select_pane_z_stays_as_command, parse_config_line_select_pane_z_stays_as_command, serialized_bindings_preserve_select_pane_z_command) all pass and lock in the regression nicely.
> 
> Squashed and merged as [f0b94c2](https://github.com/psmux/psmux/commit/f0b94c20d3b9ea9369ff7ce2ae6c45494299a615) with you as the author so the credit stays with you. This will ship in the next release.
> 
> Really appreciate the contribution and the thoughtful issue write up that came with it. Hope to see more from you.

Thanks for the quick merge! I've been looking for a solid tmux alternative for a long time, and I'm really enjoying using this tool. Appreciate all your hard work on this project!

psmux:
Thank you for the contribution @LizardLiang . I'm really happy that you enjoy it. Share the word with your friends and colleagues!

PR #251, opened 2026-04-21
fix(picker): skip AUTH ack reliably in session-info fetch (#250)

### LINKED ISSUE #250: Session picker shows "OK" instead of session info for random entries (race in AUTH ack parsing)
## Summary

Opening the session picker with `PREF + s` intermittently renders some rows as just `OK` instead of the real `sessname: N windows (created ...)` info line. Which rows are affected changes every time the picker is reopened, so it's clearly a race.

## Screenshot

Image

## Repro

1. Have several `~/.psmux/*.port` sessions running.
2. Press `PREF + s` to open the session picker.
3. Close it and reopen it a few times.

Some entries show `N.   OK` with no session name, and the set of affected entries changes each open.

## Root cause

`src/client.rs` ~1692-1709 populates the picker by connecting to each session's server, sending `AUTH <key>\n` followed immediately by `session-info\n`, then reading **exactly two lines** and assuming:

- line 1 = the `OK\n` AUTH ack
- line 2 = the session-info response

```rust
let _ = ss.set_read_timeout(Some(Duration::from_millis(25)));
let _ = write!(ss, "AUTH {}\n", sess_key);
let _ = ss.write_all(b"session-info\n");
let mut br = BufReader::new(ss);
let mut al = String::new();
let _ = br.read_line(&mut al);                       // "assumed" OK ack
let mut line = String::new();
if br.read_line(&mut line).is_ok() && !line.trim().is_empty() {
    line.trim().to_string()                          // "assumed" info
} else {
    format!("{}: (no info)", base)
}
```

Server protocol (`src/server/connection.rs:59` and `:1283-1294`):
- `AUTH <key>` → writes `OK\n` synchronously in the connection thread.
- `session-info` → dispatches `CtrlReq::SessionInfo` onto the main event loop via mpsc and blocks on `rrx.recv()` until the main loop responds, then writes `<info>\n`.

The session-info reply therefore has **variable latency** — it has to round-trip through the server's main event loop, which can be busy driving another client's frames, processing input, etc.

With the per-read timeout set to only **25 ms**:

- **Happy path:** first `read_line` reads `OK\n` within 25 ms, second reads `<info>\n`. Correct row.
- **Race path:** when the `OK\n` ack arrives later than 25 ms after the client's send (thread-creation overhead, scheduling, ambient load), the first `read_line` times out with `al = ""` and the error is swallowed by `let _ = ...`. The second `read_line` runs with a fresh 25 ms budget, and by then `OK\n` has landed in the socket — so it's consumed as `line = "OK\n"`, `info = "OK"`, and the row renders as `N.   OK` with no session name. The actual `<info>\n` is never read.

The picker re-enumerates from scratch on every open, so which sessions lose the race varies run to run.

## Notes

- This isn't introduced by #247 (digit-jump picker) — the enumeration block is byte-identical to v3.3.3 (`git diff v3.3.3..HEAD -- src/client.rs` shows only UI / key-handling changes). The race was latent before and is now more visible.
- The same AUTH+command pattern is used in a few places with different timeouts:
  - `list-tree` in `client.rs` ~1605-1626 (50 ms connect / 100 ms read) — same "blindly take second line" logic, same latent race, just less hit.
  - `choose-buffer` in `client.rs` ~2046-2057 (100 ms connect / 200 ms read) — same logic.
  - `list-sessions` in `main.rs` ~390-395 — does the right thing: checks `line.trim() == "OK"` before consuming a second line.
- No server restart is required to reproduce — the client is the one misinterpreting the framing.

tarikguney:
Closes #250.

## Summary

- The session picker rendered random rows as \`OK\` instead of \`sessname: N windows (created ...)\`, with the affected rows changing on every reopen. Root cause was a framing race in the AUTH+session-info exchange: a late \`OK\n\` ack landed in what the code assumed would be the info line.
- Extract the fetch into \`session::fetch_session_info\`, which detects the \`OK\` ack explicitly (only consumes a second line if the first was actually \`OK\`) and also filters \`ERROR:\` responses so auth failures don't get displayed as session names.
- Bump the per-read timeout to 150 ms so the degraded \`(not responding)\` label is reached much less often on busy systems.

## Test plan

New \`tests-rs/test_session.rs\` mounts onto \`src/session.rs\` via \`#[cfg(test)] mod tests\` (same pattern used by \`test_client.rs\`). Tests spin up an in-process \`TcpListener\` that impersonates the server's protocol and call the real \`fetch_session_info\` — no re-implementation of the parser in the tests.

- [x] \`happy_path_returns_info_line\` — standard \`OK\n<info>\n\` sequence returns the info.
- [x] \`issue_250_late_auth_ack_is_not_reported_as_session_info\` — server delays the ack past the client's per-read timeout; asserts the result is never \`Some(\"OK\")\`. Verified to fail on the old implementation with \`assertion left: Some(\"OK\")\` \`right: Some(\"OK\")\`.
- [x] \`only_ok_ack_received_returns_none\` — server sends ack and nothing else.
- [x] \`connect_refused_returns_none\` — port is closed.
- [x] \`auth_rejected_returns_none\` — server writes \`ERROR: Invalid session key\n\`.

\`\`\`
running 5 tests
test session::tests::auth_rejected_returns_none ... ok
test session::tests::happy_path_returns_info_line ... ok
test session::tests::connect_refused_returns_none ... ok
test session::tests::issue_250_late_auth_ack_is_not_reported_as_session_info ... ok
test session::tests::only_ok_ack_received_returns_none ... ok
test result: ok. 5 passed; 0 failed
\`\`\`

No server restart required to benefit — this is a client-side fix and is wire-compatible with any still-running pre-update server.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1689,24 +1689,13 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                                         if let Ok(p) = port_str.trim().parse::<u16>() {
                                                             let sess_addr = format!("127.0.0.1:{}", p);
                                                             let sess_key = read_session_key(base).unwrap_or_default();
-                                                            let info = if let Ok(mut ss) = std::net::TcpStream::connect_timeout(
-                                                                &sess_addr.parse().unwrap(), Duration::from_millis(25)
-                                                            ) {
-                                                                let _ = ss.set_read_timeout(Some(Duration::from_millis(25)));
-                                                                let _ = write!(ss, "AUTH {}\n", sess_key);
-                                                                let _ = ss.write_all(b"session-info\n");
-                                                                let mut br = BufReader::new(ss);
-                                                                let mut al = String::new();
-                                                                let _ = br.read_line(&mut al);
-                                                                let mut line = String::new();
-                                                                if br.read_line(&mut line).is_ok() && !line.trim().is_empty() {
-                                                                    line.trim().to_string()
-                                                                } else {
-                                                                    format!("{}: (no info)", base)
-                                                                }
-                                                            } else {
-                                                                format!("{}: (not responding)", base)
-                                                            };
+                                                            let info = crate::session::fetch_session_info(
+                                                                &sess_addr,
+                                                                &sess_key,
+                                                                Duration::from_millis(25),
+                                                                Duration::from_millis(150),
+                                                            )
+                                                            .unwrap_or_else(|| format!("{}: (not responding)", base));
                                                             session_entries.push((base.to_string(), info));
                                                         }
                                                     }
--- a/src/session.rs
+++ b/src/session.rs
@@ -122,6 +122,50 @@ pub fn send_auth_cmd_response(addr: &str, key: &str, cmd: &[u8]) -> io::Result<S
     Ok(buf)
 }
 
+/// Fetch a one-line `session-info` response from a session server.
+///
+/// Connects to `addr`, authenticates with `key`, sends `session-info`, and
+/// returns the server's info line. Returns `None` if the connection fails,
+/// authentication is rejected, or the info line doesn't arrive in time.
+///
+/// The AUTH ack (`OK\n`) is filtered out explicitly rather than assumed to
+/// be the first line: when the AUTH reply is slow enough for the first
+/// `read_line` to time out, a naive two-read protocol picks up the late
+/// `OK` in the second read and mis-reports it as the session info. See
+/// issue #250.
+pub fn fetch_session_info(
+    addr: &str,
+    key: &str,
+    connect_timeout: Duration,
+    read_timeout: Duration,
+) -> Option<String> {
+    let sock_addr: std::net::SocketAddr = addr.parse().ok()?;
+    let mut s = std::net::TcpStream::connect_timeout(&sock_addr, connect_timeout).ok()?;
+    s.set_read_timeout(Some(read_timeout)).ok()?;
+    let _ = s.set_nodelay(true);
+    write!(s, "AUTH {}\n", key).ok()?;
+    s.write_all(b"session-info\n").ok()?;
+    let _ = s.flush();
+
+    let mut br = std::io::BufReader::new(s);
+    let mut line = String::new();
+    if std::io::BufRead::read_line(&mut br, &mut line).ok()? == 0 {
+        return None;
+    }
+    if line.trim() == "OK" {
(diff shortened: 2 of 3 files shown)


psmux:
We will merge this but then address the root cause of fixing the protocol.

psmux:
Thanks @tarikguney, merged. Really clean catch and the in-process TCP fakes in the test file are a great pattern. The five test scenarios you wrote pinned the framing race precisely.

I built a follow-up commit on top (79ca29c) that takes this from a per-site patch to a root-cause fix. The short version:

1. The same race that produced "OK" in the session picker was latent in the choose-tree, choose-buffer, and command-prompt buffer pickers too. They each had their own ad-hoc `read_line; read_line` loop. I extracted your robust ack-skipping logic into `session::fetch_authed_response` and `session::fetch_authed_response_multi` and migrated all three picker call sites in `client.rs` to use them. Your `fetch_session_info` is now a thin wrapper, so your five regression tests still cover the original code path verbatim.

2. Added `session::validate_auth_key` that rejects keys containing CR, LF, or NUL before they go on the wire. The current key source (file under the user's profile) is trusted, but this is cheap defense-in-depth in case a future caller passes attacker-controlled input. All `AUTH ` writers funnel through it now.

3. Added a hard 256 KB cap on response payloads via `Read::take`, so a buggy or hostile peer that sends an unbounded line cannot force unbounded client memory or wall time. Bounds both memory and the worst-case time spent in `read_line`.

4. Performance: opening the session picker with N sessions used to cost up to `N * read_timeout`. Added `session::fetch_session_infos_parallel` which fans the per-session fetches across scoped threads, so the wall time is bounded by one `read_timeout` regardless of N. Single-input case skips the thread spawn entirely.

5. Tests:
   - 17 new Rust unit tests in `tests-rs/test_issue250_root_cause.rs` covering CRLF/NUL key rejection, all five ack-timing races on the new helper, the response size cap, the parallel fan-out wall-time bound (8 servers each delaying 120 ms must complete in well under N * 120 ms), mixed success/failure handling, and the empty/single input fast paths.
   - New `tests/test_issue250_root_cause.ps1` stands up six real psmux sessions, runs 480 live AUTH+session-info fetches over real TCP through the installed binary including a 5 ms tight-timeout stress run, and asserts zero "OK" or "ERROR:" leaks. All passed cleanly.
   - Full suite: `cargo test` 1796 passed, 0 failed.

Your tests still pass unchanged on top of the refactor, which was the point. Appreciate the contribution!

PR #254, opened 2026-04-23
fix: correct active pane border indicator for all split layouts

LizardLiang:
The `pane-active-border-style` indicator is not visible when panes are split, and behaves incorrectly in nested layouts.

### Repro

1. Start psmux
2. Split into left and right panes
3. Switch focus between panes

### Expected

The border segment adjacent to the focused pane should be highlighted with `pane-active-border-style`. For exactly 2 panes, the shared separator is split at the midpoint so each half reflects its adjacent pane's active state (matching tmux `pane-border-indicators` behaviour).

### Actual

No visible change when switching focus — the separator appears the same color regardless of which pane is active. In nested layouts (e.g. left pane + top-right / bottom-right), the separators are either all highlighted or split at an arbitrary midpoint unrelated to the focused pane's position.

### Root cause

Two separate issues:

**1. Post-render re-coloring pass overwrites everything**

After `render_json` draws the separators, a post-render pass re-colors all border characters based on adjacency to the active pane rect. For a simple 2-pane split, the separator is *always* geometrically adjacent to whichever pane is active — so the entire separator is always painted with `active_border_style`, producing no visible change when switching focus.

**2. Midpoint logic applied unconditionally in nested layouts**

When both adjacent children are leaf panes, the code used a midpoint split: the top half of `│` represents the left pane, the bottom half represents the right pane (and vice-versa for `─`). This works correctly for a simple 2-pane layout, but in a nested layout (e.g. `Split(H: [Left, Split(V: [TopRight, BotRight])])`) the horizontal `─` separator between TopRight and BotRight was being split left/right at an arbitrary midpoint, which does not correspond to the spatial position of either pane.

### Fix

**Re-coloring pass:** Restrict the post-render pass to junction characters only (`┼ ├ ┤ ┬ ┴`). This lets `render_json` own the style of straight `│` and `─` cells without being overwritten.

**Separator rendering:** Unify the logic into two cases:
- **Exactly 2 panes (whole window):** Use the midpoint half-highlight for the shared separator, matching tmux's `pane-border-indicators colour` behaviour.
- **3+ panes:** Use per-cell adjacency — each cell of the separator is highlighted only if the active pane directly borders that specific cell. This ensures only the segments forming the actual border of the focused pane are highlighted, regardless of layout complexity.

`LayoutJson::count_leaves()` is added to compute the total pane count. When zoomed, `total_panes` is set to 1 to reflect that only one pane is visible and to skip the traversal.

---

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3208,7 +3208,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                 }
             }
 
-            fn render_json(f: &mut Frame, node: &LayoutJson, area: Rect, dim_preds: bool, border_fg: Color, active_border_fg: Color, clock_mode: bool, clock_colour: Color, active_rect: Option<Rect>, mode_style_str: &str, zoomed: bool, border_status: &str, border_format: &str) {
+            fn render_json(f: &mut Frame, node: &LayoutJson, area: Rect, dim_preds: bool, border_fg: Color, active_border_fg: Color, clock_mode: bool, clock_colour: Color, active_rect: Option<Rect>, mode_style_str: &str, zoomed: bool, border_status: &str, border_format: &str, total_panes: usize) {
                 match node {
                     LayoutJson::Leaf {
                         id,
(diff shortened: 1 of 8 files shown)


LizardLiang:
Closing to fix commits before re-opening.

PR #255, opened 2026-04-23
fix: correct active pane border indicator for all split layouts

LizardLiang:
The `pane-active-border-style` indicator is not visible when panes are split, and behaves incorrectly in nested layouts.

### Repro

1. Start psmux
2. Split into left and right panes
3. Switch focus between panes

### Expected

The border segment adjacent to the focused pane should be highlighted with `pane-active-border-style`. For exactly 2 panes, the shared separator is split at the midpoint so each half reflects its adjacent pane's active state (matching tmux `pane-border-indicators colour` behaviour).

### Actual

No visible change when switching focus — the separator appears the same color regardless of which pane is active. In nested layouts (e.g. left pane + top-right / bottom-right), the separators are either all highlighted or split at an arbitrary midpoint unrelated to the focused pane's position.

### Root cause

Two separate issues:

**1. Post-render re-coloring pass overwrites everything**

After `render_json` draws the separators, a post-render pass re-colors all border characters based on adjacency to the active pane rect. For a simple 2-pane split, the separator is *always* geometrically adjacent to whichever pane is active — so the entire separator is always painted with `active_border_style`, producing no visible change when switching focus.

### Fix

**Re-coloring pass:** Restrict the post-render pass to junction characters only (`┼ ├ ┤ ┬ ┴`). This lets `render_json` own the style of straight `│` and `─` cells without being overwritten. Keeping `│` and `─` in the pass causes it to overwrite the midpoint styles computed by `render_json` for the 2-pane case, so removal is necessary.

**Separator rendering:** Unify the logic into two cases:
- **Exactly 2 panes (whole window):** Use the midpoint half-highlight for the shared separator, matching tmux's `pane-border-indicators colour` behaviour.
- **3+ panes:** Use per-cell adjacency — each cell of the separator is highlighted only if the active pane directly borders that specific cell. This ensures only the segments forming the actual border of the focused pane are highlighted, regardless of layout complexity.

`LayoutJson::count_leaves()` is added to compute the total pane count. When zoomed, `total_panes` is set to 1 to reflect that only one pane is visible and to skip the traversal.

---

Current State
image

After change
image

3+ pane still working as usaul
image

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3243,7 +3243,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                 }
             }
 
-            fn render_json(f: &mut Frame, node: &LayoutJson, area: Rect, dim_preds: bool, border_fg: Color, active_border_fg: Color, clock_mode: bool, clock_colour: Color, active_rect: Option<Rect>, mode_style_str: &str, zoomed: bool, border_status: &str, border_format: &str) {
+            fn render_json(f: &mut Frame, node: &LayoutJson, area: Rect, dim_preds: bool, border_fg: Color, active_border_fg: Color, clock_mode: bool, clock_colour: Color, active_rect: Option<Rect>, mode_style_str: &str, zoomed: bool, border_status: &str, border_format: &str, total_panes: usize) {
                 match node {
                     LayoutJson::Leaf {
                         id,
@@ -3512,7 +3512,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
 
                         // Render children first
                         for (i, child) in children.iter().enumerate() {
-                            if i < rects.len() { render_json(f, child, rects[i], dim_preds, border_fg, active_border_fg, clock_mode, clock_colour, active_rect, mode_style_str, zoomed, border_status, border_format); }
+                            if i < rects.len() { render_json(f, child, rects[i], dim_preds, border_fg, active_border_fg, clock_mode, clock_colour, active_rect, mode_style_str, zoomed, border_status, border_format, total_panes); }
                         }
 
                         // Draw separator lines between children using direct buffer access.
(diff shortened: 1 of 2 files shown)


psmux:
Thank you so much for this fix, @LizardLiang! Your analysis is spot on. The legacy `both_leaves` half-highlight path was indeed coloring inner separators as if their adjacent leaf were active, even when the active pane lived elsewhere in the layout tree. The `count_leaves` + `total_panes == 2` gate is exactly the right approach.

Master had refactored the inner `render_json` function out to module level (`render_layout_json` in `src/client.rs`) which made GitHub mark this PR as conflicting. Rather than ask you to rebase, I ported your changes onto current master and merged them in commit 70e8602 with you credited as co-author.

What landed:
* `LayoutJson::count_leaves()` exactly as you wrote it.
* `total_panes` threaded through `render_layout_json` and the legacy half-highlight branches gated with `total_panes == 2`.
* `total_panes = 1` when zoomed, to skip traversal.
* The post-render re-coloring pass restricted to junction characters (`┼ ├ ┤ ┬ ┴`) only, so the new per-cell coloring of `│` and `─` is no longer clobbered for 3+ pane layouts.
* The two other call sites (`preview.rs` choose-tree previews, `main.rs` render-test) updated to pass `total_panes`.

I also added regression tests:
* `tests-rs/test_pr255_active_border.rs` (7 Rust unit tests) including a `TestBackend` render assertion that for a 3-pane `H[active, V[inactive, inactive]]` layout, no `─` cell on the right side is colored as active.
* `tests/test_pr255_active_border.ps1` (12 PowerShell E2E + TUI verification tests) covering 1/2/3/4-pane creation, `select-pane` cycling, geometry, zoom toggle, and `capture-pane` on all panes.

Full regression: 1828 / 1828 Rust tests pass, all PowerShell E2E suites pass.

Closing this PR since the fix is now in master. Huge thanks again, this was a clean, well reasoned change!

psmux:
Merged via 70e8602 (with you credited as co-author). Thanks again @LizardLiang!

Issue #246, opened 2026-04-21
Output rendering artifact: rows render with most cells cleared, leaving sparse characters at wide column positions (local, not SSH)


tarikguney:
## Summary

Terminal output inside a psmux pane is rendered with most cells on some rows cleared, leaving only a sparse set of characters scattered at wide column positions. Other rows on the same screen render fully and correctly. This appears visually similar to #73, but reproduces **locally** — no SSH involved — so the mangling is not purely an ANSI/SSH-transport issue.

Observed inside Claude Code (React Ink TUI). Not reproducible on demand; exact trigger unknown.

## Screenshot

<img width="1386" height="667" alt="Image" src="https://github.com/user-attachments/assets/80c7d702-a61a-43b0-b452-71a955970336" />

## Environment

- Windows 11
- psmux version: 
- No SSH; direct local session
- Host app: Claude Code (Ink-based TUI)

## Analysis (hypothesis, not yet confirmed by instrumentation)

The symptom shape (sparse characters at wide column positions within otherwise-blank rows) is consistent with psmux snapshotting the vt100 parser **between** successive `reader.read()` calls that together form a single logical Ink frame.

Mechanism:

1. **Ink emits cursor-positioned partial updates.** A logical frame can be a sequence of `CUP(r,c1) → text → CUP(r,c2) → text → …`, often preceded by a row-clear. These sequences can easily exceed the reader's 64KB buffer and get split across multiple `read()` calls.
2. **psmux's reader holds the parser mutex for one `read()` at a time** (`src/pane.rs:1137–1154`). Between lock acquisitions, the server's snapshot code (`src/layout.rs:620–741`) can grab the parser mutex and serialize whatever partial state exists. That partial state is emitted as a full `dump_buf` JSON frame.
3. **The partial frame is sent to the client and rendered** via ratatui, latching the sparse-cell state on screen until a subsequent correct frame is produced and rendered.

Why psmux and not the bare terminal: a raw terminal has no snapshot boundary — Ink's partial writes hit the real cell grid and are overwritten by the next read nanoseconds later. psmux's snapshot+push architecture introduces a discrete boundary that can freeze a mid-sequence state and display it as a finished frame. Claude Code emits valid VT sequences either way; psmux is what can latch them.

## Suggested investigation / fix direction

[...]

PR #267, opened 2026-04-28
fix: deliver latest frame under client backpressure

tarikguney:
## Summary

Fixes a second-layer candidate for the sparse-cell rendering artifact tracked in #246.

The original #246 parser-side race fix is present in the locally reproduced build, and the existing sparse-render harness does not reproduce against it. The remaining suspicious path is persistent-client frame backpressure: `push_frame()` documented latest-frame-wins behavior when the bounded channel is full, but actually skipped the newest frame.

Because pushed frames are complete snapshots, not deltas, a saturated client should drop stale queued frames and deliver the newest corrective snapshot.

## Changes

- Store registered frame channels instead of only their senders so `push_frame()` can access the receiver backlog.
- Scope the persistent writer's receiver lock to each drain pass instead of holding it forever.
- On full frame queues, drain stale queued frames and retry the newest frame.
- Add a focused unit test for saturated frame queue behavior.

## Validation

- `cargo test --quiet push_frame_replaces_stale_backlog_when_full`
- `cargo check --quiet`

Refs #246

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -110,7 +110,6 @@ if line.trim() == "PERSISTENT" {
     let directive_rx = crate::types::register_directive_channel(client_id);
 
     std::thread::spawn(move || {
-        let frame_rx = frame_chan.rx.lock().unwrap();
         loop {
             // 0. Check for queued directives (non-blocking) — these take priority
             while let Ok(directive) = directive_rx.try_recv() {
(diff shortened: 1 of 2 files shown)


psmux:
Merged into master at 247a7c9 with Co-authored-by credit. Thanks @tarikguney for the fix!

**Verification**: The backpressure bug was independently confirmed — \push_frame\ was silently dropping the newest frame when the channel was full (16 stale frames queued, newest discarded). Your fix correctly drains stale snapshots and retries the latest frame. All 1908 tests pass with zero regressions.

psmux:
@tarikguney Great contribution! Your fix was verified independently and merged into master (247a7c9) with Co-authored-by credit to you in the commit.

The bug was real: when the bounded frame channel (capacity 16) was full, the newest frame was silently dropped instead of replacing the stale backlog. Since frames are complete snapshots, the latest one is the only one that matters. Your approach of draining stale frames and retrying the newest is the correct solution.

All 1908 existing tests pass with zero regressions, plus 2 new unit tests were added to guard against this regression.

PR #270, opened 2026-04-29
fix: bypass split_with_gaps when zoomed to prevent 1-px pane bleed

LizardLiang:
### Repro

1. Start psmux
2. Split into two panes
3. Zoom the active pane (`prefix + z`)

### Expected

The zoomed pane fills the entire terminal. No content from the hidden pane is visible.

### Actual

A 1-pixel strip of the hidden pane bleeds into the edge of the zoomed pane. The zoomed pane is 1 column (or row) short of the full terminal size.

### Root cause

When zoomed, `toggle_zoom` sets the hidden pane's size to `0` and the active pane's to `100`. `render_layout_json` was passing these sizes to `split_with_gaps`, which has a "guarantee minimum 1 cell" loop — it detects any slot with size `0` and steals 1 column/row from the largest sibling to give it.

The old `if zoomed { return; }` guard appeared **after** the child-rendering loop, so both children were already rendered before it fired.

### Fix

Move the zoom check before `split_with_gaps` is called. When zoomed, find the child with non-zero size and recurse with the full `area` rect passed directly — no splitting, no gap arithmetic, no minimum-cell theft. The hidden pane is never rendered.

## Test plan

- [x] `cargo test test_zoom_bleed` — 4/4 passing
  - `zoomed_left_active_hidden_pane_label_never_rendered`
  - `zoomed_right_active_hidden_pane_label_never_rendered`
  - `zoomed_top_active_hidden_pane_label_never_rendered`
  - `zoomed_bottom_active_hidden_pane_label_never_rendered`
- [x] `cargo check --quiet`

## Screenshots

image

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -723,6 +723,16 @@ pub fn render_layout_json(
                 vec![(100 / children.len().max(1)) as u16; children.len()]
             };
             let is_horizontal = kind == "Horizontal";
+
+            if zoomed {
+                if let Some(i) = effective_sizes.iter().position(|&s| s != 0) {
+                    if let Some(child) = children.get(i) {
+                        render_layout_json(f, child, area, dim_preds, border_fg, active_border_fg, clock_mode, clock_colour, active_rect, mode_style_str, zoomed, border_status, border_format, total_panes);
+                    }
+                }
+                return;
+            }
+
             let rects = split_with_gaps(is_horizontal, &effective_sizes, area);
 
             for (i, child) in children.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


psmux:
@LizardLiang Great catch and clean fix! The root cause analysis was spot-on — split_with_gaps stealing 1px from the zoomed pane for the hidden one is exactly the kind of subtle rendering bug that's easy to miss. Moving the zoom check before the split arithmetic is the right approach, and the TestBackend tests are thorough (all 4 directions covered).

Merged — thank you for the contribution! 🎉

PR #283, opened 2026-05-08
fix: respond to ESC[6n cursor position queries reactively

### LINKED ISSUE #282: pwsh pane freezes when creating new window or split after lock/unlock
pwsh panes occasionally freeze when creating a new window or split after a session event such as locking and unlocking Windows (Win+L) or after a prior hang. When it occurs no prompt appears, no cursor is visible, and the pane produces no output. Resizing the terminal can temporarily unblock it.

### Repro

The following steps are the best chance to trigger the issue but are not guaranteed to reproduce it every time:

1. Start psmux with a pwsh pane
2. Lock the Windows session (Win+L), wait a few seconds, unlock
3. Create a new window or split

### Expected

New pwsh pane loads normally and shows a prompt.

### Actual

New pwsh pane freezes — blank, no cursor, no output.

### Probably cause

Examined via PSMUX_SERVER_DEBUG=1 — the pane goes silent immediately after pwsh emits an unanswered ESC[6n (Cursor Position Request):

[silence] pane=N silent 5s, last 88 bytes:
  <ESC>[6n<ESC>[?9001h<ESC>[?1004h<ESC>[m<ESC>]0;...\pwsh.exe<BEL><ESC>[?25h

Windows ConPTY children emit ESC[6n at startup and again after certain session events. The host must respond with ESC[row;colR or the child blocks indefinitely. psmux only writes a single preemptive ESC[1;1R at spawn time — by the time pwsh re-issues the query after lock/unlock, the pipe is empty and nothing answers.

A reactive responder (scan for ESC[6n in the parser thread, inject ESC[row;colR via the server loop) addresses the confirmed path and has been tested locally with positive results.

Environment:
- OS: Windows 11
- Shell: pwsh (PowerShell 7, installed via Scoop)

LizardLiang:
Closes #282 

## Problem

PowerShell (`pwsh`) freezes after Windows lock/unlock when running inside psmux. The root cause: pwsh emits `ESC[6n` (Cursor Position Request) at startup **and** again after session events like lock/unlock. The host must respond with `ESC[row;colR` or pwsh blocks indefinitely waiting for the response.

At spawn time, psmux already sends a preemptive `ESC[1;1R` to unblock the initial startup query. But this preemptive write is a one-shot — it cannot cover re-issued CPR queries after lock/unlock.

## Solution

- Add `scan_cpr_query()` in the reader thread to detect `ESC[6n` in PTY output
- When detected, set a per-pane `cpr_pending` flag and a global `CPR_DATA_PENDING` atomic
- In the server event loop (gated on `data_ready`), drain pending CPR flags and write the real `ESC[row;colR` response using the current cursor position from the VT parser
- Retain the preemptive `ESC[1;1R` write at spawn time to cover the initial startup race window

## Changes

- `src/types.rs` — add `cpr_pending: Arc<AtomicBool>` to `Pane` and `WarmPane`; add `CPR_DATA_PENDING` global atomic
- `src/pane.rs` — add `scan_cpr_query()`, thread `cpr_pending` through `spawn_reader_thread` (all 6 construction sites), add CPR signal in reader thread
- `src/server/helpers.rs` — add `drain_cpr_pending()` tree walker
- `src/server/mod.rs` — call `drain_cpr_pending` when `CPR_DATA_PENDING` is set
- `src/popup.rs` / `src/proxy_pane.rs` — populate `cpr_pending` field (popup's inline reader doesn't detect CPR; proxy forwards over TCP)
- `tests-rs/test_cpr_responder.rs` — 10 unit tests covering detection, false-positives, response format, and the 0x1b pre-check optimization

## Test plan

- [x] `cargo test cpr` — all 10 tests pass
- [x] `cargo build` — clean
- [x] Lock Windows and unlock while pwsh is running in a psmux pane — verify pwsh remains responsive and prompt re-renders correctlys.

### DIFF
--- a/src/cross_session_server.rs
+++ b/src/cross_session_server.rs
@@ -274,6 +274,7 @@ pub fn handle_pane_forward_inject(
         proxy_pane.data_version.clone(),
         proxy_pane.cursor_shape.clone(),
         proxy_pane.bell_pending.clone(),
+        proxy_pane.cpr_pending.clone(),
         proxy_pane.output_ring.clone(),
     );
     // Graft into the target window tree
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -97,7 +97,7 @@ pub fn create_window(pty_system: &dyn portable_pty::PtySystem, app: &mut AppStat
         }
         let epoch = std::time::Instant::now() - Duration::from_secs(2);
         let configured_shell = if app.default_shell.is_empty() { None } else { Some(app.default_shell.as_str()) };
-        let pane = Pane { master: wp.master, writer: wp.writer, child: wp.child, term: wp.term, last_rows: rows, last_cols: cols, id: wp.pane_id, title: hostname_cached(), title_locked: false, child_pid: wp.child_pid, data_version: wp.data_version, last_title_check: epoch, last_infer_title: epoch, dead: false, vt_bridge_cache: None, vti_mode_cache: None, mouse_input_cache: None, cursor_shape: wp.cursor_shape, bell_pending: wp.bell_pending, copy_state: None, pane_style: None, squelch_until: None, output_ring: wp.output_ring };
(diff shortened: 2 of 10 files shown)


psmux:
Hey @LizardLiang, this is excellent work. Merged in 8066250.

I reviewed the entire diff and I want to call out how clean this contribution is. You clearly studied the existing `bell_pending` architecture and replicated it precisely for the CPR case. The two tier atomic signaling (`cpr_pending` per pane + `CPR_DATA_PENDING` global to skip the tree walk), the `scan_cpr_query` pre check optimization with `data.contains(&0x1b)`, and the decision to keep the preemptive spawn time write alongside the new reactive path all show a deep understanding of how the server loop and reader threads interact.

The 10 regression tests are thorough, covering detection, false positives, partial sequences, and the 0 based to 1 based coordinate conversion. The cosmetic reformatting in `helpers.rs` is also welcome.

Really appreciate you taking the time to diagnose the root cause (pwsh re issuing DSR after lock/unlock) rather than just patching the symptom. This has been a pain point for users and your fix addresses it properly.

Thank you for this contribution!

PR #306, opened 2026-05-20
Add initial tmux control mode support

Blink-ai-dev:
Adds an initial `-C` / `-CC` control-mode path for psmux, focused on automation clients that create and drive sessions programmatically.

Included:
- parses `-C` and `-CC` as global flags
- supports control framing with `%begin`, `%end`, `%error`, `%session-changed`, `%exit`
- supports synthetic pane/window ids for `new-session -d -P -F '#{pane_id} #{window_id}'`
- rewrites synthetic `-t %...` / `-t @...` targets back to real psmux pane/window ids
- emits `%window-add`, `%window-close`, and best-effort `%output`
- initializes pane capture baselines immediately to avoid dropping output from fast `send-keys` after `new-session`
- preserves printable UTF-8 output and escapes control bytes

Validation:
- `cargo test control_mode -- --nocapture`
- `cargo test`
- `cargo build --release`
- manual installed-binary smoke with UTF-8 output
- manual installed-binary smoke for immediate `send-keys` after `new-session`
- downstream Paperclip live smoke: external `psmux -L paperclip kill-session -t ...` transitions the API row to `status='killed'`; approval detector creates a pending approval row

Known limitations:
- `%output` is based on `capture-pane` polling and snapshot deltas, not server-side PTY byte streaming
- attach-existing pane enumeration is not implemented yet
- `split-window` / `new-window` full tracking is not implemented yet
- output produced by a very fast startup command inside `new-session -d '<command>'` can still be absorbed into the initial baseline

### DIFF
--- a/src/control_mode.rs
+++ b/src/control_mode.rs
@@ -1,0 +1,592 @@
+use std::collections::HashMap;
+use std::io::{self, BufRead, Write};
+use std::process::{Command, Stdio};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+#[derive(Clone, Debug)]
+struct TrackedPane {
+    pane_id: String,
+    window_id: String,
+    actual_pane_id: String,
+    actual_window_id: String,
+    session: String,
+    last_capture: Option<String>,
+    closed: bool,
+}
+
+type TrackedPanes = Arc<Mutex<HashMap<String, TrackedPane>>>;
+type ControlWriter = Arc<Mutex<io::BufWriter<io::Stdout>>>;
+
+pub(crate) fn run_control_mode(
+    namespace: Option<&str>,
+    attached_session: Option<&str>,
+    no_echo: bool,
+) -> io::Result<()> {
+    let tracked: TrackedPanes = Arc::new(Mutex::new(HashMap::new()));
+    let writer: ControlWriter = Arc::new(Mutex::new(io::BufWriter::new(io::stdout())));
+    let stop_notifications = Arc::new(AtomicBool::new(false));
+    let _notification_thread = start_notification_thread(
+        namespace.map(str::to_string),
+        tracked.clone(),
+        stop_notifications.clone(),
+        writer.clone(),
+    );
+
+    {
+        let mut out = writer
+            .lock()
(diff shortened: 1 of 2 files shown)


psmux:
Psmux already has support for CC but is this a fix for some issue in it? Does it work or totally not work for you? Are you using it with ITerm?

psmux:
Hey, thank you for putting real time into this. Control mode is exactly the kind of tmux parity that makes psmux useful with existing tooling, so I appreciate you going after it.

I owe you a heads up though: psmux already ships control mode, and it landed after this PR was opened back in May, so you could not have known. That is on the repo, not on you.

Here is what is already in `master` today:

* `src/control.rs` is the server side protocol. It emits real notifications as state changes (`%output`, `%window-add`, `%window-close`, `%window-renamed`, `%window-pane-changed`, `%layout-change`, `%session-changed`, `%sessions-changed`, `%pane-mode-changed`, `%continue`, `%pause`, `%extended-output`, `%subscription-changed`, `%exit`, and more), with tmux compatible octal output escaping and a layout serializer whose checksum matches tmux's own `layout-custom.c`.
* `psmux -C` and `psmux -CC` are wired up in `main.rs` via `run_control_mode`, connecting over the existing authenticated TCP channel.
* iTerm2 is supported specifically. `emit_initial_state` sends the initial burst in the exact order iTerm's `TmuxGateway` needs (`%window-add` first, then `%sessions-changed`, then `%session-changed` last, since that is the one that flips iTerm into the writable state and kicks off window restoration). The client side also handles the `\x1bP1000p` DCS opener, ConPTY over SSH corruption bypass, the bare `\r` command terminator, and the closing `%exit` plus ST.

The tricky difference is the design. This PR drives control mode from the client by spawning a fresh psmux process per command and polling `capture-pane` every 400ms to synthesize `%output`. That was a reasonable path to get something working, but polling loses fast output between ticks, adds latency and CPU, and cannot reproduce the real layout and subscription notifications that an integration like iTerm relies on. The server push model in `src/control.rs` avoids all of that because the notifications come straight from the state changes.

Because of that overlap I am going to close this one. Please do not read that as the work being unwelcome. If you want to keep going in this area, the highest value follow ups would be things like extra `%subscription` coverage, `send-keys` fidelity through the control channel, or an iTerm integration test, and I would happily review those. Give `src/control.rs` and the `-CC` path a look first so we build on what is there.

Thanks again for contributing, and sorry the timing worked out this way.

PR #307, opened 2026-05-20
An improvement in send_ctrl_c_event so it's less likely to drop the ctrl event

catoxliu:
Wait before FreeConsole to make it less likely to drop the Ctrl event.

    GenerateConsoleCtrlEvent dispatches the CTRL_C_EVENT asynchronously
    via a system thread pool. The original 5ms sleep before FreeConsole()
    was often too short for the async dispatch to complete, causing the
    signal to be lost before reaching the target console's process group.

   Moved it before FreeConsole() so we
    remain attached to the child's console while the async thread finishes.
    psmux itself is protected during this window by the preceding
    SetConsoleCtrlHandler(None, 1) which ignores the signal.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -1117,17 +1117,26 @@ pub mod mouse_inject {
             let err = GetLastError();
 
             log(&format!("GenerateConsoleCtrlEvent => ok={} err={}", ok, err));
+            
+            // Brief sleep to let the async CTRL_C_EVENT handler thread finish
+            // before we re-enable default handling.
+            std::thread::sleep(std::time::Duration::from_millis(5));
 
+            // GenerateConsoleCtrlEvent is Asynchronous delivery
+            // it returns as soon as the signal is queued for delivery, not when it's actually handled by the target process.
+            // FreeConsole will reset the table of control handlers in the client process.
+            // If current process finishes detaching before the console subsystem finishes distributing the event,
+            // the context tracking who generated the signal is destroyed, 
+            // or the signal execution queue for that instance is aborted.
+            // So the sleep above won't guarantee no race condition here,
+            // in the worst case, the signal might be lost and never delivered to the child process.
+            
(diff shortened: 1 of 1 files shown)


psmux:
Hey @catoxliu, thanks for this! Great catch on the ordering.

You're absolutely right that sleeping AFTER `FreeConsole()` is logically wrong for signal delivery purposes. `GenerateConsoleCtrlEvent` dispatches asynchronously, and detaching before the console subsystem finishes distributing the event risks losing the signal context. Moving the sleep before `FreeConsole()` keeps us attached while the async dispatch completes, which is the correct sequence.

I ran a 50 iteration stress test (rapid fire `send-keys C-c` against a running ping process) and the fix passes cleanly with no regressions. 50/50 signals delivered, 100% delivery rate. Also tested with Python processes and TUI attached sessions.

Applied and pushed in [9346519](https://github.com/psmux/psmux/commit/9346519) with your Co-authored-by credit.

Thanks again for the contribution, Shijian!

psmux:
Hey @catoxliu, thanks for this! Great catch on the ordering.

You're absolutely right that sleeping AFTER `FreeConsole()` is logically wrong for signal delivery purposes. `GenerateConsoleCtrlEvent` dispatches asynchronously, and detaching before the console subsystem finishes distributing the event risks losing the signal context. Moving the sleep before `FreeConsole()` keeps us attached while the async dispatch completes, which is the correct sequence.

I ran a 50 iteration stress test (rapid fire `send-keys C-c` against a running ping process) and the fix passes cleanly with no regressions. 50/50 signals delivered, 100% delivery rate. Also tested with Python processes and TUI attached sessions.

Applied and pushed in [9346519](https://github.com/psmux/psmux/commit/9346519) with your Co-authored-by credit.

Thanks again for the contribution, Shijian!

psmux:
Closing as the fix has been applied in commit 9346519. Thanks @catoxliu!

PR #311, opened 2026-05-23
feat: add `#{pane_last_input}` — read-only live human-input signal

quazardous:
Minimal follow-up to #309. That PR (an option-driven file marker) was rightly declined as too much policy/lifecycle for core. This is the **smallest primitive** that still unblocks the use case, addressing each concern raised there.

## What

One **read-only per-pane format variable**, `#{pane_last_input}` = **ms since the last printable human keystroke** into that pane (empty until the first one).

```powershell
psmux display-message -t dev -p '#{pane_last_input}'   # "740", or "" if none yet
```

## Why / how it answers the #309 objections

- **No file, no lifecycle** — it's one `Option<Instant>` on the `Pane`, freed with the pane.
- **No policy / heuristic in core** — core only exposes a timestamp; the consumer decides ("value < N ms" = typing now).
- **Per-pane**, not session-level (the multi-pane gap you flagged).
- **Human typing only** — set in `forward_key_to_active` (the sole client-keystroke path); `send-keys`/`send-paste` go through `send_text_to_active`, so injected input never updates it. Only printable text counts (`is_human_text_key`); Enter/navigation/shortcuts and Ctrl/Alt chords are excluded. This is the separation `capture-pane` can't make.

That's the bit that's impossible to build outside core today: no hook or control-mode event fires on input, and `bind-key` consumes the key (fine for one chord, unworkable per-keystroke). With this primitive, the full feature can live as a plugin/external tool — psmux owns none of the policy.

## Diff

One field on `Pane` + its inits; ~6 lines in `forward_key_to_active` (+ a small `is_human_text_key` helper); one `#{pane_last_input}` arm in the format engine; a classifier unit test; a doc note.

## Tests / validation

- `cargo test` — classifier unit tests (`tests-rs/test_pane_last_input.rs`); `cargo clippy` clean.
- Verified live: the var is recognized and **stays empty after `send-keys`** (injected input correctly ignored — the separation in action). The positive path (a real keystroke populating it) needs an attached client, so it's covered by the unit test + the single-line set site.


tarikguney:
Thanks for the follow-up, @quazardous . This is closer. Before merge, I think these are must-have so the contract is clear:

- Please frame this as a **text input route signal**, not human detection.
- Please remove “human” wording from **all names and text** (field names, format name, comments, docs, tests). The naming should describe route/class only.
- Please define it explicitly: interactive text route = `handle_key -> forward_key_to_active` (updates), injected text route = `send-keys` / `send-paste` / `send-text` (does not update).
- Please document key scope clearly: printable text counts; Enter/arrows/shortcuts do not.
- Please add one caveat: if a bot sends real key events through the interactive route, this signal will also update.
- Please keep tests strict to this contract (interactive positive path, injected negative path, key-classifier boundaries).

Main ask: the full PR language and naming should reflect “text input route signal,” not “human/activity detection.”

CC: @psmux

quazardous:
Done, pushed. Reframed entirely as a **text-input route signal**:

- Removed all "human" wording from names/text. `last_human_input` → `last_text_input`, `is_human_text_key` → `is_text_input_key`, and the format var → **`#{pane_last_text_input}`**.
- Documented the contract: interactive route (`handle_key → forward_key_to_active`) updates it; injected route (`send-keys` / `send-paste` / `send-text` → `send_text_to_active`) does **not**. Printable text only — Enter/arrows/shortcuts/`Ctrl`/`Alt` excluded. Added the caveat that a bot injecting *real key events* through the interactive route also updates it (measures the route, not the actor).
- Also fixed @Copilot's catch: stamp exactly the panes that **receive** the key — every non-dead pane under sync-input, else the active pane only if alive — instead of always the active pane even when dead.

On tests: the key-classifier boundaries are unit-tested strictly (printable vs control/Ctrl/Alt/nav/F-key). The route positive/negative isn't unit-testable in the current setup — the test suite uses empty windows (no lightweight `Pane`; a real one needs a PTY/child), so I can't drive `forward_key_to_active` vs `send_text_to_active` from a unit test. I verified the injected-negative via integration (`display-message -p '#{pane_last_text_input}'` stays empty after `send-keys`); route separation is otherwise structural (only `forward_key_to_active` writes the field). Happy to add a process-spawning integration test if you'd prefer that over the structural guarantee.

quazardous:
Hi, thx for considering this little contribution :)

tarikguney:
Thanks for the follow-up, @quazardous — this is in good shape now.

You addressed the core asks in code:
- reframed as a **text-input route signal** (not human/activity detection),
- renamed symbols to the route-based contract (`last_text_input`, `is_text_input_key`, `#{pane_last_text_input}`),
- clarified interactive vs injected route behavior,
- tightened key-scope boundaries,
- and fixed sync/dead-pane stamping so only receiving panes are updated.

I’m approving this direction.

One final cleanup request before merge: please align the **PR title/body** with the updated contract (they still contain old `human` / `pane_last_input` wording). After that, this is good to go from my side.

CC @psmux: implementation looks aligned and approved pending that metadata text cleanup.

PR #314, opened 2026-05-24
Add KKTerm to related projects

ryantsai:
Adds KKTerm to the related projects section as a Windows-first, local-first terminal for managing connections, sessions, and workspaces.


psmux:
Does it integrate with Psmux?

PR #316, opened 2026-05-25
fix: client freezes permanently when server write timeout fires (Windows)

LizardLiang:
## Problem

Running multiple long-context Claude Code sessions simultaneously triggers a permanent client freeze. The server pushes large JSON frames (~50–100 KB each); when the client is CPU-bound rendering, the TCP receive buffer fills and the 5-second write timeout fires on the server writer thread. The client then displays the last rendered frame indefinitely — keyboard input still works, but the screen never updates. Both client and server are alive and responsive; only the display is stuck.

### Layer 1 — writer-thread deadlock

The old writer thread held `frame_chan.rx.lock()` while doing blocking TCP writes. `push_frame()` on the server main loop also tried to acquire this lock when the channel was full → both sides stalled. The server could not push new frames; the writer could not flush old ones.

**Fix:** drain the frame channel into a local `Vec` while holding `rx.lock()`, release the lock, then write to TCP with no lock held. `push_frame()` now uses `try_lock()` so the server main loop is never blocked by the writer thread.

### Layer 2 — missing EOF signal on writer exit (Windows-specific)

On Windows, `TcpStream::try_clone()` creates a new OS handle to the same underlying socket. When the write timeout fires and the writer thread exits, it drops `ws_bg` (a clone). Because `write_stream` in the connection handler still holds a reference, the OS socket stays open. The client reader thread never receives EOF, never signals `Disconnected`, and the client hangs indefinitely — even though both sides are alive.

**Fix:** a RAII `Guard` on the writer thread calls `shutdown(Both)` on exit. Unlike `drop`, `shutdown()` operates at the socket level and sends FIN to the client regardless of how many handles exist. Client reader gets `Ok(0)` → `frame_tx` drops → `Disconnected` → reconnect.

## Additional hardening

| Change | Why |
|---|---|
| `rrx.recv()` → `rrx.recv_timeout(5s)` | A stalled command handler could park the writer thread forever with no timeout |
| Write timeout on client writer (5s) | Matches server side; unblocks the client if *it* is the slow writer |
| Background reconnect thread | On `Disconnected`, retries up to 5× (immediate + 500 ms / 1 s / 1.5 s / 2 s backoff) on a background thread so the event loop stays responsive (Ctrl+C, resize) during the retry window |
| `ws_shutdown` cloned from `write_stream` not `ws_bg` | Shallower clone chain; `Err => return` makes fd-exhaustion failure explicit rather than a silent `Option::None` that silently skips `shutdown` |
| `deregister_frame_channel` / `deregister_persistent_stream` | Server stops pushing to dead channels immediately on writer exit rather than waiting for the next `push_frame` `Disconnected` detection |

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -877,6 +877,78 @@ struct ClientDragState {
     total_pixels: u16,
 }
 
+/// Connect to the server, authenticate, enter persistent mode, spawn the reader
+/// thread, and return a (writer, frame_rx) pair ready for the event loop.
+/// Sets a 5-second write timeout so blocked writes never freeze the client.
+fn establish_connection(addr: &str, key: &str) -> io::Result<Connection> {
+    let stream = std::net::TcpStream::connect(addr)?;
+    stream.set_nodelay(true)?;
+    let mut writer = stream.try_clone()?;
+    writer.set_nodelay(true)?;
+    writer.set_write_timeout(Some(Duration::from_secs(5)))?;
+    let mut reader = BufReader::new(stream);
+
+    let _ = writer.write_all(format!("AUTH {}\n", key).as_bytes());
+    let _ = writer.flush();
+    let mut auth_line = String::new();
+    reader.read_line(&mut auth_line)?;
+    if !auth_line.trim().starts_with("OK") {
+        return Err(io::Error::new(io::ErrorKind::PermissionDenied, "auth failed"));
+    }
+
+    let _ = writer.write_all(b"PERSISTENT\n");
+    let _ = writer.write_all(b"client-attach\n");
+    let _ = writer.flush();
+
+    // 2-second read timeout keeps the thread from blocking forever on process exit.
+    let _ = reader.get_ref().set_read_timeout(Some(Duration::from_secs(2)));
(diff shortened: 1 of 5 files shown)


psmux:
Hey @LizardLiang, this is a really solid contribution. Thank you for diving deep into the root cause here.

The two layer analysis is spot on. The `rx.lock()` held during blocking TCP writes creating a deadlock with `push_frame()` is exactly the kind of subtle concurrency bug that only shows up under real world load. And the Windows specific EOF issue with cloned `TcpStream` handles is a great catch. I would not have found that easily.

The RAII `Guard` pattern for `shutdown(Both)` is clean and the right approach. `try_lock()` on the server main loop so it never blocks is the correct fix for Layer 1. The additional hardening (`recv_timeout` instead of blocking `recv`, client write timeout, background reconnect with backoff, immediate deregistration of dead channels) all make sense and improve resilience across the board.

I verified there are no conflicts with recent changes on master and the architecture looks sound. Merged.

Thanks again for taking the time to trace this through both layers and write up the detailed explanation. Much appreciated!

PR #320, opened 2026-05-25
Strip trailing newline from list-sessions format output

### LINKED ISSUE #319: Strip trailing newline from list-sessions format output
`list-sessions -F` currently returns a formatted line with an extra trailing newline in the payload. libtmux parses that newline into the last field, which breaks tmuxp/libtmux compatibility.
The server should return the raw formatted line only, and let the socket writer append the transport newline once.
Needed for #318.

zandivx:
Fixes #319.


zandivx:
Obsolete

PR #322, opened 2026-05-25
Fall back to global window options

### LINKED ISSUE #321: Fall back to global window options
`show-options -w` can return empty for options that are not actually window-scoped, like `pane-base-index`. libtmux expects a value here and tmuxp compatibility breaks when the fallback is missing.
If a window-scope lookup is empty, fall back to the global option value instead.
This is needed for #318.

zandivx:
Fixes #321.


zandivx:
Fixed via [c99ca5b](https://github.com/psmux/psmux/commit/c99ca5b30bc7352133e6f51ed2cdc6ad6a37a427).

PR #326, opened 2026-05-27
fix: capture-pane CLI expands POSIX short flag clusters

quazardous:
## Summary

`psmux capture-pane -ep -t <target>` returned exit 0 with empty stdout, while `psmux capture-pane -e -p -t <target>` (same flags, separated) worked correctly.

## Root cause

The CLI dispatcher for `capture-pane` in `main.rs` matched flag strings exactly (`"-p"`, `"-e"`, `"-J"`), so a POSIX short flag cluster like `-ep` or `-pe` fell through to `_ => {}` and was silently dropped. As a result `print_stdout` stayed `false` and the command was routed to the fire-and-forget `send_control` branch — no response was printed, hence the empty stdout and exit 0.

## Fix

Add a `match` arm that detects an argument of the form `-XYZ` whose every character after the dash is one of the capture-pane boolean shorts (`p`, `e`, `J`), and emit each flag individually. Value-taking flags (`-t`, `-S`, `-E`, `-b`) take a following argument and are not eligible for clustering — they stay on their existing arms.

## Smoke matrix (attached 29×120 session)

| Form    | Before | After |
|---------|--------|-------|
| `-p`    | 1281   | 1281  |
| `-e -p` | 2312   | 2312  |
| `-ep`   | **0**  | **2312** |
| `-pe`   | **0**  | **2312** |

Zero regression on the separated forms; the cluster form now matches the separated-flag output and contains ANSI SGR escapes as expected for `-e`.

## Test plan

- [x] Repro confirmed pre-patch (`-ep`/`-pe` → 0 bytes ; `-p`/`-e -p` → text)
- [x] 4-form smoke matrix verified post-patch
- [ ] Adding a Rust unit test for the cluster expansion would be a reasonable follow-up; current behavior validated by the integration smoke above

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1183,6 +1183,18 @@ fn run_main() -> io::Result<()> {
                                 i += 1;
                             }
                         }
+                        a if a.len() > 2
+                            && a.starts_with('-')
+                            && !a.starts_with("--")
+                            && a.chars().skip(1).all(|c| matches!(c, 'p' | 'e' | 'J')) =>
+                        {
+                            // POSIX cluster of capture-pane booleans (-ep, -pe, -pJ, -eJ,
+                            // -epJ, ...). -t/-S/-E/-b take a value, so they are NOT
+                            // eligible for clustering.
+                            if a.contains('p') { cmd.push_str(" -p"); print_stdout = true; }
+                            if a.contains('e') { cmd.push_str(" -e"); }
+                            if a.contains('J') { cmd.push_str(" -J"); }
+                        }
                         _ => {}
                     }
                     i += 1;


psmux:
Hey @quazardous, another solid catch. Applied in [72d08aa](https://github.com/psmux/psmux/commit/72d08aa) with your Co-authored-by credit.

**Bug confirmed before patch:**

`-p=214  -ep=0  -pe=0  -e -p=432`

POSIX flag clusters silently fell through to `_ => {}` in the CLI dispatcher, leaving `print_stdout = false` and routing to the fire-and-forget branch. Empty stdout, exit 0. Nasty because it looks like success.

**After patch:**

`-p=130  -ep=330  -pe=330  -e -p=330  -pJ=130  -epJ=330`

All cluster combinations now match their separated equivalents.

**What I verified (19/19 E2E tests pass):**

- `-ep`, `-pe`, `-pJ`, `-Jp`, `-epJ`, `-Jpe` all produce identical output to their separated forms
- Content integrity: marker text present in all cluster outputs
- ANSI SGR escapes present when `-e` is in the cluster
- `-eJ` without `-p` correctly produces no stdout (fire-and-forget)
- Invalid cluster `-px` silently ignored (`x` not in {p,e,J})
- Value-taking flag `-t` not affected by cluster logic
- TCP server path works correctly (separated flags)
- TUI visual verification: attached session, `-ep` matches `-e -p`
- Full regression: 2110 Rust tests pass, 0 failures

**tmux parity note:** tmux natively supports POSIX flag clustering for ALL commands through its `args_parse_flags()` character-by-character loop in `arguments.c`. This fix brings `capture-pane` to parity. Other commands may have the same gap but that's a separate effort.

Thanks David, two for two this week.

psmux:
Applied in commit 72d08aa with Co-authored-by credit. Thanks @quazardous!

Issue #211, opened 2026-04-13
Windows 11 PowerShell-style mouse selection parity


jun2077681:
## Summary

The client-side `rsel` drag selection (blue overlay, pwsh-style) has a few behavioural gaps compared to Windows 11's native PowerShell selection in Windows Terminal:

1. **Pane boundary bleed**: the overlay spans the full terminal width on intermediate rows, so dragging inside a pane visually bleeds into neighbouring panes on multi-line selections.
2. **Forced copy-on-release**: `rsel` always copies to the clipboard and clears the selection on mouse-up. Windows 11 PowerShell keeps the selection visible until an explicit copy action (right-click, `Ctrl+C`, `Ctrl+Shift+C`), letting users refine the selection or dismiss it with a click.
3. **No multi-click word/line selection**: double-click / triple-click do nothing special. Native terminals expand to a word / line.
4. **No modifier-extend**: Windows 11 uses Shift+click to extend an existing selection — unfortunately Windows Terminal reserves Shift+click for its own native selection override, so this needs a different modifier (Ctrl+click is a natural fallback).
5. **No block selection**: Alt+drag does not produce a rectangular selection.
6. **No `Ctrl+Shift+C` / `Ctrl+Shift+V` / smart `Ctrl+C`** bindings tied to the `rsel` state.

Item 1 is a latent rendering bug (intermediate rows iterate `0..=area.width-1` regardless of pane rect). Items 2-6 are behaviour the existing `rsel` path was never designed for but are expected by users coming from Windows 11 PowerShell.

## Proposal

Add a new `pwsh-mouse-selection` session option (boolean, default `off`) that opts into Windows 11 PowerShell-parity behaviour:

- Pane-aware clipping (fixes #1 for anyone with the option on)
- Selection persists after mouse-up; copy only on right-click / `Ctrl+Shift+C` / smart `Ctrl+C`
- Double-click word / triple-click line selection (within the originating pane)
- `Ctrl+click` extends the current selection to the click point
- `Alt+drag` produces a rectangular (block) selection
- `Ctrl+Shift+C`, `Ctrl+Shift+V`, and smart `Ctrl+C` (copy-if-selected, SIGINT otherwise) key bindings
- White overlay (closer to Windows Terminal's native selection colour)

Default stays `off` so existing users see zero behavioural change.

## Environment

- psmux version: 3.3.2
- Terminal: Windows Terminal
- OS: Windows 11

## Notes

[...]

PR #327, opened 2026-05-27
fix: clip mouse selection copy to pane

burakkose:
## Summary
Fixes `pwsh-mouse-selection` clipboard extraction so multi-line mouse selections are clipped to the originating pane, matching the visible selection overlay.

Previously, the overlay was pane-clipped but `Ctrl+C`, `Ctrl+Shift+C`, right-click copy, and legacy copy-on-release could still extract full terminal-width intermediate rows, which could include text from neighboring panes.

Related to #211 

## Approach

 - Thread the originating pane rect into mouse-selection text extraction.
 - Intersect each selected row with that pane rect before reading cells from the layout tree.
 - Preserve existing behavior when no pane clip is provided.

## Tests

- `cargo test -q extract_selection_text`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -196,6 +196,45 @@ fn normalize_selection(start: (u16, u16), end: (u16, u16), block: bool) -> (u16,
     }
 }
 
+fn selection_row_cols(
+    row: u16,
+    r0: u16,
+    c0: u16,
+    r1: u16,
+    c1: u16,
+    block: bool,
+    term_width: u16,
+    pane_clip: Option<Rect>,
+) -> Option<(u16, u16)> {
+    if term_width == 0 {
+        return None;
+    }
+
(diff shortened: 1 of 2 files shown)


psmux:
Nice catch @burakkose, and solid implementation.

Verified this against tmux's behavior. In tmux, each pane operates on its own backing screen (sized to the pane, not the terminal), so multi line selections naturally stay within pane boundaries. psmux was doing the overlay clipping correctly (the visual highlight was confined to the pane) but the actual text extraction for clipboard was reading the full terminal width on intermediate rows.

Your approach of threading the originating pane rect into xtract_selection_text and intersecting each row with that rect is clean and correct. The selection_row_cols helper is well structured. Tests cover both reading order and block mode across a split boundary, which is exactly what was needed.

All 2030 existing unit tests pass with zero regressions. Merged.

Thanks for the contribution!

PR #334, opened 2026-05-28
fix: copy pwsh selection on mouse release

burakkose:
## Summary

Copies `pwsh-mouse-selection` text automatically when the mouse button is released, then clears the transient client-side selection highlight.

This restores the tmux-like copy-on-release part of the workflow while keeping the pane-clipped extraction from #327, so multi-line selections copy only text from the originating pane instead of pulling text from neighboring panes.

Related to #211.

## Approach

- On mouse release, extract the current pwsh mouse selection using the originating pane rect.
- Copy non-empty selected text to the system clipboard.
- Queue the copied text for OSC52 clipboard propagation.
- Clear the transient client-side selection after copying.
- Preserve existing behavior for non-`pwsh-mouse-selection` paths.

## Tests

- `cargo build -q`
- `cargo test -q extract_selection_text`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -104,7 +104,7 @@ psmux split-window -- "C:/Program Files/Git/bin/bash.exe"
 | `mouse` | Bool | `on` | Mouse support |
 | `mouse-selection` | Bool | `on` | psmux's client-side drag selection. Set `off` to let in-pane TUI apps (opencode, nvim, etc.) handle their own mouse selection without psmux drawing on top |
 | `scroll-enter-copy-mode` | Bool | `on` | Enter copy mode on mouse scroll (set `off` to disable) |
-| `pwsh-mouse-selection` | Bool | `off` | Windows 11 PowerShell-style word/line selection (double/triple-click) |
+| `pwsh-mouse-selection` | Bool | `off` | tmux-like release-copy selection with word/line multi-click and pane-clipped extraction |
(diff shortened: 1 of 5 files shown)


tarikguney:
Thanks for the contribution, @burakkose — this is a useful improvement.

Current `pwsh-mouse-selection` behavior in `psmux` was originally modeled after Windows 11 / Windows Terminal selection UX: selection can persist after drag, and copy happens via explicit action (right-click / copy shortcut).  
tmux copy-mode mouse UX is different: drag selection is typically copied on mouse release and then cleared (`copy-pipe-and-cancel` flow).

So while both are “mouse selection,” the semantics differ:
- Windows-style: **select first, copy explicitly**
- tmux-style: **release triggers copy + clear**

This PR re-orients `pwsh-mouse-selection` toward tmux-compatible expectations by copying on left-button release while keeping pane-clipped extraction. That improves consistency for users expecting tmux-like mouse copy behavior in this path.

I also pushed follow-up updates in this PR to keep behavior and docs aligned:
- updated contradictory inline comment in `src/client.rs` to describe release-copy semantics
- updated `docs/configuration.md` option text and mouse section to explicitly describe left-release copy + clear behavior
- updated `docs/features.md` wording to reflect tmux-like release-copy behavior
- updated `docs/faq.md` mouse answer to include this mode

Really appreciate the implementation from @burakkose and the care taken to preserve pane clipping while adjusting selection lifecycle behavior.

Issue #345, opened 2026-06-01
Panic when entering Chinese text in command-prompt rename window input


TimFruit:
## Description

`psmux` crashes when using an interactive `command-prompt` binding to rename a window with Chinese characters.

The window name itself appears to support Chinese text when using the CLI command directly, so this seems specific to the interactive prompt/input handling path.

## Environment

  - psmux version: `3.3.5`
  - Install method: WinGet
  - OS: Windows 10 Pro x64
  - Shell: PowerShell 7.6.2

## Config

Relevant binding:

```tmux
set -g prefix C-a
bind-key , command-prompt -I '#W' 'rename-window "%%"'
```

## Steps to Reproduce

1. Start `psmux`.
2. Press `Ctrl+a ,` to open the rename-window prompt.
3. Enter a Chinese window name, for example:

   ```text
   中文窗口
   ```

4. Press `Esc` or otherwise cancel/edit the prompt.

## Actual Behavior

`psmux` crashes and returns to PowerShell with this panic:

```text
thread 'main' (...) panicked at src\client.rs:2895:157:
assertion failed: self.is_char_boundary(idx)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Expected Behavior

The prompt should either:

- correctly handle UTF-8 / non-ASCII input such as Chinese characters, or
- reject unsupported input gracefully without panicking.

## Additional Notes

Directly renaming the window from PowerShell works:

```powershell
psmux rename-window '中文窗口'
```

This suggests the issue is likely in the interactive prompt text editing logic, possibly using byte indices as character indices for UTF-8 strings.

psmux:
Hey, thanks so much for the detailed repro and the screenshot. That made this very fast to confirm.

Reproduced exactly: opening the command prompt with `prefix :` and typing CJK characters (or any multi byte UTF 8 character, including most emoji) panicked the client with a `byte index N is not a char boundary` error and tore down the session.

Root cause: the command prompt was tracking the cursor as a character count but using it as a byte index into the underlying `String`. The first multi byte character left the cursor sitting in the middle of a UTF 8 sequence, and the very next keystroke (insert, backspace, or arrow key) hit a `String::insert` / `String::replace_range` / slice operation on a non boundary and panicked.

Fix in c0c0ac1:

* Cursor is tracked in bytes everywhere it touches the buffer.
* Insert advances by `c.len_utf8()`.
[...]

PR #347, opened 2026-06-03
fix: correct zoom pane sizing for PTY resize and active rect highlight

Edemilorhea:
## Problem
When a pane is zoomed, the non-active pane (bottom pane in horizontal split, right pane in vertical split) would appear visually shifted by a few pixels in height/width. The border highlight and cursor position of the zoomed pane were also slightly misaligned.

## Root Cause
Two separate but related issues:
1. Client-side (src/client.rs)
compute_active_rect_json was used to compute active_rect for border highlighting and cursor positioning. This function internally calls split_with_gaps, which subtracts 1px gap per separator and returns child rects with offsets applied. However, when a pane is zoomed, the renderer passes the full content_chunk area to the child — ignoring gaps. The mismatch between what the renderer drew and what active_rect computed caused the visual shift.
2. Server-side (src/tree.rs)
In resize_all_panes, when a pane is zoomed, the PTY was being resized to the gap-reduced rect instead of the full viewport area. This meant the zoomed pane's PTY dimensions didn't match what was actually rendered on screen.

## Solution
src/client.rs — Replace compute_active_rect_json with the zoom-aware variant:

// Before
let active_rect = compute_active_rect_json(&root, content_chunk);

// After
let active_rect = compute_active_rect_json_zoom_aware(&root, content_chunk, state.zoomed);

When state.zoomed is set, compute_active_rect_json_zoom_aware returns the full content_chunk directly, matching what the renderer actually uses.

src/tree.rs — In resize_all_panes, when win.zoom_saved.is_some(), override the active pane's rect to the full area before sending the PTY resize signal, so the zoomed pane gets the correct full-viewport dimensions.

## Impact

- Zoomed pane now fills the full viewport with correct PTY dimensions
- Border highlight and cursor position align correctly when zoomed
- No regression to normal (non-zoomed) split behavior — the zoom-aware path only activates when state.zoomed is set

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -4333,7 +4333,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
             let mut border_path = Vec::new();
             collect_layout_borders(&root, content_chunk, &mut border_path, &mut client_borders);
 
-            let active_rect = compute_active_rect_json(&root, content_chunk);
+            let active_rect = compute_active_rect_json_zoom_aware(&root, content_chunk, state.zoomed);
             let clock_col = clock_colour_str.as_deref().map(|s| map_color(s)).unwrap_or(Color::Cyan);
             let border_status = state.pane_border_status.as_deref().unwrap_or("off");
             let border_format = state.pane_border_format.as_deref().unwrap_or("");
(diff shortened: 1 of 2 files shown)


psmux:
Hey @Edemilorhea, thanks for the careful diagnosis on this one and for tracking down both the active_rect highlight offset AND the PTY sizing mismatch in resize_all_panes. Both intentional hunks landed on master in 6a20f48 (rebased onto current main so the recent #345 UTF-8 cursor fix and the zoom_cursor_rect work from 0aba873 stay intact). I added a small follow-up regression test in a8cf04c that asserts the active zoomed pane reports the full 120x40 viewport (was 118x38 before your fix because of the gap + min-size steal you correctly identified). Closing this PR since the rebase + merge is already in. Really appreciate the contribution.

PR #353, opened 2026-06-04
fix(server): Simplify frame buffer and avoid starvation

MattKotsenas:
## Summary

The persistent-client frame-push path between the server main loop and each per-client writer thread is currently a bounded `sync_channel(16)` wrapped in a mutex. #316 fixed the biggest issue with this design (holding lock across TCP writes), but two issues remain.

### Frame starvation

The writer thread's response-processing block ends with `continue`, skipping the frame-drain step whenever any command response was pending. Under heavy activity the push path starves (this is the original #224 issue that motivated #267).

### Buffering complexity

The bounded channel was added to preserve intermediate frames during fast typing. However, `push_frame`'s drain-on-Full path discards all queued frames the moment the channel fills, so under sustained load the buffering doesn't behave any differently than a single slot. The bounded channel + drain-on-Full + `Mutex<Receiver>` adds notable complexity for behavior that is approximately single-slot in practice.

## Fix

Replace the channel with a single-slot `Arc<Mutex<Option<String>>>`.

The producer locks, replaces, and unlocks. It never holds the lock across I/O. The consumer locks, takes, unlocks, and _then_ writes to TCP outside the lock.

Because the lock is never held during I/O, lock contention is no longer an issue. Further, because frames are full snapshots, a superseded frame has no value. That means that overwriting unconsumed frames is correct by construction. At most 3 frames are in play at once: 1 being created on the main loop (shared across clients), 1 ready in each client's slot, and 1 in flight on each client's writer thread.

## Tests

Includes regression tests that push a burst of frames without an intervening read.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -259,16 +259,14 @@ if line.trim() == "PERSISTENT" {
     let _ = ws_bg.set_write_timeout(Some(Duration::from_secs(5)));
     let (resp_tx, resp_rx) = mpsc::channel::<mpsc::Receiver<String>>();
 
-    // Register a bounded frame channel for server-pushed frames (event-driven
-    // rendering).  The channel queues up to FRAME_CHANNEL_CAPACITY frames,
-    // allowing short bursts (e.g. fast typing) to be delivered without dropping
-    // intermediate states, while still bounding memory for sustained throughput
-    // scenarios (e.g. rapid scroll in copy mode).
-    let frame_chan = crate::types::register_frame_channel(client_id);
+    // Register a frame slot for server-pushed frames (event-driven rendering).
+    // Slot holds at most one pending frame; push_frame() overwrites any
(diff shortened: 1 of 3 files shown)


psmux:
Merged. Thank you for this, and for the unusually clear writeup.

I went through both halves before taking it. The `continue` at the end of the response block is a real starvation point: once a command response is handled, the loop skips the frame drain for that iteration, so a client under steady response load stops getting pushed frames. Draining both sources every iteration is the right call.

On the single-slot replacement, I checked the concern that motivated the older bounded channel (preserving intermediate frames during fast typing). Because frames are full snapshots, the latest one always carries the complete state, so a superseded frame genuinely has no value, and the swap chain keeps storage at O(clients) while never holding the lock across a write. Your backpressure proof tests pass and the full suite is green on my end.

Appreciate the careful work here. Thanks again.

PR #354, opened 2026-06-04
Fix display-popup default shell

### LINKED ISSUE #351: Windows 10: `Prefix + P` opens an empty popup and does not accept command input
## Description
  On Windows 10, `Prefix + P` opens an empty popup. The popup does not show a shell prompt or command prompt, and typed commands are not displayed or
  executed.

  The popup only shows:

  ```text
  Press 'q' or Escape to close
 ```

  ## Steps to Reproduce

  1. Use the following minimal config:

  unbind C-b
  set -g prefix C-a
  bind-key C-a send-prefix

  bind-key P display-popup -E -w 80% -h 80%

  2. Start psmux with this config.
  3. Press Ctrl+a, then P.

  ## Actual Behavior

  An empty popup appears. It only shows:

  Press 'q' or Escape to close

  The popup does not accept command input. Only q or Escape closes it.

  ## Expected Behavior

  Prefix + P should open a usable interactive popup, for example with a shell or command input available.

  ## Environment

  - OS: Windows 10
  - Shell: PowerShell 7.6.2
  - psmux --version: tmux 3.3.5
  - Prefix key: Ctrl+A

  ## Runtime Verification

  psmux show-options -g prefix
  prefix C-a

  psmux list-keys -T prefix P
  bind-key -T prefix P display-popup -E -w 80% -h 80%

  ## Screenshot

  Attached screenshot shows the empty popup after pressing Ctrl+a, then P.

Image

tarikguney:
Fixes #351.

## Summary
- Make `display-popup` without a command open a PTY-backed popup using the default interactive shell.
- Preserve existing command-backed popup behavior.
- Add a regression script covering no-command popup PTY creation and popup input routing.

## Validation
- `cargo build --release` passed locally.
- Local regression test was not run after user requested PR pipelines handle test execution.

### DIFF
--- a/src/popup.rs
+++ b/src/popup.rs
@@ -41,9 +41,19 @@ pub fn create_popup_pane(
     };
     let pair = pty_sys.openpty(pty_size).ok()?;
 
-    let mut cmd_builder = portable_pty::CommandBuilder::new(
-        if cfg!(windows) { "pwsh" } else { "sh" },
-    );
+    let mut cmd_builder = if command.trim().is_empty() {
+        crate::pane::build_command(None, true, false)
+    } else {
+        let mut builder = portable_pty::CommandBuilder::new(
+            if cfg!(windows) { "pwsh" } else { "sh" },
+        );
+        if cfg!(windows) {
+            builder.args(["-NoProfile", "-Command", command]);
+        } else {
+            builder.args(["-c", command]);
+        }
+        builder
+    };
     if let Some(dir) = start_dir {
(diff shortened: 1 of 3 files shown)


tarikguney:
Low-bandwidth summary:

`display-popup -E` with no command opened a static popup instead of a shell.

That static popup only showed the close hint. It had no PTY, so typed keys had nowhere to go.

This PR makes the empty-command popup start the default shell in a PTY. Command-backed popups keep the old behavior.

PR #357, opened 2026-06-05
fix: prevent Windows paste from submitting existing CLI input

tarikguney:
## Summary
When a Windows paste started with a newline/tab, `psmux` could send Enter/Tab immediately before the rest of the paste arrived. That submitted the current Copilot CLI input unexpectedly. This change buffers leading plain Enter/Tab as part of paste detection so the whole clipboard payload is delivered together.

## Details
In the Windows client input loop (`src/client.rs`), `KeyCode::Enter` and `KeyCode::Tab` were only appended to `paste_pend` when `paste_pend` was already non-empty. If the first injected key in a Ctrl+V burst was Enter/Tab, the code fell through to immediate `send-key`, causing premature command submission in CLI overlays before remaining paste bytes were processed.

This PR adds `should_buffer_leading_paste_control(...)` and uses it in Enter/Tab handling to buffer leading plain Enter/Tab when `paste-detection` is enabled. It also starts `paste_pend_start` when buffering begins. Behavior remains unchanged for modified Enter/Tab and when `paste-detection` is disabled.

Tests were added in `tests-rs/test_client.rs` to verify:
- leading plain Enter buffers when detection is on
- leading plain Tab buffers when detection is on
- leading Enter does not buffer when detection is off
- modified Enter does not buffer

## Test Plan
- [ ] Run Windows unit tests (`cargo test --test test_client`) and verify new helper tests pass
- [ ] Repro manually in Windows: type text in Copilot CLI input, paste content that starts with newline/tab, confirm existing input is not submitted prematurely

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3936,11 +3936,12 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
         // paste even though the user explicitly disabled paste detection.
         #[cfg(windows)]
         {
-            let flush_all = !paste_detection_enabled;
-            if !paste_confirmed && !paste_stage2
-                && paste_pend.len() >= 1
-                && (paste_pend.len() <= 2 || flush_all)
-            {
+            if should_zero_latency_flush_paste_pend(
+                &paste_pend,
+                paste_detection_enabled,
+                paste_confirmed,
+                paste_stage2,
+            ) {
                 if input_log_enabled() {
                     input_log("paste", &format!(
                         "zero-latency flush {} char(s) as typing",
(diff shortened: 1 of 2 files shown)


psmux:
Thanks for this @tarikguney. I reproduced the underlying problem on Windows and verified the fix end to end, then merged it (commit a2f4718).

**Reproduced the bug (before this PR)**

In a psmux pane I typed an invalid partial command and left the cursor on the line without pressing Enter, then pasted clipboard content that began with a newline. The leading pasted newline was forwarded as an immediate Enter, so the partial command was submitted early and the shell raised a "not recognized" error before the rest of the paste arrived. That is exactly the premature submit you described for the Copilot and Claude Code input.

**Verified the fix**

* All 10 new helper unit tests pass (`should_buffer_leading_paste_control` and `should_zero_latency_flush_paste_pend`).
* With a reader that enables bracketed paste the way Copilot CLI and Claude Code do, the leading newline now arrives INSIDE the bracketed paste payload (`ESC[200~` then `0x0D` then the text then `ESC[201~`), with no premature standalone Enter before the opener. So the pending input is no longer submitted early.
* The full suite stays green with no regressions.

The change is nicely scoped: leading plain Enter or Tab is buffered into the paste only when paste detection is on, modified Enter and Tab are untouched, and behavior is unchanged when paste detection is off.

I added tangible repro and verification scripts in commit 0741c87 so this stays covered going forward. Nice work, merged.

PR #364, opened 2026-06-08
Fix/send keys ctrlc force signal

Edemilorhea:
# Add opt-in force-signal mode for send-keys Ctrl+C
## Summary
This PR adds an opt-in force-signal mode for `send-keys`:
~~~tmux
bind -n C-F12 send-keys -f C-c
~~~
On Windows, `send-keys -f C-c` bypasses the raw-mode TUI heuristic and sends a `CTRL_C_EVENT` to the foreground process.
Users can configure this in `psmux.conf` with any key binding they prefer. `C-F12` is only an example.
## Motivation
Recent Ctrl+C handling preserves normal keyboard Ctrl+C behavior for raw-mode TUIs such as opencode or neovim. This is useful because those applications may want to handle raw `0x03` themselves.
However, there are still cases where users need a separate, explicit “force interrupt” binding. For example, a raw-mode TUI may consume Ctrl+C, or handle it in a way that does not reliably terminate the foreground process.
This PR provides a user-configurable escape hatch without changing the default keyboard Ctrl+C behavior.
## Behavior
- Normal keyboard Ctrl+C remains unchanged.
  - Raw-mode TUIs can continue handling Ctrl+C themselves.
- `send-keys C-c` without `-f` keeps the existing heuristic behavior.
- `send-keys -f C-c` explicitly forces a Windows `CTRL_C_EVENT`.
- `--force-signal` is also accepted as a long-form alias.
- Users can bind this command to any shortcut in `psmux.conf`.
## Example
~~~tmux
# Dedicated force interrupt binding for raw-mode TUIs.
# C-F12 is only an example; users can choose any preferred key.
bind -n C-F12 send-keys -f C-c
~~~
This is useful for applications like opencode, where plain Ctrl+C may be consumed by the application or may trigger undesirable exit behavior. With this explicit binding, users can reliably interrupt or terminate the foreground process while preserving normal Ctrl+C pass-through behavior.
## Changes
- Adds force-signal plumbing for `send-keys -f/--force-signal C-c`.
- Preserves the force-signal flag through CLI forwarding.
- Documents the new psmux extension.
- Adds regression coverage for both `-f` and `--force-signal`.
## Validation
Tested locally on Windows:
~~~powershell
cargo build
cargo test send_keys_flag_force_signal -- --nocapture
cargo test send_keys_flag_f_force_signal_short -- --nocapture
~~~

### DIFF
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -48,12 +48,22 @@ psmux send-keys -N 5 Up
 # Send copy mode command
 psmux send-keys -X copy-mode-up
 
+# Force Ctrl+C to the foreground process (Windows CTRL_C_EVENT)
+# Useful for a dedicated interrupt binding when a raw-mode TUI consumes Ctrl+C.
+psmux send-keys -f C-c
+
 # Special keys supported:
 # Enter, Tab, Escape, Space, Backspace
 # Up, Down, Left, Right, Home, End
 # PageUp, PageDown, Delete, Insert
 # F1-F12, C-a through C-z (Ctrl+key)
 ```
+
+`send-keys -f C-c` (or `--force-signal`) is a psmux extension for Windows. It
+bypasses the raw-mode TUI heuristic and sends a CTRL_C_EVENT to the foreground
+process. This keeps normal keyboard Ctrl+C pass-through behavior unchanged while
+allowing an explicit binding such as `bind -n C-F12 send-keys -f C-c` to force an
+interrupt.
 
 ## Pane Information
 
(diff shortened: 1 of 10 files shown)


psmux:
Thank you for this, and apologies for the slow verdict. The problem you were solving is real, and your writeup described it well: raw mode TUIs can eat Ctrl+C, and users need a guaranteed interrupt they can bind.

I am closing this one without merging because the ground shifted under it. Since this branch was cut, #454 landed (commit b55a9ec), which reworked the interrupt path to deliver a genuine `CTRL_BREAK_EVENT` to the ConPTY child, precisely so that Ctrl+C immune programs actually stop, matching what Windows Terminal does. That work touched the same plumbing this PR modifies (`platform.rs`, the send-keys signal path), which is why the branch now conflicts, and it covers the core need this PR was built for: a reliable, explicit way to interrupt a stubborn foreground process.

If after trying current master you still see a gap that a bindable `send-keys -f C-c` would close, for example wanting CTRL_C semantics specifically rather than CTRL_BREAK, I would genuinely welcome a fresh PR on top of the new signal path, and the flag design you have here (`-f` with a `--force-signal` long form, preserved through CLI forwarding) is the right shape for it. Thank you for the clean flag plumbing and the regression tests.

PR #375, opened 2026-06-10
fix: reap session registry files orphaned by a reboot or port reuse

tarikguney:
## What you see

After Windows reboots for an OS update, you reopen psmux and the session picker is full of old sessions stuck in `(not responding)`. They should be gone, because the psmux servers that ran those sessions died with the reboot. But their entries linger and you cannot get rid of them.

image

## Why every dead session still shows up

Two things combine here. Keep them separate, because they are different problems:

**A. The picker only labels, it never prunes.** The session picker builds its list straight from the leftover `<name>.port` files on disk. For each one it tries a single AUTH + `session-info` fetch, and if that fetch comes back with no usable info for *any* reason, it just prints `<name>: (not responding)`. It never deletes the file. So every dead session's leftover files show up in the list, no matter why the fetch failed.

**B. Startup cleanup was too weak to remove them first.** psmux does have a startup cleanup, but the old check was naive: it did a bare TCP connect to the port, and if anything answered it called the session "alive" and kept the files. Anything that was not a clean connection-refused was treated as "inconclusive" and also kept. So leftovers survived cleanup and reached the picker.

### What is actually on each old port after a reboot

Every session writes three files under `~/.psmux/`:

- `<name>.port` — the TCP port its server listens on
- `<name>.key` — a per-session secret for authentication
- `<name>.sid` — the session id

psmux servers do not use fixed ports; each asks the OS for a free ephemeral port at startup. After a reboot those files remain on disk but the servers are gone, and the OS starts handing out ephemeral ports again from the bottom of the range. So for each old `.port` the port number now usually points at one of:

1. **Nothing (free).** Most of them. A connect is refused.
2. **An unrelated program.** Some other app happens to listen on that number. A bare connect succeeds, so the old check wrongly called it "alive".
3. **A different, brand new psmux server.** Possible, but rare right after a reboot: psmux has only spawned a server or two by then (the warm standby plus whatever you just opened), so at most those one or two ports can land on an old session's number. The connect succeeds and it even speaks the psmux protocol, so the old check definitely called it "alive".

Cases 2 and 3 are why a bare "did the port answer?" test is wrong: it cannot tell our server from whoever else is on that port now. But none of these three is, by itself, why the whole list appears. That is problem A above.

## The "(not responding)" misnomer

The label is misleading: it suggests silence, but the port often did answer. Take case 3, the cleanest example. The picker connects and sends:

```
AUTH <old-key>
session-info
```

It expects back the server's `OK` followed by a session-info line like `mysession: 2 windows (created ...)`. But the server now on that reused port has a different key, so it replies:

```
ERROR: Invalid session key
```

The picker treats any `ERROR` as "no info" and prints `mysession: (not responding)`. So the port responded, it just rejected us.

### Detailed reason

`(not responding)` is just the fallback string the picker substitutes whenever the fetch returns "no info". So the real question is: what did the picker need to hear, and what did it actually get?

**What it sends** (per session): `AUTH <key>\n` then `session-info\n`.

**What it needs back:** the server's `OK` ack, followed by a real session-info payload line. That payload line is what would have been shown:

```
OK
mysession: 2 windows (created Mon Apr 20 11:10:58 2026)
```

**What counts as "no info":** the response reader returns nothing for any of these:

- empty line, EOF, or read timeout (true silence — e.g. an unrelated program that never speaks our protocol, case 2)
- the line is just `OK` with no payload after it
- the line starts with `ERROR:` (an explicit auth rejection — case 3)

So different leftover sessions hit "no info" for different reasons: free ports refuse the connect, unrelated programs stay silent, and a reused psmux port replies `ERROR`. The pre-fix picker collapsed all of them to the same `(not responding)` label and left the files in place. The `ERROR` path is the one worth calling out because it proves the point: that case is not silence, it is a definite rejection, which is hard proof the session is dead (a different server holds the port).

## The fix

Stop trusting "did the port answer?", require proof of identity (the AUTH key), and make the picker prune instead of just label.

1. **Identity-aware, bounded liveness probe.** For each session the picker does exactly one bounded probe: connect, AUTH with that session's own key, ask for `session-info`. One attempt, fixed timeouts (~50ms connect + ~250ms read), then a decision. It never retries forever.
   - `OK` + a real info line means it is our server -> **Alive**, shown with its info.
   - `ERROR` (a different server reused the port), connection failed (free port — on loopback any connect failure means nothing is listening), or connected-then-silent past the read timeout (unrelated program) -> **Dead**, files reaped, not shown.
   - No usable AUTH key on disk to verify with -> **Unreachable**, kept and shown honestly as `(not responding)` (ambiguous, so not deleted).

2. **Safe to reap on timeout.** Deleting a session's files is non-destructive even if we are briefly wrong: a live server rewrites its `.port`/`.key`/`.sid` every 5 seconds (`ensure_session_registry_files`). So a momentarily-slow live server self-heals; a truly dead one stays gone.

3. **Boot-time guard.** Any registry file last written before the machine booted cannot belong to a server running since startup, so startup cleanup deletes it unconditionally. No network, immune to port reuse. This clears reboot zombies the instant you run any psmux command.

4. **The picker lists and prunes in one parallel pass.** The probe above runs on one thread per session, so opening the picker is bounded by ~one probe window regardless of how many sessions exist. There is no separate sequential cleanup pass, so the picker stays responsive. Dead sessions simply do not appear; live ones show real info.

## The path, end to end

```
open picker
  -> classify_sessions_parallel()       (one bounded probe per session, in parallel)
       for each <name>.port:
         connect failed (free port)  --------> Dead        -> reap, omit
         AUTH -> OK + info           --------> Alive(info)  -> show real info
         AUTH -> ERROR (reused)      --------> Dead        -> reap, omit
         connected, silent > timeout --------> Dead        -> reap, omit
         no usable key on disk       --------> Unreachable -> keep, show "(not responding)"
  -> list shows live sessions; dead ones are gone, not zombies

startup (any psmux command)
  -> cleanup_stale_port_files()
       file older than last boot?    --------> delete .port/.key/.sid  (boot guard, no network)
       otherwise identity-probe the port (conservative belt-and-suspenders)
```

A live server that gets reaped by mistake reappears within 5s via its registry self-heal tick, so reaping is always safe.

## Observability

Added an env-gated logger, off by default, matching the existing `PSMUX_*_DEBUG` facility:

```
PSMUX_SESSION_DEBUG=1  ->  ~/.psmux/session_debug.log
```

It records each decision, for example:

```
[12:01:03.221][probe] port 50123: AUTH rejected by a different server (reused port) -> stale
[12:01:03.224][cleanup] reaping 'mysession': port file predates last boot (server died on restart)
[12:01:03.230][picker] reaping dead session 'oldsess' from chooser
```

So next time this happens you can see exactly which sessions were reaped and why, versus which were kept.

## Tests

- picker probe: authenticated server -> Alive; auth rejection -> Dead; connection refused -> Dead; connected-but-silent -> Dead (and the probe stays bounded, does not hang)
- startup cleanup: reused-port (key rejection) is reaped; authenticated server is preserved; pre-boot files are reaped while just-after-boot and post-boot files are kept
- existing stale/inconclusive cleanup behavior unchanged

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -2681,15 +2681,11 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 popup_rect_last = None;
                                 if choose_tree_preview_default { preview_enabled = true; }
                                 let dir = format!("{}\\.psmux", home);
-                                // Reap registry files for sessions whose server
-                                // is gone (crashed, or killed by a reboot) before
-                                // listing — otherwise their port files linger and
-                                // surface here as "(not responding)" zombies.
-                                crate::session::cleanup_stale_port_files();
-                                // Collect (label, addr, key) for every reachable port file first,
-                                // then fan out the per-session AUTH+session-info fetches in parallel.
-                                // Sequential fetches made the picker open in O(N * read_timeout);
-                                // parallelism keeps it bounded by the single-fetch timeout.
+                                // Collect (label, addr, key) for every port file, then run ONE
+                                // bounded liveness probe per session in parallel. This both lists
+                                // and prunes: total wall time is ~one probe window regardless of
+                                // how many sessions exist, so the picker stays responsive (no
+                                // sequential O(N * timeout) cleanup pass).
                                 let mut targets: Vec<(String, String, String)> = Vec::new();
                                 if let Ok(entries) = std::fs::read_dir(&dir) {
                                     for e in entries.flatten() {
@@ -2709,13 +2705,35 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                         }
                                     }
                                 }
-                                let fetched = crate::session::fetch_session_infos_parallel(
+                                let verdicts = crate::session::classify_sessions_parallel(
                                     targets,
-                                    Duration::from_millis(25),
-                                    Duration::from_millis(150),
-                                    |label| format!("{}: (not responding)", label),
+                                    Duration::from_millis(50),
+                                    Duration::from_millis(250),
                                 );
-                                session_entries.extend(fetched);
+                                for (label, liveness) in verdicts {
+                                    match liveness {
+                                        crate::session::SessionLiveness::Alive(info) => {
+                                            session_entries.push((label, info));
+                                        }
+                                        crate::session::SessionLiveness::Dead => {
+                                            // Server gone (crashed, killed by a reboot, or its
+                                            // port reused by another server). Reap it so it never
+                                            // shows as a "(not responding)" zombie. A live-but-slow
+                                            // server self-heals on its next 5s registry tick.
+                                            if crate::debug_log::session_log_enabled() {
+                                                crate::debug_log::session_log("picker",
+                                                    &format!("reaping dead session '{}' from chooser", label));
+                                            }
+                                            crate::session::remove_session_registry(&label);
+                                        }
+                                        crate::session::SessionLiveness::Unreachable => {
+                                            // Could not connect (transient, non-refused). Keep it
+                                            // and show the honest "(not responding)" rather than
+                                            // deleting on an ambiguous failure.
+                                            session_entries.push((label.clone(), format!("{}: (not responding)", label)));
+                                        }
+                                    }
+                                }
                                 if session_entries.is_empty() {
                                     session_entries.push((current_session.clone(), format!("{}: (current)", current_session)));
                                 }
--- a/src/client.rs
+++ b/src/client.rs
@@ -2681,6 +2681,11 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 popup_rect_last = None;
                                 if choose_tree_preview_default { preview_enabled = true; }
                                 let dir = format!("{}\\.psmux", home);
+                                // Reap registry files for sessions whose server
+                                // is gone (crashed, or killed by a reboot) before
+                                // listing — otherwise their port files linger and
+                                // surface here as "(not responding)" zombies.
+                                crate::session::cleanup_stale_port_files();
                                 // Collect (label, addr, key) for every reachable port file first,
                                 // then fan out the per-session AUTH+session-info fetches in parallel.
                                 // Sequential fetches made the picker open in O(N * read_timeout);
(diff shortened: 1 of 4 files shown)


TrvoDK:
Thanks for #375 — the parallel `classify_sessions_parallel` pass is a big improvement, and I can confirm it fixed the **session chooser** (`Ctrl-b s`): on v3.3.6, opening it against a registry full of reboot-orphaned `<name>.port` files now classifies + **prunes** them in one bounded pass, and the dead entries are gone.

While verifying that, I noticed the **`choose-tree` picker (`Ctrl-b w`) didn't get the same treatment** and still has the pre-#375 shape. In `client.rs`, the `do_choose_session` branch now calls `classify_sessions_parallel(...)` and reaps `Dead` verdicts via `remove_session_registry(...)`. The `do_choose_tree` branch still iterates the `.port` files **sequentially**, doing a per-session `fetch_authed_response_multi(addr, key, b"list-tree\n", 50ms, 100ms)` with no classification and no prune — so it's the same "the picker only labels, it never prunes" problem #375's write-up describes, just in the picker that didn't get migrated:

| opening `Ctrl-b w` (choose-tree) against… | behaviour on v3.3.6 |
|---|---|
| only live sessions | opens fine; cheap (isolated cold-open ~85 ms at 1 session → ~129 ms at 30) |
| orphaned `.port` files present | dead entries are **skipped but never pruned**, and re-probed sequentially on **every** open — so an accumulated / post-reboot registry never self-heals through this picker |
| after one `list-sessions` / `Ctrl-b s` (which run the classify pass) | the orphans are pruned, and `Ctrl-b w` is back to the live-only cost |

**Would routing `do_choose_tree` through `classify_sessions_parallel` too be welcome?** It already does the bounded classify + prune that the tree loop lacks. (The classify pass returns `session-info`, not the window/pane tree, so `Ctrl-b w` would still need its own `list-tree` fetch for the data — but the dead-session pruning + bounded probe would be shared, and the two pickers would stop diverging.)

**A separate issue I then isolated — the attach client crashes on a stale entry, no picker needed.** I'd seen an `os error 10061` near the picker and assumed it was choose-tree; it isn't. With a **single** stale/unreachable entry in the registry (one leftover `.port` whose server is gone — a `(not responding)` zombie), the **`attach` client exits ~2–5 s after attaching, while idle, with no keypress**, printing `psmux: No connection could be made because the target machine actively refused it. (os error 10061)`. It's **deterministic** (every run with a stale entry present; never without one — controlled both ways) and **client-only**: the session server it's attached to keeps running. So some periodic client-side task connects across the registry and propagates the connect error on a stale port instead of skipping it — the same `.port` files this PR is about. I haven't pinned the exact call site. One honest caveat: I reproduce it through a headless ConPTY `attach` (I can't drive a native terminal in CI), but the failing path looks like a plain psmux-internal connect, not anything terminal-specific. Happy to share the self-contained repro.

The harness that surfaced both of these is a small isolated, non-destructive ConPTY rig that drives the real `attach`/picker against a **sandboxed `$USERPROFILE`** server (it never touches the real one). I'm keeping it as an external repo and using it to file repros like these; glad to share self-contained repros for either point here.

PR #383, opened 2026-06-14
(fix: swap-pane:) actually swap panes; add -t targets, position tokens, and post-swap resize

jolchmo:
# swap-pane: actually swap panes; add -t targets, position tokens, and post-swap resize

## Summary
`swap-pane` only moved the active-pane marker instead of exchanging the two panes
in the window layout, so it was effectively a no-op. This PR makes `swap-pane`
actually swap the panes (matching tmux), adds the `-t` targeting tmux supports
(pane id/index and `{position}` tokens), and resizes the swapped panes to their
new cells.

## The bug
`window_ops::swap_pane()` located the neighbour pane in the requested direction
and then only set `win.active_path` to it — it never exchanged the two panes in
the layout tree. Consequences:
- `swap-pane -U`/`-D` (and `swapp`) did nothing visible — they just moved focus.
- `rotate-window` was a no-op too, and same-window `join-pane`/`move-pane` are
  rejected, so there was no working way to reorder panes.

In tmux, `swap-pane` swaps the two panes' positions in the window
(man page: *"swap-pane … Swap two panes."*).

### Real swap
- `tree::swap_nodes(root, a, b)` exchanges the two subtrees at two distinct leaf
  paths via `std::ptr::swap` (because `Pane` is not `Clone`), keeping split kinds
  and `sizes`. Guards against equal paths and ancestor/descendant paths.
- `window_ops::swap_pane` and the new `swap_pane_with_path` call `swap_nodes`
  and move focus to follow the swapped-in pane.

### Resize after swap (tmux parity)
- After a successful swap, call `resize_all_panes()` so each pane's PTY and
  `vt100` size is updated to its new cell. tmux keeps the layout cells fixed and
  resizes the swapped panes into them; without this the program inside the moved
  pane keeps its old terminal size.

### `-t` target support
- `swap-pane -t <target>` now works for `session:window.pane`, pane index, and
  `%id` (via `parse_target` → `find_path_by_id` / `path_by_position`), routed
  through a new `CtrlReq::SwapPaneTarget`.
- Added tmux's pane **position tokens** as targets: `{top-left}`, `{top-right}`,
  `{bottom-left}`, `{bottom-right}`, `{top}`, `{bottom}`, `{left}`, `{right}`
  (`window_ops::resolve_position_token`, `CtrlReq::SwapPanePosition`). Each
  resolves to whatever pane currently occupies that corner/edge, so e.g.
  `bind -n M-w swap-pane -t '{top-right}'` is layout-independent. tmux's
  `target-pane` grammar already defines these tokens.
- Added `swap-pane -L`/`-R` alongside the existing `-U`/`-D`.
- Excluded `swap-pane` from the control path's "focus the `-t` target before
  running the command" step — otherwise the active pane became the target and
  the swap was a self-swap (no-op).

## Test plan
- Unit tests added: `tree::swap_node_tests` (non-sibling / sibling swap,
  invalid-path rejection) and `window_ops::position_token_tests` (corner/edge
  resolution).
- Manual (Windows 11, PowerShell): in a 3-pane layout, `swap-pane -t <big-pane>`
  exchanges the focused pane with the target, focus follows, and the swapped-in
  pane is resized to the target cell; `swap-pane -t '{top-right}'` hits the
  top-right pane regardless of pane order.

## Notes / scope
- Relative tokens `{up-of}` / `{down-of}` / `{left-of}` / `{right-of}` are not
  added here (only the absolute corner/edge tokens).
- A one-shot CLI on a custom `-L` socket can mis-route a `{...}` target as a
  session name; the key-binding / in-server path resolves it correctly (the
  common use). Can be addressed in a follow-up if desired.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1253,11 +1253,43 @@ fn execute_command_string_single(app: &mut AppState, cmd: &str) -> io::Result<()
             }
         }
         "swap-pane" | "swapp" => {
-            if let Some(port) = app.control_port {
-                let dir = if parts.iter().any(|p| *p == "-U") { "-U" } else { "-D" };
+            // `-t <target>` swaps the active pane with an explicit target pane
+            // (e.g. `swap-pane -t :.4` or `swap-pane -t %2`).  Without -t, fall
+            // back to the directional -U/-D/-L/-R swap.
+            let target = parts.iter().position(|p| *p == "-t")
+                .and_then(|i| parts.get(i + 1)).map(|s| s.to_string());
+            if let Some(tgt) = target {
+                if let Some(port) = app.control_port {
+                    let _ = send_control_to_port(port, &format!("swap-pane -t {}\n", tgt), &app.session_key);
+                } else {
+                    let path = if tgt.starts_with('{') {
+                        // Layout position token like {top-right} — resolve to
+                        // whatever pane currently occupies that corner.
+                        crate::window_ops::pane_path_at_position(app, &tgt)
+                    } else {
+                        let parsed = crate::cli::parse_target(&tgt);
+                        let win = &app.windows[app.active_idx];
+                        match parsed.pane {
+                            Some(p) if parsed.pane_is_id => crate::tree::find_path_by_id(&win.root, p),
+                            Some(p) => crate::tree::path_by_position(&win.root, p.saturating_sub(app.pane_base_index)),
+                            None => None,
+                        }
+                    };
+                    if let Some(path) = path {
+                        crate::window_ops::swap_pane_with_path(app, path);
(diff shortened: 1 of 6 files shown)


tarikguney:
Thanks for the work on this — the main fix is solid and definitely needed.

I pushed a small follow-up on top to make it safer to merge:
- invalid `-t` pane indexes no longer fall through to pane 0,
- MRU now tracks the pane that stays focused after swap,
- local errors now show in the psmux status line (instead of stderr),
- added a regression test for the pane-base-index edge case.

Kept everything in this same PR and only touched swap correctness + user-facing error behavior.

PR #385, opened 2026-06-14
fix(conpty): detect ESC[6n split across parser read batches

marhel:
I've observed additional robustness problems that this PR tries to fix. 

What we observed was that the initial command sometimes seems to hang somewhere during startup. After some debugging, process dumping and code reading it seems that one plausible cause matching what we see is the way that psmux scans for CPR query (Cursor Position Request) escape codes (ESC[6n) using scan_cpr_query. This is done in "batches" where the reader thread drains the ConPTY output pipe collecting some bytes, and (among other things) scans for a CPR to respond to.

However, a query that straddles a batch boundary (…ESC[6 ends one batch, n… begins the next) is in neither batch whole, so it is never detected: cpr_pending is never set, the reply is never sent, and whatever issued the request waits for it forever.

Now, why would that cause a startup hang? In short it is because psmux creates the psuedoconsole with the INHERIT_CURSOR flag, which according to [Microsoft's documentation for CreatePseudoConsole](https://learn.microsoft.com/en-us/windows/console/createpseudoconsole) requires the calling application to provide a cursor position response and "failure to [respond] may cause the calling application to hang while making another request of the pseudoconsole system", which matches what we see; when the pane command tries to open the console, it hangs.

Making the CPR query scanning detect escape codes even when split over two batches makes the hangs we observe go away. It will also respond correctly to a CPR query later in a process lifetime whenever this situation previously would cause psmux to ignore the request.

### DIFF
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -1199,6 +1199,54 @@ fn scan_cpr_query(data: &[u8]) -> bool {
     data.contains(&0x1b) && data.windows(CPR.len()).any(|w| w == CPR)
 }
 
+/// Detects ESC[6n across batch boundaries. The parser thread scans output in
+/// coalesced batches; a cursor-position request split across two batches is
+/// invisible to the per-batch `scan_cpr_query` (no carry-over), so `cpr_pending`
+/// is never set, the reply is never written, and whatever issued the request
+/// waits for it forever. The asker can be pwsh after a lock/unlock, or conhost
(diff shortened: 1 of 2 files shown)


tarikguney:
Thanks for digging into this and putting together a focused fix.

Your write-up of the split-batch ESC[6n case makes sense, and the CprScanner change addresses a real blind spot in the current per-batch detection. This aligns with the intermittent startup/lock-unlock hangs we’ve been seeing, even if we should still treat it as one important failure path rather than the only one.

Appreciate the clear tests and targeted scope.

PR #387, opened 2026-06-14
fix: detach-client without -P cleanly detaches instead of reconnecting (#386)

### LINKED ISSUE #386: detach-client (without -P) does not detach: client reconnects instead of exiting
## Summary

`psmux detach-client` (without `-P`) returns exit code 0 but does **not** actually detach an attached client — the client immediately **reconnects**, so the detach has no effect. Only `detach-client -P` actually detaches, but `-P` also kills the client's parent process. This means there is no way to cleanly detach a client from the CLI without killing its parent, which breaks scriptable/programmatic detach (e.g. a session-switcher that wants to detach a client to hand control back to its launcher).

Real tmux's `detach-client` cleanly detaches: the client process exits and the session stays alive, no reconnect.

## Environment

- psmux 3.3.6 (Windows, WinGet `marlocarlo.psmux`)
- Reproduced on an isolated `-L` socket (no real sessions involved)

## Reproduction

```
psmux -L pf new-session -d -s s
# in a second terminal:
psmux -L pf attach -t s          # psmux -L pf list-clients -> 1 client (e.g. /dev/pts/5)
# back in the first terminal:
psmux -L pf detach-client -s s   # exit 0
psmux -L pf list-clients         # STILL 1 client (tty changed, e.g. /dev/pts/10)
```

The `psmux attach` **process is still alive** and the session is still `(attached)` — the client reconnected instead of exiting.

### Expected

After `detach-client -s s`, `list-clients` is empty, the `psmux attach` process exits, and the session persists detached (this is what `tmux detach-client -s s` does).

### Actual

The client reconnects (new tty, same process), session stays attached. `detach-client -P` is the only variant that makes the client exit — but it also terminates the parent process.

## Root cause

The non-`-P` detach handlers in `src/server/mod.rs` force-close the client's TCP stream **without first sending a clean `"DETACH"` directive**:

- `CtrlReq::ForceDetachClient` — calls `shutdown_client_stream(target_cid)` with no directive.
- `CtrlReq::ForceDetachClientByTty`, `CtrlReq::DetachAllOtherClients`, `CtrlReq::DetachAllClients` — only send a directive in the `kill_parent` (`-P`) branch (`"DETACH-KILL-PARENT"`); the non-`-P` path just calls `shutdown_client_stream`.

The plain `"DETACH"` directive is only sent on session-end / kill paths (`send_directive_to_all_clients("DETACH")`). On the client side (`src/client.rs`), a client sets `quit=true` (clean exit) **only** when it receives a `"DETACH"` / `"DETACH-KILL-PARENT"` directive; on a bare stream drop, if `!quit && port_file_exists`, it spawns a reconnect thread. So a non-`-P` detach (stream shutdown with no directive) is seen as a transient disconnect and the client reconnects.

## Suggested fix

In the non-`-P` detach paths, send a clean `"DETACH"` directive to each targeted client before `shutdown_client_stream` (mirroring the `-P` path but without kill-parent), so the client sets `quit=true` and exits instead of reconnecting. PR follows.

zer0ken:
## What

`detach-client` (without `-P`) now actually detaches the targeted client(s) instead of letting them reconnect.

Fixes #386.

## Why

The non-`-P` detach handlers in `src/server/mod.rs` (`ForceDetachClient`, `ForceDetachClientByTty`, `DetachAllOtherClients`, `DetachAllClients`) force-closed the client's TCP stream via `shutdown_client_stream` **without first sending a clean `"DETACH"` directive**. Only the `-P` branch sent a directive (`"DETACH-KILL-PARENT"`).

On the client side (`src/client.rs`), a client only sets `quit=true` (clean exit) when it receives a `"DETACH"`/`"DETACH-KILL-PARENT"` directive; on a bare stream drop it treats the disconnect as transient and spawns a reconnect thread (since the port file still exists). So a non-`-P` detach made the client **reconnect** (new tty, same process) — `detach-client` returned exit 0 but had no effect. The only working CLI detach was `-P`, which also kills the parent process.

This diverges from tmux, whose `detach-client` cleanly exits the attached client while the session persists.

## How

Send a `"DETACH"` directive to each targeted client before `shutdown_client_stream` in the non-`-P` paths, mirroring the existing `-P` flow but without kill-parent. The client then sets `quit=true` and exits cleanly. `-P` behavior is unchanged.

## Verification

- `cargo check` passes.
- Manual repro (psmux 3.3.6, isolated `-L` socket): before the fix, `detach-client -s <session>` left the client attached with a changed tty (reconnect); the same flow on tmux 3.2a cleanly detaches. With this change the psmux client exits and the session persists detached, matching tmux.

Invariants preserved: `prefix+d` / `prefix+:detach-client` still detach the current client; `detach-client -a` from inside a client still does not exit the calling client; the session/panes survive a detach.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -3699,6 +3699,11 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                     if app.latest_client_id == Some(target_cid) {
                         app.latest_client_id = app.client_registry.keys().max().copied();
                     }
+                    // Tell the client to detach cleanly so it exits instead of
+                    // reconnecting. Without a DETACH directive the client treats the
+                    // dropped stream as a transient disconnect and reconnects.
+                    crate::types::send_directive_to_client(target_cid, "DETACH");
+                    std::thread::sleep(Duration::from_millis(50));
                     // Shut down the TCP stream to force disconnect
                     crate::types::shutdown_client_stream(target_cid);
                     // Recompute effective size from remaining clients
@@ -3733,10 +3738,14 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                         .find(|(_, ci)| ci.tty_name == tty)
                         .map(|(cid, _)| *cid);
                     if let Some(cid) = target_cid {
+                        // Send a clean detach directive first so the client exits
+                        // instead of reconnecting on a bare stream drop.
                         if kill_parent {
                             crate::types::send_directive_to_client(cid, "DETACH-KILL-PARENT");
-                            std::thread::sleep(Duration::from_millis(50));
+                        } else {
+                            crate::types::send_directive_to_client(cid, "DETACH");
                         }
+                        std::thread::sleep(Duration::from_millis(50));
                         app.client_sizes.remove(&cid);
                         let was_present = app.client_registry.remove(&cid).is_some();
                         if was_present {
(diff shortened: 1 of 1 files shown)


zer0ken:
Closing — this PR was opened prematurely on my side and not intended for submission yet. Apologies for the noise.

PR #393, opened 2026-06-14
Fix: attach -t SESSION attaches to most-recent session instead of target

andrew-tawfeek:
## Summary

`psmux attach -t SESSION1` attaches to the **most-recently-used** session (e.g. `SESSION2`) instead of `SESSION1`. The positional form `psmux attach SESSION1` works correctly — only the `-t` form is broken.

## Root cause

In `main()`, the argv pre-filter that builds `cmd_args` strips `-t <value>` **after the subcommand for every command**, on the assumption that downstream code reads the already-resolved `PSMUX_TARGET_SESSION` env var:

```rust
} else {
    // After subcommand: strip only -t (and its value)
    if args[i] == "-t" && i + 1 < args.len() { i += 2; continue; }
}
```

But the `attach`/`attach-session` handler does **not** read that env var — it re-derives its target by scanning the (already stripped) arg list for `-t`. The lookup therefore never matches, and resolution falls through the `.or_else` chain to `resolve_last_session_name_ns()`, i.e. the most-recent session.

The positional form is unaffected because positional args are not stripped from `cmd_args` — which is the tell that only the `-t` path was broken.

## Fix

Read `-t` from the **raw `args`** (the same source the global `-t` handler at the top of `main()` uses), extracted into a small pure helper `cli::attach_target_session()` that also applies the `-L` socket-namespace prefix. The positional / default / last-session fallbacks are unchanged.

## Tests

Adds `tests-rs/test_attach_target_session.rs` (wired into `cli.rs`'s test module), covering:

- `-t NAME` returns the requested session (the core regression)
- all aliases (`a`, `at`, `attach`, `attach-session`)
- `-t=NAME` (equals form, post-`normalize_flag_equals`)
- no `-t` → `None` (falls through to positional/default/last)
- `-L` socket-namespace prefixing
- full `session:window.pane` target spec
- trailing `-t` with no value → `None` (no panic)

```
running 7 tests
test cli::tests_attach_target_session::... ok (x7)
test result: ok. 7 passed; 0 failed
```

Built and tested with the MSVC toolchain on Windows 11 (`cargo build --release`, `cargo test --bin psmux`).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -629,6 +629,29 @@ pub fn extract_session_from_target(target: &str) -> String {
     parsed.session.unwrap_or_else(|| "default".to_string())
 }
 
+/// Resolve the explicit `-t <session>` target for attach-session from the full
+/// argv, applying the `-L` socket-namespace prefix when present.
+///
+/// This MUST read from the raw `args`, not from the post-subcommand filtered
+/// command args: the arg pre-filter in `main` strips `-t <value>` for every
+/// subcommand (it assumes downstream reads the resolved env var). The attach
+/// handler re-derives the target from argv, so if it scans the *stripped* args
+/// it never finds `-t`, falls through to "most recent session", and attaches to
+/// the wrong session — `psmux attach -t A` opening B. Reading raw argv here (the
+/// same way the global `-t` handler does) keeps the explicit target intact.
+///
+/// Returns `None` when no `-t` flag is present (the caller then tries a
+/// positional name, then the default/last-session fallbacks).
+pub fn attach_target_session(args: &[String], l_socket_name: Option<&str>) -> Option<String> {
+    args.iter()
+        .position(|a| a == "-t")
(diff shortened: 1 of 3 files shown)


andrew-tawfeek:
Obviously used Claude to assist with PR -- using `-t` is the standard form for `tmux` which is how I walked into this bug -- I was able to fix it locally and rebuild it, soo pushed a PR to share and help close out this issue with others

radiaku:
this happened to me, and thank for the pull, I rebuild my onw.

psmux:
Fixed on master in #438. Same root cause you described here: attach was resolving to the most recent session instead of the -t target, because the global arg scan stripped -t before the attach handler read it, dropping through to the last_session marker. The fix reads -t from the full argv so the target wins, with bare attach and positional attach behaviour preserved. Landed with a regression test and green Windows CI. Thanks @andrew-tawfeek for calling this out early. Closing as fixed; it will be in the next build.

PR #395, opened 2026-06-15
test(ps1): non-destructive test-env helper, adopt in new-session tests

marhel:
This PR adds a shared, non-destructive test helper and adopts it in the three new-session regression tests, as a first step toward making the PowerShell suite safe to run on a machine that also runs psmux for real (like I do).

Today many tests assume a throwaway sandbox and can disrupt a real session on the same machine: they kill psmux by image name (`Stop-Process`/`taskkill`), run un-namespaced `kill-server` (the kill-all-by-image-name fallback), write or remove the real `~/.psmux.conf`, and send commands to the default (non-namespaced) session. The helper runs a test against a fully isolated, throwaway environment instead, so none of that reaches the real install.

**`tests/psmux_test_helpers.ps1`:**
- `New-PsmuxTestEnv` — redirects `USERPROFILE`/`HOME` to a throwaway dir (data dir, config, sockets, warm pool all land there; the real `~/.psmux` is untouched) and scrubs the session vars a live psmux leaks into a pane shell (`PSMUX_SESSION`, `PSMUX_TARGET_SESSION`, `TMUX`, `TMUX_PANE`).
- `Remove-PsmuxTestEnv` — namespace-scoped server shutdown (never a bare `kill-server`), force-nets only orphans of the binary under test, restores the env, deletes the temp dir.
- `Get-PsmuxExe`, `Invoke-PsmuxListWindows` — lift duplicated exe-resolution and the control-mode `list-windows` probe.

The three tests shrink ~70 lines net. Verified behavior-preserving (A-B against the same debug build); all three GREEN. `stress_new_session_readiness.ps1` could have been included in this first set, but is left out since it''s being modified by #388.

**Proposed direction (follow-up PRs):** migrate the rest of the ps1 tests onto the helper in small batches (config tests, image-name killers, the TCP/control-mode family, reporting). As we go, lift the patterns each batch has in common — spinning up an isolated session, running a command, probing via control mode, asserting, tearing down — into the helper, so it grows into a shared vocabulary for the suite and replaces the copy-pasted helpers (e.g. `Write-Pass`/`Write-Fail`, inline in 317 files each). As an initial migration step, we could gate each not-yet-secured destructive test on `PSMUX_TEST_SANDBOX` (the var `run_all_tests.ps1` already requires) with an early exit, removed per-test once it''s been made safe to run outside a sandbox.

### DIFF
--- a/tests/psmux_test_helpers.ps1
+++ b/tests/psmux_test_helpers.ps1
@@ -1,0 +1,161 @@
+# Shared, non-destructive helpers for the psmux PowerShell test suite.
+#
+# Dot-source:  . "$PSScriptRoot\psmux_test_helpers.ps1"
+#
+# Isolation model — safe to run on a machine that also runs psmux for real:
+#   * New-PsmuxTestEnv points USERPROFILE *and* HOME at a throwaway temp dir, so
+#     the data dir (~/.psmux: sockets, .port/.key files, warm pool), config, and
+#     everything home-relative live under that dir. The real ~/.psmux is never
+#     touched, and the prior USERPROFILE/HOME are saved for restore.
+#   * It also scrubs the inherited session/target vars (PSMUX_SESSION,
+#     PSMUX_ACTIVE, ...). Run from inside a psmux session those would otherwise
+#     trip the nesting guard, so `new-session` never spawns a real server and
+#     the test silently no-ops. Saved and restored alongside USERPROFILE/HOME.
+#   * Remove-PsmuxTestEnv shuts servers down namespace-scoped (`-L <ns>
+#     kill-server`, NEVER a bare kill-server, which has a nuclear kill-all-by-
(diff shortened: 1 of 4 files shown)


psmux:
Merged. Martin, thank you for this, and for the direction it sets.

I read the helper closely before merging because teardown code is exactly where a test suite can hurt a real install, and this is the opposite of that: namespace scoped shutdown only, the force net matched strictly on the binary under test path plus registered namespaces, env saved and restored including the leaked session vars, and the real `~/.psmux` never touched. The documented reasoning about which vars to scrub and which to leave alone is the kind of comment future contributors will thank you for.

I run psmux for real on the machine that runs these tests, so the isolation model matters to me personally. The three migrated tests came out cleaner and shorter too. I am in favor of the follow-up batches you proposed; migrating the image name killers next would retire the most dangerous patterns first.

PR #398, opened 2026-06-16
fix: resolve #{session_path} to the server working directory

marhel:
This PR fixes `#{session_path}` to resolve to the session working directory (which is the server working directory as of #369)

- It returned the user home directory. tmux defines `#{session_path}` as the
  working directory of the session.
- Now resolves to the server's current directory (`std::env::current_dir()`),
  the same source `#{pane_current_path}` already falls back to. Fixed in both
  `expand_var` arms — the windowed match and the no-window early-return path.
- TDD: a `test_format` test asserts it expands to the server cwd, not home.

**Observable change:** `#{session_path}` now reports the session (server)
working directory instead of the home directory.

## Repro

```shell
$ tmux new -d -c C:\WINDOWS
# should return C:\WINDOWS + C:\WINDOWS but returns the value of $USERPROFILE (or $HOME) + C:\WINDOWS
$ tmux ls -F "#{session_path} + #{pane_current_path}" 
C:\Users\your.name + C:\WINDOWS
```

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1046,6 +1046,9 @@ pub fn expand_var(var: &str, app: &AppState, win_idx: usize) -> String {
                 "session_name" => app.session_name.clone(),
                 "session_windows" => app.windows.len().to_string(),
                 "session_id" => format!("${}", app.session_id),
+                "session_path" => std::env::current_dir()
+                    .map(|d| d.to_string_lossy().into_owned())
+                    .unwrap_or_default(),
                 "pid" | "server_pid" => std::process::id().to_string(),
                 "version" => VERSION.to_string(),
                 "host" | "hostname" => hostname_cached(),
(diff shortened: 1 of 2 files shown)


tarikguney:
Confirmed this is a real bug on current master, and reproduced it live:

```
$ psmux ls -F "#{session_path} + #{pane_current_path}"
C:\Users\abguney + C:\Windows\System32
C:\Users\abguney + Q:\sources\cc-http2
C:\Users\abguney + Q:\sources\psmux\psmux
...
```

`#{session_path}` returns `C:\Users\abguney` (my `USERPROFILE`) for every session, ignoring the actual session working directory that `#{pane_current_path}` reports correctly. The culprit is `src/format.rs:1098`:

```rust
"session_path" => env::var("HOME").or_else(|_| env::var("USERPROFILE")).unwrap_or_default(),
```

tmux defines `#{session_path}` as the session's working directory, and as of #369 that''s the server cwd — the same `std::env::current_dir()` source `#{pane_current_path}` already falls back to. Routing `session_path` to that source is the correct, consistent fix. Merging.

Issue #313, opened 2026-05-24
First input byte dropped + PSReadLine bell on every freshly-attached pane (DSR response malformed under PSEUDOCONSOLE_WIN32_INPUT_MODE)


braseidon:
Hi, I've been using psmux lately for Claude Code with Agent Teams. I don't know Rust at all (all my experience is web code), so I was hesitant to make an issue until Claude tried a fix, and then I rebuilt psmux and the symptom was fixed. 

My current setup: Windows 11, PowerShell 7, Windows Terminal, basic config (from your docs for use with Claude Code). I built the psmux binary from the latest commit as of today (May 24th). Some of this report is gonna be pasted from what Claude said. Hopefully this explains the issue clearly, if not let me know and I can relay anything you need

Repo steps:
1. Open a fresh Windows Terminal tab.
2. Run `psmux` to start a new session.
3. Wait 10+ seconds (rules out PSReadLine cold-start as cause).
4. Press any key OR Ctrl+V paste any text.

**Expected:** input appears intact.
**Actual:** first character missing; bell ding; second character onward intact.

**Real-world impact:** breaks Claude Code agent-team worker spawning. The team-leader's flow is `split-window` followed by `send-keys "cd 'F:\...' && env ... && claude.exe ..."` into the freshly-spawned worker pane. The leading `c` of `cd` is consumed by the same bug, so the worker pane runs `d 'F:\...'` and immediately errors with `'d' is not recognized as the name of a cmdlet`. The worker never starts and the lead silently hangs waiting for it.

## Suspected cause

`conpty_preemptive_dsr_response()` at `src/pane.rs:28-31` writes a raw VT cursor-position-report (`\x1b[1;1R`, 6 bytes) to the ConPTY input pipe immediately after pane creation. Called from 5 sites: `src/window_ops.rs:1315`, `src/pane.rs:169`, `src/pane.rs:233`, `src/pane.rs:285`, `src/pane.rs:441`, and `src/popup.rs:88`.

The ConPTY is created with `PSEUDOCONSOLE_WIN32_INPUT_MODE` as a base flag (`crates/portable-pty-psmux/src/win/psuedocon.rs:127` and `:177`), which interprets the input pipe as Win32-input-mode escape sequences of the form `\x1b[Vk;Sc;Uc;Kd;Cs;Rc_` - note the underscore terminator. (Exactly the format used by the recent #305 / `60e1714` Ctrl+letter fix.)

[...]

Issue #305, opened 2026-05-20
Set-PSReadLineKeyHandler not working


jlanzarotta:
Hello,

I have the following configuration in my `profile.ps1` file...

```
# History of the commands.
# Install-Module -Name PSReadLine
Import-Module PSReadLine

Set-PSReadLineOption -PredictionSource HistoryAndPlugin
Set-PSReadLineOption -PredictionViewStyle ListView

# Vim Like KeyBinding for Next and Previous History Selection
# Using `Ctrl+j` for Next selection and `Ctrl+k` for Previous selection
# Remove following commands if you want to use just arrow key for navigation
Set-PSReadLineKeyHandler -Key Ctrl+j -Function HistorySearchForward
Set-PSReadLineKeyHandler -Key Ctrl+k -Function HistorySearchBackward
```

It makes if very handy to be able to `Ctrl+j` and `Ctrl+k` up and down through the history list to make a selection. However, once I start `psmux`, these keybinding stop working. Thoughts?

Thanks.

psmux:
Hi @jlanzarotta, thanks for reporting this.

I tested your exact scenario on psmux 3.3.4 and your Ctrl+J / Ctrl+K keybindings for HistorySearchForward / HistorySearchBackward are actually working correctly — the key events are properly forwarded to the PTY and PSReadLine receives them.

However, I suspect the real issue you're seeing is that the prediction ListView is not appearing. By default, psmux sets �llow-predictions off to prevent a PSReadLine crash (https://github.com/psmux/psmux/issues/109) during ConPTY startup. This has a side effect: after your profile loads, psmux forces:

PredictionSource = None PredictionViewStyle = InlineView

This means the prediction dropdown list (which you'd normally navigate with Ctrl+J/K in ListView mode) never shows up — even though your keybindings are still registered.

Fix
Add this to your ~/.psmux.conf (create it if it doesn't exist):

set -g allow-predictions on

With this setting, psmux will preserve your profile's PredictionSource HistoryAndPlugin and PredictionViewStyle ListView settings, and the prediction list will appear as expected.

Verification
I confirmed:

With low-predictions off (default): Ctrl+J/K still work for basic history search (HistorySearchForward/Backward), but PredictionSource gets forced to None so the ListView never shows
With llow-predictions on: Your full config is preserved — PredictionSource = HistoryAndPlugin, PredictionViewStyle = ListView, and Ctrl+J/K navigate the prediction list as intended
[...]

PR #401, opened 2026-06-17
Fix <Esc> (and CPR replies) being swallowed under PSEUDOCONSOLE_WIN32_INPUT_MODE

### LINKED ISSUE #380: Esc key is not being sent over
When using VIM mode the `ESC` key is not being sent to the terminal. I use vim on the terminal and it does not send the key to the editor. 

## Expected
`USER (Esc Key) --> PSMUX --> Vim`

## Actual
`USER (Esc Key) --> PSMUX     Vim`

Here are my settings:
<details>

<summary>`.psmux.config`</summary>

```
set -g escape-time 0
set -g mouse on

bind -n C-k clear-history

set-option -a terminal-features 'xterm-256color:RGB'
set-option -g xterm-keys on

# Set vi key-bindings
set-window-option -g mode-keys vi
bind-key h select-pane -L
bind-key j select-pane -D
bind-key k select-pane -U
bind-key l select-pane -R

# resize panes
bind -r H resize-pane -L 5
bind -r J resize-pane -D 5
bind -r K resize-pane -U 5
bind -r L resize-pane -R 5

# swap windows left/right
bind-key -n C-S-Left swap-window -t -1
bind-key -n C-S-Right swap-window -t +1

# window splitting
unbind c
bind-key c new-window -c "#{pane_current_path}"
unbind %
bind | split-window -h -c "#{pane_current_path}"
unbind '"'
bind - split-window -v -c "#{pane_current_path}"

set-option -g automatic-rename on
set-option -g automatic-rename-format '#{b:pane_current_path}'

# Use vim keybindings in copy mode
setw -g mode-keys vi

set -g @plugin 'psmux-plugins/ppm'
set -g @plugin 'psmux-plugins/psmux-sensible'
set -g @plugin 'psmux-plugins/psmux-theme-rosepine'
set -g @plugin 'psmux-plugins/psmux-yank'
run '~/.psmux/plugins/ppm/ppm.ps1'
```

</details>

## Environment Info

* OS: Windows
*  Shell: Powershell
*  Terminal: Alacritty

ingoncalves:
## Summary

Fixes `<Esc>` becoming unresponsive in TUI apps (Neovim, Claude Code, …) a few seconds after they launch and never recovering — the open bug #380. The root cause is the same `WIN32_INPUT_MODE` ConPTY input-parser hazard that #313 / #305 already addressed for other keys; this PR covers the bare-`<Esc>` keypress path and the reactive CPR responder.

## Root cause

Panes are created with `PSEUDOCONSOLE_WIN32_INPUT_MODE`, whose input parser expects win32 key sequences (underscore-terminated). Two code paths still wrote raw VT to that pipe:

1. **Bare `<Esc>` keypress** — `forward_key_to_active` fell through to `encode_key_event`, which writes a lone `ESC` byte to the PTY pipe. ConPTY treats it as the start of a CSI sequence and buffers it, so once the child enables VT / Win32 input mode (a moment after launch) the key never arrives, and it never recovers. The Ctrl+letter path already injects to avoid exactly this — see the comment it left in `src/input.rs`.
2. **CPR reply** — the reactive `ESC[6n` responder wrote a raw `…R`-terminated cursor-position report into the same pipe, corrupting the parser. This is the reactive twin of the spawn-time DSR bug fixed in #313 (the spawn-time response was made a no-op there, but the reactive path was not).

## Changes

- **`src/input.rs`** — deliver bare `<Esc>` as a `VK_ESCAPE` key event via `WriteConsoleInputW` (`mouse_inject::send_modified_key_event`), bypassing ConPTY's VT input parser. Falls back to the PTY pipe only when no child pid is known. Mirrors the Ctrl+letter approach from #305.
- **`src/server/helpers.rs`, `src/server/mod.rs`** — route CPR replies for the active, warm, and popup panes through `send_vt_sequence` (direct console-input injection) via a shared `deliver_cpr_reply` helper; the raw PTY-pipe write is kept only as a fallback.

## Testing

- Built on `x86_64-pc-windows-gnu`. With the bare-`<Esc>` fix, `<Esc>` works reliably in `nvim --clean` and Claude Code under a fresh server, where it was previously permanently dead within seconds of launch. Verified with `PSMUX_INPUT_DEBUG=1`.
- The CPR change is a code-correctness fix matching the #313 root-cause analysis; no behavior change on non-Windows or when no child pid is available (the pipe fallback path is unchanged there).

Fixes #380. Related: #313, #305.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -2102,6 +2102,42 @@ pub fn forward_key_to_active(app: &mut AppState, key: KeyEvent) -> io::Result<()
                 return Ok(());
             }
         }
+
+        // Bare <Esc>: deliver it as a VK_ESCAPE key event via WriteConsoleInputW
+        // instead of writing a lone \x1b to the PTY pipe.  Once the child app
+        // (Neovim, Claude Code, ...) turns on VT / Win32 input mode a moment
+        // after launch, ConPTY treats a piped \x1b as the start of a CSI
+        // sequence and buffers it, so the keypress never reaches the child and
+        // <Esc> appears dead.  Direct console-input injection bypasses that
+        // parser, mirroring the Ctrl+letter path above (#305).  Fall back to the
+        // pipe only when no child pid is known.
+        if matches!(key.code, KeyCode::Esc) && key.modifiers.is_empty() {
+            fn deliver_esc(p: &mut Pane) {
+                let injected = p.child_pid
+                    .map(|pid| crate::platform::mouse_inject::send_modified_key_event(pid, '\u{1b}', false, false, false))
+                    .unwrap_or(false);
+                if !injected {
+                    let _ = p.writer.write_all(b"\x1b");
+                    let _ = p.writer.flush();
+                }
+                crate::debug_log::input_log("esc",
+                    &format!("deliver_esc injected={} pid={:?}", injected, p.child_pid));
+            }
+            let win = &mut app.windows[app.active_idx];
+            if app.sync_input {
+                fn esc_all(node: &mut Node) {
+                    match node {
+                        Node::Leaf(p) if !p.dead => deliver_esc(p),
+                        Node::Leaf(_) => {}
+                        Node::Split { children, .. } => { for c in children { esc_all(c); } }
+                    }
+                }
+                esc_all(&mut win.root);
+            } else if let Some(active) = active_pane_mut(&mut win.root, &win.active_path) {
+                if !active.dead { deliver_esc(active); }
+            }
+            return Ok(());
+        }
     }
 
     let encoded = match encode_key_event(&key) {
--- a/src/server/helpers.rs
+++ b/src/server/helpers.rs
@@ -386,8 +386,32 @@ fn check_pane_bells(node: &Node) -> bool {
 /// Injects ESC[row;colR into any pane whose reader thread detected ESC[6n.
 /// pwsh re-issues the CPR query after lock/unlock; without this response it
 /// blocks indefinitely since the preemptive write at spawn time is long gone.
+/// Deliver a CPR (cursor-position-report) reply to a child process.
+///
+/// The child ConPTY is created with `PSEUDOCONSOLE_WIN32_INPUT_MODE`, whose
+/// input parser only accepts win32 key sequences terminated with `_`.  Writing
+/// a raw VT CPR reply (terminated with `R`) to the PTY pipe leaves that parser
+/// in a half-state that swallows the next keystroke -- notably a lone <Esc>,
+/// which is why <Esc> stops working a few seconds after an app such as Neovim
+/// issues its `ESC[6n` probe.  This is the reactive twin of the spawn-time DSR
+/// bug fixed in #313.  Inject the reply straight into the child's console input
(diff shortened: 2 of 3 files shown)


psmux:
Thanks for this @ingoncalves. The writeup and the root cause analysis are genuinely excellent, and tying it back to #313 and #305 is exactly the right instinct.

Before merging I tried hard to reproduce the underlying bug on current master, because my rule is to prove a defect tangibly before changing code. I could not get Esc to die here, and I want to lay out what I ran so we can find the gap together.

Setup: current master, Windows 11 build 26200. I built a harness around a real ConPTY (the same primitive Alacritty and WezTerm spawn) so psmux receives input exactly as it does under Alacritty, with zero dependence on window focus.

Results across four methods:

1. Bare Esc delivered as a native key event into psmux, forwarded to a child that has VT input mode on: the child receives 0x1b.
2. Same, right after a cursor position query and reply: Esc still arrives.
3. Raw 0x1b fed through the real ConPTY into psmux, before and after psmux answers ESC[6n: the child gets ESC[..R and then a clean 0x1b for the next Esc.
4. Real nvim inside psmux inside the ConPTY: insert text, Esc, write buffer. nvim leaves insert mode and writes the file, which only happens if Esc reached it.

On master today, the lone 0x1b that encode_key_event writes to the pane pipe does reach the child, and the raw CPR reply does not swallow the next keystroke here. So neither of the two failure modes this PR targets reproduces on my build.

My read is that this was real on your Windows / ConPTY version but has since been mitigated, either OS side in a newer conhost or by the input work that landed after the bug was filed. Your change looks low risk and mirrors the accepted #305 pattern, and in my tests it produces identical delivery, so I am not worried about regressions. I just do not want to merge a fix for a defect I cannot make appear, and then be unable to prove it did anything.

Could you share:

1. Your exact Windows build (winver output).
2. Whether a fresh build of current master still shows dead Esc for you.
3. A `PSMUX_INPUT_DEBUG=1` log around a dead Esc, ideally showing the byte the child pipe receives.

If we can get a reproduction that fails on master, I will pull this in immediately and add it to the test suite so it never comes back. Really appreciate the effort you put into this one.

psmux:
Guilherme, I am going to close this one for now, purely as housekeeping and not as a judgment on the work, which remains the most thorough root cause writeup this repo has received.

Where we left it in July: I built a real ConPTY harness and tried four different ways to make Esc die on current master (native key event, post CPR, raw 0x1b through ConPTY, and real nvim insert/Esc/write), and every one delivered the Esc. Without a failing reproduction on my side I cannot prove your change does anything here, and merging unprovable fixes is how input regressions sneak in. I asked for your Windows build number, a check against fresh master, and a `PSMUX_INPUT_DEBUG=1` log around a dead Esc, and I have not heard back, which is completely fine, life happens.

The offer stands exactly as before: if you (or anyone reading this later with a dead Esc in nvim or Claude Code) can share a winver and a debug log showing the swallowed byte, I will reopen this immediately, merge the fix, and add the repro to the suite so it never comes back. The #313/#305 lineage you traced is preserved here for whoever hits it next. Thank you again for the depth of the analysis.

PR #403, opened 2026-06-18
refactor: centralize psmux data-dir path construction into paths.rs

marhel:
This PR is a refactoring that centralizes the psmux data-dir path construction into `paths.rs`. This makes future changes to the exact name and location of the data dir or the port/key/sid files trivial.

This PR introduces `src/paths.rs` as the single source of truth for the psmux data directory and the files under it, and routes every call site through it.

- New helpers: `home_dir()`, `psmux_dir()` / `psmux_dir_opt()`,
  `port_file()` / `key_file()` / `sid_file()` / `port_file_opt()`,
  `psmux_dir_file()`.
- Converts all load-bearing sites (`.port`/`.key`/`.sid`, `next_session_id`, `last_session`, warm-pool/claim files, `latency.log`/ `crash.log`/ `cc_debug.log`) across session, main, server, client, preview, commands and cross_session.
- Drops the now-dead `home` parameter from `preview::fetch_*`/`get_or_fetch*` and `server::ensure_session_registry_files`.

Forward-slash sites (`debug_log.rs`, `ssh_input.rs`, `util.rs`, `#{socket_path}`) are intentionally left inline here — routing them through `psmux_dir_file` would normalize the separator (an observable change), so that is left for a later PR.

**No observable change.** `psmux_dir()` returns exactly the previous inline `format!("{}\.psmux", home)`, so every produced path is byte-identical. Verified by the full unit suite staying green.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1366,13 +1366,12 @@ fn try_reconnect(addr: &str, key: &str) -> Option<Connection> {
 
 pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::PsmuxWriter>>, input: &crate::ssh_input::InputSource) -> io::Result<()> {
     let name = env::var("PSMUX_SESSION_NAME").unwrap_or_else(|_| "default".to_string());
-    let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
-    let path = format!("{}\\.psmux\\{}.port", home, name);
+    let path = crate::paths::port_file(&name);
     let port = std::fs::read_to_string(&path).ok().and_then(|s| s.trim().parse::<u16>().ok())
         .ok_or_else(|| io::Error::new(io::ErrorKind::Other, format!("can't find session '{}' (no server running)", name)))?;
     let addr = format!("127.0.0.1:{}", port);
     let session_key = read_session_key(&name).unwrap_or_default();
-    let last_path = format!("{}\\.psmux\\last_session", home);
+    let last_path = crate::paths::psmux_dir_file("last_session");
     if !crate::session::is_warm_session(&name) {
         let _ = std::fs::write(&last_path, &name);
     }
@@ -1893,8 +1892,7 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
     // Diagnostic latency log: set PSMUX_LATENCY_LOG=1 to enable
     let latency_log_enabled = env::var("PSMUX_LATENCY_LOG").unwrap_or_default() == "1";
     let mut latency_log: Option<std::fs::File> = if latency_log_enabled {
-        let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
-        let path = format!("{}\\.psmux\\latency.log", home);
+        let path = crate::paths::psmux_dir_file("latency.log");
         std::fs::File::create(&path).ok()
     } else { None };
     let mut loop_count: u64 = 0;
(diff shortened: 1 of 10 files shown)


psmux:
Martin, I like this refactor and I want it, but it has drifted out of date: master has since grown new path call sites (the `.pid` anchor files from #448, warm spawn lock files from #444, and assorted churn in `server/mod.rs` and `session.rs`), so the branch now conflicts and, more importantly, a fresh sweep would catch call sites that did not exist when you wrote it.

Could you rebase and re run the conversion against current master? The design is exactly right: one source of truth in `paths.rs`, byte identical output so nothing observable changes, and the forward slash sites deliberately left alone rather than silently normalized. That last decision is the kind of restraint that makes a refactor safe to merge.

Once rebased I will merge this ahead of #404 so the kill-server work can sit on top of it, as you planned. Thank you for doing the unglamorous work here.

psmux:
Thank you Martin. This is a clean centralization and I do want it in. Holding only for a rebase, not for any concern with the change itself.

It currently conflicts with master, and it touches many of the same call sites in `main.rs` and `session.rs` that several fixes just landed on (#463, #464, #436 and #465 all merged today). Could you rebase this onto current master now that those are in? Since it is a pure path helper extraction with no observable change (every produced path stays byte identical), once it rebases green it is an easy merge and I will take it right away.

marhel:
I'm working on it!

marhel:
Redid this against current master from scratch: I reset the branch to master and re-applied the refactor as a single fresh commit, doing the conversion as a new sweep rather than replaying the old diff, so the call sites that landed since I first wrote this are covered too. The branch history is one new commit on top of master, not the old commit rebased. 

What's new since the previous version: the sweep now also routes the `.pid` anchor files from #448 (session write/remove, the registry writer, and the crash-hook cleanup) and the warm `.spawnlock` files from #444, plus the additional `regpath`/`keypath` sites that churn added to `server/mod.rs` (rename, respawn, etc.). To keep those in the same style as `.port`/`.key`/`.sid`, I added two helpers alongside the existing ones: `pid_file()` and `spawnlock_file()`. `ensure_session_registry_files` still writes the bare `.pid` exactly as before - this is a path-construction refactor only, so #448's reader is untouched and unaffected.

Everything else is unchanged in spirit: `psmux_dir()` returns exactly the previous inline `format!("{}\.psmux", home)`, so every produced path is byte-identical, and the fallible helpers (`psmux_dir_opt`/`port_file_opt`) preserve each call site's early-return-on-missing-home behavior. The forward-slash sites (debug logs, config paths, `#{socket_path}`) and the two `~` home-expansion sites are deliberately left alone, since routing them through the helpers would normalize the separator. Normalizing may be a good thing, but is left for a different PR.

psmux:
Hey Martin, thank you for this one. I put it through a full pass before merging since it touches the path construction that basically every session, warm pool, and registry file operation in psmux depends on, and I wanted to be sure "no observable change" actually held up rather than just take the claim on faith.

Here is what I checked:

Read through the entire diff line by line looking for anything suspicious. It is exactly what the description says, a mechanical swap of inline format strings for calls into the new paths.rs helpers. Nothing sneaky, no new network or filesystem behavior, nothing that reads or sends data anywhere it should not.

Ran the full unit suite on your branch, all 2338 tests green, including your 9 new paths.rs tests and the existing plugin path discovery tests (those are a separate code path that your PR correctly left alone).

Built and installed the branch for real and ran a 16 point end to end pass: cold new session, TCP command path, rename session (verified the port/key files actually move to the new base and the old ones disappear), kill session cleanup, warm pool claim and timing, list sessions and kill server directory iteration, and the -L namespace prefix path. All 16 passed.

Launched a real attached TUI window and drove it through split window, zoom, new window, and capture pane to make sure nothing in the rendering or TCP dispatch path regressed.

Specifically checked plugins and themes since those matter a lot to people building on top of psmux. Plugin discovery in config.rs uses its own separate path logic that your PR does not touch at all, and I confirmed a plugin actually loads and its option is readable. Status style and theming also work fine.

Everything checked out clean, so this is merged as a1cb81f. The paths.rs module you added is a real improvement, it is one of those refactors that pays for itself the next time someone has to touch the port or key file logic, which happens more often in this codebase than I would like. Appreciate the careful "no observable change" framing in the description too, it made the review a lot easier to reason about.

Thanks again for the contribution.

PR #404, opened 2026-06-18
fix(kill-server): scope the force-kill fallback to the data dir, by pid

marhel:
This PR stops `kill-server` from killing potentially unrelated processes that happen to be named `psmux.exe` it will now only terminate known pids (with a re-used pid guard based on process creation time) - but beware, it will still happily kill your own namespaced work sessions when running a bare `kill-server` (without -L). Tmux-faithful scoping of that remaining over-reach is a follow-up.

## Problem

After its graceful pass, a bare `kill-server` (without -L) ran a force-kill fallback that enumerated *every process on the machine* and `TerminateProcess`'d anything whose exe name was `psmux.exe` / `pmux.exe` / `tmux.exe`. It matched on the *name*, not on whether the process was actually one of our servers, and it ignored the data dir and namespace entirely. tmux's `kill-server` is socket-scoped and never crosses to another socket's server.

## Repro

```
# Start the super important MindServer
C:\Program Files\PSy Mind Upload Xplorer\psmux.exe --with-lifesupport --critical --redundancy 3
LOG: MindServer started with 3x redundancy. I'd like to see the bug that brings THIS down!
                      # meanwhile, in a different terminal
                      cd C:\dev
                      psmux new-session -s hacking
                      ...
                      psmux kill-server
LOG: MindServer #1 terminated
LOG: MindServer #2 terminated
LOG: MindServer #3 terminated
LOG: Lifesupport disabled, flatlining
```

Hypothetical 3-way redundancy is no help when the reaper matches on the *name* — all three live on one machine, all three are `psmux.exe`. (Same user here; a non-elevated `kill-server` is bounded by `OpenProcess` rights to your own processes — but across all your sessions, namespaces, and data dirs. Run elevated, it reaches other users' processes too.)

A harmless, actually-runnable version:

```
mkdir %TEMP%\unrelated && copy C:\Windows\System32\calc.exe %TEMP%\unrelated\psmux.exe
& %TEMP%\unrelated\psmux.exe
psmux kill-server          # in a separate session
# Before: %TEMP%\unrelated\psmux.exe is terminated (exit code 1).
# After:  untouched — only this data dir's own servers are reaped.
```

## Fix

A fallback scoped by construction to the registry:

- The server records `{pid}:{creation_filetime}` in `{base}.pid` alongside `.port`/`.key`/`.sid`.
- `force_kill_targets(dir, ns_prefix)` selects candidates only from *this* data dir's `.pid` files, filtered by the same `-L` prefix as the graceful pass — a namespaced `kill-server` never reaches another namespace.
- Before terminating, `confirms_identity` requires the live process's creation time to *exactly* match the recorded one. A recycled pid has a different creation time and is rejected; an unqueryable process is rejected too — fail-safe, no kill on uncertainty. No name matching anywhere.

With 3 sessions in each of 3 namespaces + default (12 servers, one data dir): `-L ns1 kill-server` reaps exactly ns1's 3 (force-killing only the wedged ones); bare `kill-server` reaps the 12 in this data dir — and nothing outside it, ever.

**Not full tmux parity yet.** A bare `kill-server` still graceful-kills every namespace in *this* data dir, where tmux would scope to the default socket only (see follow-up). What this change fixes is the dangerous part: `kill-server` no longer kills *unrelated* processes — nothing outside this data dir, and never by exe name. The over-reach that remains is confined to your own psmux sessions in your own data dir; it can no longer flatline the MindServer next door.

## Testing

Unit tests cover the selection and the gate: dir scoping, namespace scoping, exact-match, fail-safe-on-`None`, FILETIME packing, and a pid-file producer/consumer round trip. A ps1 integration test drives a real `kill-server` against an unrelated `pwsh` sleep used as a stand-in wedged server (exploiting that we no longer name-check), confirming the real `GetProcessTimes` path: in-scope + matching → killed; creation-time mismatch (pid reuse) → spared; other namespace → spared. Both the "kill" and "spare" directions were verified non-vacuous via edit-to-red. The integration test refuses to run unless `PSMUX_TEST_SANDBOX=1` (and is pointed at the build under test via `PSMUX_EXE`), since a real `kill-server` against a pre-fix or PATH binary is destructive.

## Follow-up (separate work)

- The ps1 tests cite bare `kill-server`'s "kill-all-by-image-name fallback" as the reason to avoid it. That sweep is gone — a bare `kill-server` will no longer terminate *unrelated* processes (the MindServer next door is safe). But the helpers should still avoid bare `kill-server`, for a sharper reason: it still graceful-kills *every one of your own* psmux sessions in the data dir (all namespaces). In practice that has repeatedly wiped my own live sessions when an agent or test fired a bare `kill-server`. Tests and tooling should use `-L <ns>`, never bare. 

The structural fix to only kill servers within a named (or default) namespace is in the works, and PR #403 will make that change safe and trivial.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -51,7 +51,7 @@ use crate::platform::enable_virtual_terminal_processing;
 use crate::cli::{print_help, print_version, print_commands};
 use crate::session::{cleanup_stale_port_files, reap_orphaned_servers, read_session_key, send_control,
     send_control_with_response, resolve_default_session_name,
-    kill_remaining_server_processes};
+    force_kill_targets, confirms_identity};
 use crate::rendering::apply_cursor_style;
 use crate::server::run_server;
 use crate::client::run_remote;
@@ -462,6 +462,11 @@ fn run_main() -> io::Result<()> {
             let psmux_dir = crate::paths::psmux_dir();
             // Compute namespace prefix for -L filtering (matches list-sessions behavior)
             let ns_prefix = l_socket_name.as_ref().map(|l| format!("{l}__"));
+            // Snapshot the force-kill candidates from this data dir's .pid files
+            // BEFORE the graceful pass removes them. Scoped to this dir and (with
+            // -L) this namespace, so the fallback can never reach another instance.
+            let fk_targets =
+                force_kill_targets(std::path::Path::new(&psmux_dir), ns_prefix.as_deref());
             let mut targets: Vec<(std::path::PathBuf, u16, String)> = Vec::new();
             let mut stale_ports: Vec<std::path::PathBuf> = Vec::new();
             if let Ok(entries) = std::fs::read_dir(&psmux_dir) {
@@ -512,25 +517,31 @@ fn run_main() -> io::Result<()> {
                             }
                         }
                     }
-                    // Remove port/key files regardless
+                    // Remove port/key/pid files regardless
                     let _ = std::fs::remove_file(&path);
-                    let key_path = path.with_extension("key");
-                    let _ = std::fs::remove_file(&key_path);
+                    let _ = std::fs::remove_file(path.with_extension("key"));
+                    let _ = std::fs::remove_file(path.with_extension("pid"));
                 })
             }).collect();
             // Wait for all threads to complete
             for h in handles { let _ = h.join(); }
-            // Clean up stale port/key files
+            // Clean up stale port/key/pid files
             for path in &stale_ports {
                 let _ = std::fs::remove_file(path);
-                let key_path = path.with_extension("key");
-                let _ = std::fs::remove_file(&key_path);
+                let _ = std::fs::remove_file(path.with_extension("key"));
+                let _ = std::fs::remove_file(path.with_extension("pid"));
             }
-            // Brief wait then verify no processes remain; if any do, force-kill them.
-            // Only do the nuclear fallback when not using -L namespace filtering.
+            // Force-kill any wedged server that ignored the graceful kill. The
+            // candidates were read from this data dir's (and, with -L, this
+            // namespace's) own .pid files before the graceful pass removed them,
+            // so nothing outside this instance is ever reached. The identity gate
+            // (exact process-creation-time match) skips any pid that has already
+            // exited or been recycled — no machine-wide, name-based scan.
             std::thread::sleep(Duration::from_millis(50));
-            if ns_prefix.is_none() {
-                kill_remaining_server_processes();
+            for t in fk_targets {
+                if confirms_identity(crate::platform::process_kill::process_creation_time(t.pid), t.creation_time) {
+                    crate::platform::process_kill::terminate_server_pid(t.pid, None);
+                }
             }
             return Ok(());
         }
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -1956,25 +1956,25 @@ pub mod process_kill {
         descendants
     }
 
-    /// Force-terminate a single process by PID, but ONLY if its creation time is
-    /// no later than `max_creation_ft` (issue #447 PID-reuse guard).
+    /// Force-terminate a single process by PID, guarded against PID reuse by
+    /// process creation time (issue #447).
     ///
-    /// The caller captures `max_creation_ft` (via `now_filetime()`) immediately
-    /// BEFORE taking the process snapshot that identified `pid`. Any process
-    /// that was legitimately enumerated in that snapshot must have been created
-    /// at or before the cutoff. If the PID has since been reused by an unrelated
-    /// process, that new process was created AFTER the cutoff, so its creation
-    /// time exceeds `max_creation_ft` and we refuse to terminate it.
+    /// `max_creation_ft` is `Some(cutoff)` for callers that identified `pid` from
+    /// a process snapshot: they capture the cutoff (via `now_filetime()`)
+    /// immediately BEFORE snapshotting, so any legitimately enumerated process
+    /// was created at or before it. A pid since reused by an unrelated process
+    /// was created AFTER the cutoff, so its creation time exceeds it and we refuse
+    /// to kill; a pid we cannot query is skipped too (fail safe).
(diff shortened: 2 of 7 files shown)


psmux:
Thank you Martin. The safety goal here is exactly right, and the identity gate design (pid plus creation FILETIME, fail safe on an unqueryable process) is clean. I cannot merge it as it stands though, and I want to flag the real blocker so a rebase is worth your time.

Master already writes a `.pid` file for every session (issue #448, the liveness anchor). It stores a bare pid, and `ensure_session_registry_files` in `server/mod.rs` already writes it. This PR repurposes that same `.pid` file to hold `pid:creation_filetime`. Two things fall out of that:

1. It now conflicts on that block (GitHub already marks it CONFLICTING).
2. More importantly, #448's reader parses `.pid` as a bare `u32`. The new `pid:creation` body would break the startup orphan reaper that master now depends on. That is a silent regression, not just a textual conflict.

Could you rebase onto current master and reconcile with the existing #448 `.pid` sentinel? Two workable paths: either extend that one file's format to `pid:creation` and update #448's reader in the same PR, or leave `.pid` as the bare pid and query the creation time at kill time instead of storing it. Once it is reconciled and the suite is green I will happily take it. The core win, that `kill-server` stops reaching unrelated processes by image name, is genuinely worth having.

marhel:
Reworked this against current master (now that the #403 path helpers have landed) and reconciled it with the #448 `.pid` anchor, so it no longer conflicts or regresses the reader. The branch is a single fresh reimplementation on top of master rather than the old commits rebased, and it is now `MERGEABLE`.

On the `.pid` collision you flagged: rather than fork the format or drop the identity gate, the `.pid` body is extended to `pid:creation_filetime` and a single `parse_pid_file_contents` is now shared by both the new force-kill selector and #448's existing readers (`read_tracked_registry` and the pid-anchor verdict). That parser still accepts a bare `pid`, so a server written by an older build mid-upgrade - or any other bare-pid writer - is read exactly as before and the startup orphan reaper never regresses. The creation time is written through the existing registry writer and the rename path, reusing #403's `pid_file()`.

The identity gate is intact: `force_kill_targets` selects only this data dir's `.pid` files (filtered by the same `-L` prefix as the graceful pass), and `confirms_identity` requires an exact process-creation-time match before terminating - a recycled pid or an unqueryable process is skipped, so the fallback fails safe. It reuses #447's `process_creation_time` and terminate primitives instead of carrying its own copy, and the machine-wide name-based killer is removed outright.

One deliberate behavior change beyond the safety fix: the force-kill pass now also runs under `-L`, scoped to that namespace. Previously it was gated to the no-`-L` case, so `-L foo kill-server` sent the graceful kill and then silently left a wedged foo server running. Running the scoped, identity-checked pass under `-L` makes the force pass mirror the graceful pass - it finishes the job for exactly the sessions we were asked to kill - and is safe by construction (only that namespace's own `.pid` files, exact identity match). The no-`-L` policy is unchanged: it still kills every server in the data dir regardless of namespace.

Minor cleanup along the way: the #447 PID-reuse guard cutoff is now `Option<u64>` (`None` when identity is already settled by an exact creation-time match or a first-class child handle) rather than a `u64::MAX` sentinel.

Selection and identity gate are unit-tested (dir/namespace scoping, bare and malformed pid files skipped, exact match, pid-reuse rejection, unqueryable fail-safe, parse/format round trip). The OS shims and wiring are covered by a sandbox-gated integration test, since the only thing that would make such a test fail is the machine-wide kill itself, which cannot be sandboxed.

psmux:
Merged. Thank you for the careful rework, this landed in exactly the shape I was hoping for.

I did not take the writeup on faith, I verified the whole thing first, and the part I most wanted to pin down was the tmux question.

I read tmux's `cmd-kill-server.c` directly. Its exec body is literally `kill(getpid(), SIGTERM)`, so the tmux server only ever signals itself. It never enumerates processes, never matches on an image name, and never reaches across to another socket's server. That makes psmux's old fallback the real outlier: a whole machine `CreateToolhelp32Snapshot` sweep that `TerminateProcess`'d anything named `psmux.exe` / `pmux.exe` / `tmux.exe`, ignoring the data dir and namespace entirely. Your change deletes that sweep and replaces it with a selection scoped by construction to this data dir's own `.pid` files, gated on an exact process creation time match. That is the right direction, and I agree it is not full parity yet (a bare `kill-server` still graceful kills every namespace in the data dir). Closing that remaining overreach in #403 is the correct sequencing. Removing the dangerous part now was the right call.

What I actually ran, every bit of it against a fresh build of the branch and isolated by `USERPROFILE` so none of my live sessions were ever in scope:

**Rust suite:** 2358 passing. The one red was `config::tests_plugin_paths::plugin_discovery_xdg_short_name`, which is a pre existing parallel run flake (a fixed temp path colliding with a sibling XDG test). It passes in isolation on master too, so it is unrelated to this PR.

**Your sandbox integration test** (`PSMUX_TEST_SANDBOX=1`): 3 of 3. In scope with a matching identity is force killed, a creation time mismatch (simulated pid reuse) is spared, and a probe registered under namespace `a` is spared by `kill-server -L b`.

**My own end to end pass on top of yours:** a real process literally named `psmux.exe` with no registry entry survives a bare `kill-server` (the old build would have terminated it purely by name), the `.pid` body now reads `pid:creation_filetime`, a default namespace session survives `kill-server -L otherns`, `list-sessions` and the startup orphan reaper still behave correctly with the new `.pid` format, and a bare `kill-server` still reaps its own data dir's sessions with zero collateral to the unrelated `psmux.exe`.

I also checked the upgrade path on purpose, since my sessions on the currently installed build still carry bare `.pid` files. Your shared `parse_pid_file_contents` reads those fine, so the reaper and the pid anchor keep working, and `force_kill_targets` simply skips them because they carry no recorded creation time, which is exactly the fail safe I wanted. Folding the #447 guard into `Option<u64>` and reusing #448's readers rather than forking the format is a nice touch.

Shipping it. Thanks again.

Issue #285, opened 2026-05-10
Mouse does not work in Neovim inside psmux


lennykean:
Mouse input does not work in native Windows Neovim when running inside psmux. Clicking, dragging, and scrolling work in Neovim outside psmux, and mouse support works in psmux itself, but Neovim inside psmux does not respond to mouse input.

Environment:
- Windows 10 `10.0.19044`
- psmux built from current `master` (`2b5a200`)
- Native Windows Neovim `v0.12.2`
- Config includes `set -g mouse on`, `set -g mouse-selection off`, and `set -g scroll-enter-copy-mode off`

psmux:
Thanks for the detailed report and config example, that helped a lot in tracking this down.

This was a resurrected bug. The heuristic fallback (`is_fullscreen_tui`) that detects TUI apps by screen content had previously been guarding mouse forwarding, but commit `719e604` consolidated the click/drag/hover checks behind `pane_wants_mouse()` without carrying the heuristic along. After that, `pane_wants_mouse()` only checked `mouse_protocol_mode()` and `alternate_screen()`, both of which ConPTY strips on older Windows builds, so all mouse events were silently dropped for apps like Neovim.

The fix restores `is_fullscreen_tui(pane)` as a tier-3 fallback inside `pane_wants_mouse()`, so if ConPTY has stripped the usual escape sequences, the function still returns `true` when the screen looks like a fullscreen TUI (cursor near bottom, last rows filled). The scroll handlers in `input.rs` and `window_ops.rs` were also unified to go through `pane_wants_mouse()` for consistency.

Fixed in `1b62ff8`. The test suite (including 10 new E2E tests covering click, scroll, your exact config, and a false-positive check at the shell prompt) all pass clean.

Your config:
```
set -g mouse on
set -g mouse-selection off
set -g scroll-enter-copy-mode off
```
works exactly as expected now. Mouse clicks in Neovim move the cursor, scroll works, and the shell prompt is unaffected.

PR #407, opened 2026-06-20
fix(#381) Improve fullscreen TUI detection by excluding shell processes

### LINKED ISSUE #381: psmux using git bash shell prints raw mouse escape sequences after running commands
# Mouse events are printed as raw escape sequences after running commands in Git Bash

## Description

When using Git Bash inside psmux with mouse support enabled, mouse handling initially works correctly. Scrolling enters copy mode as expected.

After running a command such as `ls` or `which`, mouse events begin appearing as raw escape sequence text in the terminal. Scrolling and mouse movement no longer behave normally and instead print characters directly to the shell.

Example output:

```text
15M65;61;15M65;61;15M64;61;15M64;61...
```

## Steps to Reproduce

1. Launch psmux configured to use Git Bash.
3. Verify that scrolling correctly enters copy mode.
4. Run a command such as:

   ```bash
   ls
   ```

   or

   ```bash
   which git
   ```
5. Move the mouse or scroll the mouse wheel.

## Expected Behavior

Mouse events should continue to be handled by psmux, and scrolling should enter copy mode

## Actual Behavior

Mouse movement and scrolling are emitted to the terminal as raw escape sequence text instead of being handled by psmux.

## Environment

* OS: Windows
* Shell: Git Bash
* psmux mouse mode: enabled
* Windows Terminal

## Additional Information

* The issue does not occur when using PowerShell/CMD through psmux.
* Git Bash works correctly when launched outside of psmux.
* Mouse support works correctly when the session first starts.
* The issue appears only after executing commands within the shell.
* Behavior happens in other terminal emulators not just windows terminal

jassanw:
This fixes an issue where Git Bash was incorrectly detected as a fullscreen TUI application. I added a check to see if the foreground process is a shell program before the heuristic logic is run. This way shell programs cannot be misclassified as full screen TUI apps.  

fixes #381

### DIFF
--- a/src/window_ops.rs
+++ b/src/window_ops.rs
@@ -158,6 +158,17 @@ pub(crate) fn is_fullscreen_tui(pane: &Pane) -> bool {
         if screen.alternate_screen() {
             return true;
         }
+
+        // Check if forground process is a shell program
+        let forground_is_shell = pane
+            .child_pid
+            .and_then(crate::platform::process_info::foreground_is_shell)
+            .is_some();
+
+        if forground_is_shell {
+            return false;
+        }
+
         // Heuristic: check if many of the last rows are non-blank AND the
         // cursor is near the bottom.  Fullscreen TUI apps fill the entire
         // screen and keep the cursor near the bottom (status bars, menus).


psmux:
Merged, thank you @jassanw. Your diagnosis was the hard part and it was correct: the fill heuristic false positives for a git bash prompt once output fills the screen, and gating it on "is the foreground a shell" is the right call.

I pushed one small follow up correction on top of your commit (7798f65). Your gate used `foreground_is_shell(pid).is_some()`. That helper returns three states: `Some(true)` for a shell, `Some(false)` for a genuine non shell fullscreen app (nvim, htop), and `None` when the snapshot fails. Since `.is_some()` is also true for `Some(false)`, a real TUI would have been forced to "not fullscreen" and lost the tier 3 mouse detection that #285 relies on under ConPTY builds that strip the mouse DECSETs. I changed it to `== Some(true)` so only an actual shell foreground suppresses the heuristic.

I also added regression tests that exercise both paths with live processes (bash and a non shell), so this stays pinned in both directions. Credited you as co author on the follow up commit. Really appreciate the thorough debugging on this one.

PR #411, opened 2026-06-21
Fix Windows Ctrl+Enter forwarding

### LINKED ISSUE #409: [WT] ctrl+enter produces `/r` instead of `n` ?in WT (breaks e.g. pi's newline etc.)
Using WT (pre-kitty support, v1.24), pwsh, psmux v3.3.6 starting [pi](https://github.com/earendil-works/pi) (v0.79.8) in tmux leads to ctrl+enter (new line) sequence not being recognized.

Pi warns about possible tmux sequence problems and recommends ` Warning: tmux extended-keys is off. Modified Enter keys may not work. Add `set -g extended-keys on` to ~/.tmux.conf and restart tmux.` but from I found this config is not supported by psmux. 

In WT/WSL/tmux this combination works well (even without the tmux config).

I'm happy to help debug :)

petrroll:
(the PR is llm generated, including the description, but I've *extensively* validated the claims as well as I could, having very limited background in terminal apps in general or console infra on windows specifically)

## Summary

This PR proposes a Windows-specific fix for plain Ctrl+Enter forwarding in psmux panes.

Fixes #409.

When running directly in Windows Terminal, plain Ctrl+Enter is exposed to raw/Node-style stdin readers as LF (`0x0A`) rather than CR (`0x0D`). Windows Terminal's regular input encoder documents and implements this behavior in `TerminalInput::_encodeRegular`:

- https://github.com/microsoft/terminal/blob/47253ef22bd4e97b4d91ee3941b8f77646ee354e/src/terminal/input/terminalInput.cpp#L863-L886

psmux previously routed live Ctrl+Enter on Windows through native `WriteConsoleInputW` injection with a `VK_RETURN` event whose character payload was CR. That preserves Ctrl metadata for Console API readers, but VT/raw stdin readers only see the translated character payload, so Node/libuv-style apps receive `\r` and treat Ctrl+Enter like ordinary Enter.

This change makes plain Ctrl+Enter use a native `VK_RETURN` event with Ctrl metadata and an LF payload (`u_char = '\n'`), with an LF byte fallback when native injection is unavailable. The intent is to mirror Windows Terminal's single-event behavior: Console API readers can still see the Ctrl+Enter key metadata, while byte-stream readers see LF.

## What changed

- Plain Windows Ctrl+Enter in the live key path now calls `send_modified_enter_event(pid, true, false, false)` and falls back to writing `\n` if native injection fails.
- `send_modified_enter_event` now uses LF as the `u_char` payload only for plain Ctrl+Enter (`ctrl && !alt && !shift`); other Enter variants keep CR.
- `encode_key_event` now returns LF for Windows plain Ctrl+Enter as the fallback byte encoding.
- Existing Enter tests were updated to expect LF for Windows plain Ctrl+Enter while preserving non-Windows CSI expectations.

## Why this approach

I considered two alternatives and avoided them:

1. **Write a VT/CSI sequence such as `ESC [ 13 ; 5 ~` for Ctrl+Enter.**
   That can help some terminal-protocol readers, but it does not provide native Console API key metadata to programs like PSReadLine, and the existing code/comments suggest `CSI 13;mod~` is not a reliable ConPTY path.

2. **Send both a native key event and a separate byte/VT sequence.**
   That could give different reader styles something to consume, but it risks double-delivery for programs or layers that observe both routes.

Using a single native `VK_RETURN` event with an LF payload seems closer to Windows Terminal's own regular Ctrl+Enter behavior and avoids duplicating input.

## Caveats / known limitations

- This PR was LLM-generated. It has been manually reviewed by me, but I am not familiar with the psmux codebase, so the approach may be architecturally wrong. Please treat it as an implementation idea/proposal rather than a confident maintainer-quality fix.
- This has **not** been runtime-tested. I don't currently have access to a windows machine where I could compiler stuff (hopefully during weekday). 
- VS Code diagnostics reported no errors for the edited files.
- A follow-up may be needed for direct `send-keys C-Enter` paths in `src/server/mod.rs` / `src/commands.rs`, which may still expand `C-Enter` through `parse_modified_special_key` to CSI bytes. This means there'd be divergence but it's probably fine as most apps should understand CSI bytes. And it worked that way before so let's keep it.
- Ctrl+Shift+Enter and Ctrl+Alt+Enter are intentionally left on the existing behavior until their desired Windows Terminal parity is confirmed.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1810,6 +1810,18 @@ pub fn encode_key_event(key: &KeyEvent) -> Option<Vec<u8>> {
             let m = modifier_param(key.modifiers);
             if m > 1 {
                 // On Windows, CSI 13;mod~ is non-standard and dropped by ConPTY.
+                // Windows Terminal reports Ctrl+Enter to raw/Node-style readers
+                // as LF (0x0A). The live Windows path injects VK_RETURN with the
+                // same LF payload; this is the fallback byte encoding.
+                #[cfg(windows)]
+                {
+                    let has_ctrl = key.modifiers.contains(KeyModifiers::CONTROL);
+                    let has_shift = key.modifiers.contains(KeyModifiers::SHIFT);
+                    let has_alt = key.modifiers.contains(KeyModifiers::ALT);
+                    if has_ctrl && !has_shift && !has_alt {
+                        return Some(b"\n".to_vec());
+                    }
+                }
                 // Send ESC+CR (\x1b\r) for Shift/Alt+Enter — the same bytes VS Code's
                 // xterm.js sends.  libuv preserves ESC as Alt prefix, so Node.js apps
                 // (Claude Code) receive \x1b\r and interpret it as Shift+Enter.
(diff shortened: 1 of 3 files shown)


psmux:
Thanks so much for this, @petrroll. Your diagnosis was correct and the pointer to `TerminalInput::_encodeRegular` was exactly right.

I confirmed the bug on Windows Terminal with a Node raw stdin receiver in a pane (Ctrl+Enter arrived as `0x0D` instead of `0x0A`), and I also verified the mechanism your PR relies on: injecting `VK_RETURN` plus Ctrl with an LF character payload does make libuv readers emit `0x0A`, so the approach is sound.

I ended up landing a smaller variant of your idea in 1bf846a so I could keep the existing sync input and CSI fallback structure in `forward_key_to_active` untouched and limit the blast radius. `send_modified_enter_event` now uses an LF payload for plain Ctrl+Enter only, and the interactive path plus the byte fallback follow suit. Same net behavior you proposed for the reported case. I kept the send-keys server path as is for now, matching the follow up note in your description.

Because of that I am going to close this PR, but the fix is entirely thanks to your report and investigation. Really appreciate the effort you put into this.

PR #417, opened 2026-06-24
fix(config): strip inline # comments from config lines (#416)

### LINKED ISSUE #416: .psmux.conf parse comment in line fail
.psmux.conf parse comment in line fail as follows.
```conf
set -g base-index 1   # test
```
This conf do not take effect after tmux restart while the following do.

```conf
# test
set -g base-index 1
```

The comment in line do not be parsed correctly.

fimilfaneea:
Fixes #416.

## Problem

The config executor only skipped *whole-line* comments (lines whose first non-whitespace character is `#`). A line with a trailing comment such as:

```conf
set -g base-index 1   # test
```

was handed verbatim to the option parser, so the directive silently had no effect — while the same line without the comment worked.

## Fix

Add `strip_inline_comment`, applied in `parse_config_line` right after the leading-`#` check. It removes an **unquoted, whitespace-preceded** `#` comment while preserving, to match tmux's tokeniser:

- `#` inside single or double quotes (e.g. a `"#{...}"` format string),
- an escaped `\#`, and
- a mid-word `#` (e.g. a `colour#aabbcc` token).

Because `parse_config_line` is the single choke point for executing config directives (config load, `source-file`, brace `if-shell` blocks, and bind-dispatched commands), the fix applies consistently everywhere a config line is run.

## Tests

`tests-rs/test_issue416_inline_comments.rs` covers the stripper directly (quotes, escapes, mid-word `#`, whole-line comment) and the end-to-end case that `set -g base-index 1   # test` now takes effect and matches the no-comment behaviour.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -598,10 +598,58 @@ pub(crate) fn is_known_command(app: &AppState, name: &str) -> bool {
     })
 }
 
+/// Strip a trailing `# comment` from a single config line.
+///
+/// In tmux a `#` begins a comment only when it is unquoted and appears at the
+/// start of the line or immediately after whitespace. A `#` inside single or
+/// double quotes (e.g. a `"#{...}"` format or `'#H'`), one escaped with a
+/// backslash, or one in the middle of a word (e.g. `colour#aabbcc`) is left
+/// untouched. This lets configs use trailing comments such as:
+///
+/// ```text
+/// set -g base-index 1   # start windows at 1
+/// ```
+///
+/// Issue #416.
+fn strip_inline_comment(line: &str) -> &str {
+    let bytes = line.as_bytes();
+    let mut in_single = false;
+    let mut in_double = false;
+    let mut prev_ws = true; // start of line behaves like a word boundary
+    let mut i = 0;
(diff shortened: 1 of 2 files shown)


psmux:
Thank you @fimilfaneea, this is a clean, well reasoned fix and the reasoning about tmux's tokeniser (quotes, escaped `\#`, mid word `#`) is exactly right.

I landed your commit on master via cherry pick so your authorship is preserved: f144ae6.

One small thing I had to correct while landing it: several raw string literals in `tests-rs/test_issue416_inline_comments.rs` used the `r#"..."#` delimiter around content that itself contained `"#` (for example `"#H"` and `"#{session_name}"`), which closes the raw string early and fails to compile. I bumped those to `r##"..."##` so the test file builds. I also added an end to end PowerShell test (`tests/test_issue416_inline_comments.ps1`) covering the config load, `source-file`, and TUI paths.

After landing: your 12 unit tests plus the new end to end and TUI checks all pass, and the full suite is green at 2191 tests.

Closing this in favour of the merged commit on master. Really appreciate the contribution.

PR #418, opened 2026-06-24
fix(attach): honour -t target instead of last_session (#408)

### LINKED ISSUE #408: attach-session ignores -t target, attaches to last_session instead (3.3.6)
## Summary

`attach-session -t <name>` ignores the requested target and instead attaches to whatever session is recorded in `~/.psmux/last_session`. The `-t` flag is honored by control commands (`display-message -t`, `list-panes -t`) but **not** by `attach-session`.

## Environment

- psmux 3.3.6 (Chocolatey package), reports `tmux 3.3.6`
- Windows 11, PowerShell 7

## Steps to reproduce

```
psmux new -d -s A
psmux new -d -s B        # last_session becomes B
psmux attach -t A        # -> attaches to B, not A
```

**Expected:** client attaches to session `A`.
**Actual:** client attaches to `B` (the value in `last_session`). `-t` is effectively ignored.

## Additional notes

- Reproduced from a clean shell with `$TMUX` empty (not a nesting issue, profile-independent).
- When the target is a "cold" session (no existing client), the failed attach loops: hundreds of TCP connections to that session's server port pile up in `TIME_WAIT` while the client stays attached to the `last_session` server.
- Suspected regression from the session-routing rework: 3.3.3 *"route switch-client to current session's server, not destination"*.

## Workaround

Since `attach` reads `~/.psmux/last_session` instead of honoring `-t`, you can write the desired session name into that file yourself and then attach without `-t`:

```powershell
Set-Content "$HOME\.psmux\last_session" "Main" -NoNewline
psmux attach
```

Confirmed working on 3.3.6. This exploits the same mechanism that causes the bug, so it is only a stopgap until `attach -t` is fixed.

fimilfaneea:
Fixes #408.

## Problem

`attach-session -t <name>` ignored the requested target and attached to whatever session was recorded in `~/.psmux/last_session`:

```
psmux new -d -s A
psmux new -d -s B        # last_session becomes B
psmux attach -t A        # -> attaches to B, not A
```

## Root cause

The global flag handling in `main()` strips `-t` **and its value** from the post-subcommand argv (`cmd_args`, and thus `sub_args`). The attach handler then looked for `-t` in those already-stripped args, never found it, and fell through to the `resolve_last_session_name_ns` fallback.

## Fix

When `-t` is present, use the target the global parser already resolved into `PSMUX_TARGET_SESSION`. That value comes from `parse_target`, so it correctly:

- extracts the session from a `session:window` form (`A:1` → `A`) — a raw argv value would keep the `:1` suffix and fail the `A:1.port` lookup,
- resolves `$id` and `=exact` targets, and
- carries the `-L` namespace prefix.

Both attach paths key the port-file lookup off `PSMUX_SESSION_NAME` (= this resolved name): remote/SSH at `client.rs` and local at `main.rs`. Positional (`attach NAME`), bare `attach` (→ last_session), and pane-only targets (`-t %2`, no explicit session → falls through) are unchanged.

## Tests

`tests-rs/test_issue408_attach_target.rs` replicates the resolution decision (same pattern as `test_issue202_switch_client.rs`): `-t` target, the `session:window` form, `-L` namespace, positional, bare, pane-only fall-through, and a regression test proving the old code returned `last_session` for `attach -t A`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -671,15 +671,20 @@ fn run_main() -> io::Result<()> {
                 // otherwise argv[0] (the exe path or subcommand name) gets
                 // picked up as the target session name.
                 let sub_args: Vec<&String> = cmd_args.iter().skip(1).copied().collect();
-                let name = sub_args
+                // `-t` and its value are stripped from cmd_args (and therefore
+                // sub_args) by the global flag handling above, so resolve the
+                // target from the original argv. Without this, `attach -t NAME`
+                // fell through to the last_session fallback and ignored the
+                // requested target (#408).
+                let name = args
                     .iter()
-                    .position(|a| *a == "-t")
-                    .and_then(|i| sub_args.get(i + 1))
+                    .position(|a| a == "-t")
+                    .and_then(|i| args.get(i + 1))
                     .map(|s| {
                         if let Some(ref l) = l_socket_name {
                             format!("{}__{}", l, s)
                         } else {
-                            (*s).clone()
+                            s.clone()
                         }
                     })
                     .or_else(|| {
--- a/src/main.rs
+++ b/src/main.rs
@@ -672,21 +672,19 @@ fn run_main() -> io::Result<()> {
                 // picked up as the target session name.
                 let sub_args: Vec<&String> = cmd_args.iter().skip(1).copied().collect();
                 // `-t` and its value are stripped from cmd_args (and therefore
-                // sub_args) by the global flag handling above, so resolve the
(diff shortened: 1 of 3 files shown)


psmux:
Thanks for this @fimilfaneea, and apologies for the duplication. We landed the same root cause fix in #438 (attach now reads -t from the full argv so it wins over last_session), together with a deterministic regression test that verifies via list-clients which session the client actually joins. Closing this as superseded, but your diagnosis in here was spot on and helped confirm it. Really appreciate the contribution.

PR #419, opened 2026-06-24
feat(choose-tree): kill highlighted window with 'x' (#410)

### LINKED ISSUE #410: psmux doesn't allow killing windows from `window selection` just whole sessions from `session selection`
`Ctrl+b+s` -> allows quick killing of a session from picker via `x`
`Ctrl+b+w` -> doesn't allow quick killing of a session from picker via `x`

In linux tmux does allow this from both.

No kill option psmux/window:

Image

Yes kill option pmsux/sessions:

Image

Tmux linux window selection allows killing (see reaction to typing `x`):

Image

fimilfaneea:
Fixes #410.

## Problem

The window picker (`choose-tree`) absorbed every non-navigation key, so there was no way to kill a window from it — only the **session** chooser had an `x` binding (which kills sessions). Users could navigate windows but not delete them.

## Fix

Add an `x` arm to the tree chooser, mirroring the existing session-chooser `x`: it focuses the highlighted window and then kills it (`focus-window <id>` + `kill-window`, since there is no kill-window-by-id control verb but `focus-window <id>` is already wired and `kill-window` targets the active window).

Scope and behaviour:
- Acts only on a **window row in the current session**. Session headers (`wid == usize::MAX`), pane rows, and rows belonging to other sessions fall through to the existing catch-all and are ignored (killing across sessions would require routing to another server — left for a follow-up).
- Honours the digit-jump buffer for target selection, same as `Enter`.
- Closes the picker after the kill.

## Testing

This lives in the client's inline key-event loop (no pure handler to unit-test, consistent with the rest of the chooser logic), so it is verified by compilation + manual use: open `choose-tree`, highlight a window, press `x`, and the window is removed. The change only adds a new match arm ahead of the existing no-op catch-all, so navigation and `Enter` behaviour are unchanged.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -3032,6 +3032,32 @@ pub fn run_remote(terminal: &mut Terminal<CrosstermBackend<crate::platform::Psmu
                                 KeyCode::Char('p') if tree_chooser => {
                                     preview_enabled = !preview_enabled;
                                 }
+                                // 'x' kills the highlighted window, mirroring the
+                                // session chooser's 'x' (issue #410). The window is
+                                // focused first so the kill targets it. Scoped to a
+                                // window row in the current session — session headers
+                                // (wid == usize::MAX), pane rows, and rows from other
+                                // sessions fall through to the catch-all below.
+                                KeyCode::Char('x') if tree_chooser => {
+                                    let sel_idx: Option<usize> = if tree_num_buffer.is_empty() {
+                                        Some(tree_selected)
+                                    } else {
+                                        match tree_num_buffer.parse::<usize>() {
+                                            Ok(n) if n >= 1 && n <= tree_entries.len() => Some(n - 1),
+                                            _ => None,
+                                        }
+                                    };
+                                    if let Some((is_win, wid, _pid, _label, sess_name)) =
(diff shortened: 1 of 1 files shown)


psmux:
Hi @fimilfaneea, thank you for this. Really nice work.

Your diagnosis is spot on: the tree picker only handled navigation, Enter, digits and `p`, then a catch all swallowed every other key including `x`, while the session picker already had its own `x`. Your approach is exactly right too: focus the window with `focus-window <id>`, then `kill-window` since kill targets the active window, scoped to a window row in the current session, honouring the digit jump buffer like Enter. That is precisely the shape this needed.

Timing was just unlucky. I had already reproduced #410 and landed an equivalent fix on master a little before I got to your PR, so the two touch the same lines and cannot both apply. The only behavioural difference is that the version on master keeps the picker open after a kill (it drops the killed row and keeps your cursor in place) so you can prune several windows in one visit, matching how the session picker stays open after its `x`. Your single kill then close behaviour was completely correct too, I just leaned toward parity with the session chooser.

Because it is already shipped I am going to close this one, but the credit is yours. You found the same root cause and the same clean fix independently, and it is genuinely appreciated. Please do keep them coming, and thanks again for taking the time to send a real patch rather than just a report.

PR #420, opened 2026-06-24
fix(detach): send DETACH so non-P detach-client actually detaches (#389)

### LINKED ISSUE #389: detach-client (without -P) does not detach: client immediately reconnects
## Summary

`psmux detach-client` (without `-P`) returns exit code `0` but does **not** actually detach an attached client. The targeted client is dropped from the server's registry, but it immediately **reconnects** (new tty, same process), so the detach has no observable effect. Only `detach-client -P` actually detaches — but `-P` also kills the client's parent process, which is not what a plain detach should do.

Real tmux's `detach-client` detaches cleanly: the client exits, the session persists, and there is no reconnect.

## Reproduction (psmux 3.3.6)

Using an isolated `-L` socket so nothing else interferes:

```sh
# terminal A — create a detached session on an isolated socket
psmux -L bugrepro new-session -d -s s

# terminal B — attach a client
psmux -L bugrepro attach -t s

# terminal A — try to detach every client of the session
psmux -L bugrepro detach-client -s s   # exits 0
psmux -L bugrepro list-clients
```

**Observed:** `list-clients` still shows an attached client for session `s`, but its tty has changed (e.g. `/dev/pts/2` → `/dev/pts/3`). The changed tty is the tell-tale sign that the old client tore down and a *new* connection took its place — i.e. the client reconnected rather than detached. The attached terminal B keeps rendering the session.

## Expected vs. actual

| | Expected (tmux parity) | Actual (psmux 3.3.6) |
|---|---|---|
| `detach-client` (no `-P`) | client exits, session persists, no reconnect | client reconnects (new tty), still attached |
| `detach-client -P` | client detaches | works, but also kills the client's parent process |

## tmux contrast (tmux 3.2a)

```sh
tmux -L t new-session -d -s s
tmux -L t attach -t s    # in another terminal
tmux -L t detach-client -s s
tmux -L t list-clients   # -> no clients; the attached terminal returns to the shell
```

tmux cleanly detaches the client; the session and its panes survive.

## Root cause

In `src/server/mod.rs`, the non-`-P` detach handlers force-close the client's TCP stream via `shutdown_client_stream(...)` **without** first sending a clean `"DETACH"` directive. Only the `kill_parent` (`-P`) branch sends `send_directive_to_client(cid, "DETACH-KILL-PARENT")`; the plain `"DETACH"` directive is otherwise sent only on session-end / kill paths.

The four affected handlers are:

- `CtrlReq::ForceDetachClient`
- `CtrlReq::ForceDetachClientByTty`
- `CtrlReq::DetachAllOtherClients`
- `CtrlReq::DetachAllClients`

In `src/client.rs`, a client sets `quit = true` (clean exit) only when it receives a `"DETACH"` / `"DETACH-KILL-PARENT"` directive. On a bare stream drop, if `!quit && port_file_exists`, it spawns a reconnect thread. So a non-`-P` detach is seen as a transient disconnect, and the client reconnects.

## Suggested fix

In each of the four handlers, on the non-`-P` path, send a clean `"DETACH"` directive to each targeted client **before** `shutdown_client_stream(...)` (mirroring the `-P` branch but using `"DETACH"` instead of `"DETACH-KILL-PARENT"`, and without killing the parent), with a brief sleep so the client can act on the directive before its stream is torn down. `-P` behavior stays unchanged.

fimilfaneea:
Fixes #389.

## Problem

`detach-client` without `-P` returns exit `0` but does **not** detach: the client is dropped from the registry, then immediately **reconnects** (new tty, same process), so it stays attached. Only `-P` actually detaches — but `-P` also kills the client's parent shell.

## Root cause

The non-`-P` detach handlers force-close the client's stream via `shutdown_client_stream(...)` **without** first sending a directive. The client treats a bare stream drop as a transient disconnect and spawns a reconnect thread (`!quit && port_file_exists`). It only sets a clean `quit` when it receives a `"DETACH"` / `"DETACH-KILL-PARENT"` directive — and the plain `"DETACH"` was only sent on session-end/kill paths, never on a user detach.

## Fix

In the four non-`-P` detach paths — `ForceDetachClient`, `ForceDetachClientByTty`, `DetachAllOtherClients`, `DetachAllClients` — send a clean `"DETACH"` directive before `shutdown_client_stream`, with the brief pause the `-P` paths already used so the client acts on it before the stream is torn down.

The client handles `"DETACH"` (client.rs) as a clean quit with **no** parent kill, so the session persists and there's no reconnect — matching tmux. `-P` keeps sending `DETACH-KILL-PARENT` and is unchanged.

## Testing

This is cross-process client/server behaviour over TCP (no pure handler to unit-test, consistent with the rest of the detach path), verified by code symmetry with the already-working `-P` path and the issue's repro:

```
psmux -L bugrepro new-session -d -s s
psmux -L bugrepro attach -t s        # terminal B
psmux -L bugrepro detach-client -s s # terminal B returns to shell; session survives
psmux -L bugrepro list-clients       # -> no clients
```

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -3730,6 +3730,14 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                     let _ = resp.send(output);
                 }
                 CtrlReq::ForceDetachClient(target_cid) => {
+                    // Send a clean DETACH so the client exits instead of treating
+                    // the dropped stream as a transient disconnect and reconnecting
+                    // (#389). The brief pause lets it act on the directive before we
+                    // tear the stream down. (For `-P` the connection handler already
+                    // sent DETACH-KILL-PARENT; this extra DETACH is harmless — the
+                    // client is already exiting and the kill-parent flag is set.)
+                    crate::types::send_directive_to_client(target_cid, "DETACH");
+                    std::thread::sleep(Duration::from_millis(50));
                     // Force-detach a specific client by shutting down its TCP stream
                     app.client_sizes.remove(&target_cid);
                     let was_present = app.client_registry.remove(&target_cid).is_some();
@@ -3773,10 +3781,14 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                         .find(|(_, ci)| ci.tty_name == tty)
                         .map(|(cid, _)| *cid);
                     if let Some(cid) = target_cid {
-                        if kill_parent {
-                            crate::types::send_directive_to_client(cid, "DETACH-KILL-PARENT");
-                            std::thread::sleep(Duration::from_millis(50));
-                        }
+                        // Send a clean detach directive before tearing down the
+                        // stream so the client exits instead of reconnecting (#389).
+                        // `-P` additionally kills the client's parent shell.
+                        crate::types::send_directive_to_client(
+                            cid,
+                            if kill_parent { "DETACH-KILL-PARENT" } else { "DETACH" },
+                        );
+                        std::thread::sleep(Duration::from_millis(50));
                         app.client_sizes.remove(&cid);
                         let was_present = app.client_registry.remove(&cid).is_some();
                         if was_present {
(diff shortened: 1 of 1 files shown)


psmux:
Hey, thank you for this, and for tying it straight to #389. Your diagnosis was spot on: without a clean directive the client treated the stream shutdown as a transient disconnect and reconnected, so the detach never stuck.

The reason I am closing this is timing, not the work. The same fix already landed on master, so it is resolved there. What shipped matches exactly what you found.

On current master all four detach paths send a clean directive before the stream is torn down:

* `ForceDetachClient` sends `DETACH`
* `ForceDetachClientByTty` sends `DETACH-KILL-PARENT` when `-P`, otherwise `DETACH`
* `DetachAllOtherClients` and `DetachAllClients` do the same, one directive per client

So the non `-P` `detach-client` path now sends `DETACH` instead of nothing, which is the heart of what you fixed. The shipped version also only waits the 50ms settle when a directive was actually queued (the send reports whether the client was still reachable), so it skips the pause when the client is already gone.

If you ever see a detach case that still reconnects on master, please reopen or open an issue with the repro and I will get on it. Genuinely appreciate you taking the time here.

PR #429, opened 2026-06-25
refactor: split psmux into lib + thin alias binaries

EOR-Effulgence:
## Summary

The `psmux`, `pmux`, and `tmux` bin targets all pointed at `src/main.rs`, so Cargo emitted:

> file `src/main.rs` found to be present in multiple build targets

This PR removes that warning by moving all program logic into a library crate and reducing each binary to a thin shim.

## Changes

- **`fix`**: `PSUEDOCONSOLE_INHERIT_CURSOR` is referenced only from a `#[cfg(test)]` regression test, so non-test builds flagged it as dead code. Annotated with `#[cfg_attr(not(test), allow(dead_code))]` — suppressed in non-test builds, still live/asserted under test.
- **`refactor`**: `src/main.rs` → `src/lib.rs` (`fn main` → `pub fn run`). Each binary is now a thin shim calling `psmux::run()`:
  - `src/main.rs` → `psmux`
  - `src/bin/pmux.rs` → `pmux`
  - `src/bin/tmux.rs` → `tmux`
- `Cargo.toml`: added `[lib]`, repointed the three `[[bin]]` targets to distinct paths.

The three binaries remain **identical aliases** — the invoked name is resolved at runtime via `current_exe()`, so behavior is unchanged. Packaging that expects all three exes (`target/release/{psmux,pmux,tmux}.exe`) is unaffected.

## Test plan

- `cargo build` — no "multiple build targets" warning; all three exes produced (identical size).
- `cargo check` — clean, no manifest warning.
- `cargo test --lib` — 2136 passed / 0 failed (incl. the `psuedocon` dead_code regression test and `readiness_tests`).
- `psmux -V` / `pmux -V` / `tmux -V` — all launch and report the same version.


psmux:
Thank you for this, and I am sorry it sat as long as it did, because the diagnosis is correct: the three bin targets do all point at `src/main.rs`, cargo does warn about it on every build, and the lib split is the textbook fix. I verified the warning is still present on current master today.

I am going to close this one without merging, and I want to be straight about why. The branch moves 4400 lines from `main.rs` into `lib.rs`, which means it conflicts with every change to `main.rs` that has landed since June, and it will conflict again with whatever lands next week; a rebase of a whole file move is effectively redoing it. The codebase is in a high churn phase right now (six PRs landed today alone), so a big structural move needs to happen at a quiet moment, freshly generated against that day's master, or it blocks everything else in flight.

When I do the split I will do exactly what you laid out here: `pub fn run()` in a lib crate, three thin shims, `current_exe()` name resolution untouched so the aliases stay byte identical, and your `cfg_attr` treatment for the `PSUEDOCONSOLE_INHERIT_CURSOR` test constant. You did the design work and I will credit you in the commit when it lands. Thank you for the careful test plan in the writeup, including checking all three exes still launch.

Issue #274, opened 2026-05-01
Server-side pipe wedge survives client kill — single hung pane locks entire session, requires reboot to recover


MarcinRybarczyk:
## Environment
- **psmux version:** 3.3.4 (windows-x64)
- **OS:** Windows 11 Enterprise, build 10.0.26200 (24H2)
- **Host terminal:** Windows Terminal 1.24.10921.0
- **Shell in pane:** PowerShell 7.6.1
- **Workload:** 3-pane tiled session, each pane runs `pwsh -NoExit -Command "claude --dangerously-skip-permissions"` (Anthropic Claude Code CLI — Node.js based)

## Summary
A single pane whose subprocess hangs on long-running stdout (foreground daemon — e.g. `node serve --port 5155`) eventually wedges the **server-to-client I/O pipe**. After the wedge, killing the `psmux attach` client process leaves the server, all panes, and all child processes alive per `psutil` — but a fresh `psmux attach -t <session>` re-attaches to the same wedged pipe and is also unresponsive. The only working recovery is rebooting the host machine.

This means a single misbehaving pane can take down the entire multi-pane session even though each pane should be I/O-isolated.

## Reproduction
1. Create a session with at least one pane: `psmux new-session -d -s repro -- pwsh -NoExit`
2. In that pane, spawn a long-running foreground process that produces continuous stdout: `node bin/sac.mjs serve --port 5155 2>&1` (any never-exiting daemon producing periodic output works — `npm run dev`, `python -m http.server`, `tail -f` on a busy log all reproduce)
3. Wait until the pane's parent process (in our case `claude.exe`) stops emitting events / responding to its own internal poll. Empirically: ~9.5 minutes after the daemon started in our case.
4. Try to focus the Windows Terminal tab containing the `psmux attach` client — keystrokes are not delivered, screen does not refresh.
5. From another pwsh process, kill **only** the `psmux attach` client process via `psutil.Process(<client_pid>).kill()`. Verify with `psmux ls` that the server still lists the session.
6. From a fresh Windows Terminal tab, run `psmux attach -t <session>` to re-attach.

**Expected:** The new client renders the current pane state and accepts input. The wedged pane may still show stale output, but other panes work; the wedged pane's subprocess can be inspected/killed normally.

**Actual:** The new `psmux attach` client also does not respond to input. Server (`psmux server`) still claims the session is alive. The only way to recover input was a full machine reboot.

## Diagnostic data captured pre-recovery
```
[...]

PR #435, opened 2026-07-01
fix(server): reap ghost clients by making ClientDetach idempotent + reaping from the writer Guard

### LINKED ISSUE #434: Ghost-client leak: a dead client's `ClientInfo` is never reaped, and `ClientDetach` over-decrements `attached_clients`
## Summary

On Windows, a long-lived session accumulates **ghost clients**: `list-clients` shows client
records whose underlying socket has already closed and whose process is long gone, but which are
never removed from the server's `client_registry`. Independently, the counter `attached_clients`
can be **over-decremented** so that `#{session_attached}` reads `0` while `list-clients` still shows
attached clients — a registry/counter desync.

Both stem from how a persistent attach connection is torn down.

## Environment

- **psmux version:** 3.3.6 (windows-x64)
- **OS:** Windows 11
- **Host terminal:** Windows Terminal
- Observed on a long-running session used by an agent CLI (many attach/detach + reconnect cycles).

## Observed

`list-clients` on a session that has been attached from a single terminal for a long time:

```
/dev/pts/73297: work: main [133x36] (utf8) [activity=362s ago]
/dev/pts/56735: work: main [133x36] (utf8) [activity=3482s ago]   <-- ghost: ~58 min stale
```

`list-sessions -F '#{session_name} attached=#{session_attached}'` shows `work attached=1`, and TCP
inspection of the session's server port shows exactly **one** `Established` client socket (plus many
`TIME_WAIT`). So the `pts/56735` record's socket is already **closed**, yet its `ClientInfo` was
never removed. Over time these accumulate; they clutter `list-clients` / the choosers and confuse
`attached`-based logic.

## Root cause

A persistent attach connection (`server/connection.rs`) is served by two threads:

- a **reader loop** that, on socket EOF/error **and** `attached_sent`, sends
  `CtrlReq::ClientDetach(client_id)` — the only *connection-teardown* path that reaps the
  `client_registry` entry and decrements `attached_clients` (explicit `detach-client` / force-detach
  commands aside);
- a **writer thread** whose `Guard::drop` fires on any writer exit (write timeout / broken pipe /
  `resp_rx` disconnect) and does `shutdown(Both)` + deregisters the frame/directive/persistent-stream
  channels, but **does not** enqueue `ClientDetach` — so it never touches `client_registry` /
  `attached_clients`.

So if a teardown is observed only by the writer path, the connection's `ClientInfo` is left
**orphaned = ghost**. Because the client auto-reconnects on EOF and each new connection gets a fresh
`client_id`, reconnect churn amplifies the symptom: the stale `client_id` lingers while the live one
carries on.

Separately, `CtrlReq::ClientDetach` (`server/mod.rs`) decrements `attached_clients`
**unconditionally**:

```rust
CtrlReq::ClientDetach(cid) => {
    app.attached_clients = app.attached_clients.saturating_sub(1); // <-- unconditional
    app.client_sizes.remove(&cid);
    app.client_registry.remove(&cid);
    ...
}
```

So a **duplicate** detach for one `cid` over-decrements the counter — it can reach `0` while real
client entries remain, producing the `attached=0`-while-`list-clients`-shows-clients desync.
(Contrast `ForceDetachClient`, which correctly guards with
`if app.client_registry.remove(&cid).is_some() { ..sub(1) }`.)

## Proposed fix

Two small, idiomatic changes (PR to follow):

1. Extract `AppState::reap_client(cid) -> bool` that removes the registry entry and updates the
   `attached_clients` / prefix / latest-client bookkeeping **only when the entry was actually
   present**. Idempotent — a second reap of the same `cid` is a safe no-op. Route the `ClientDetach`
   handler through it (side effects, incl. destroy-unattached, run only on a real reap).
2. Have the writer thread's `Guard::drop` also enqueue `ClientDetach(client_id)`, so a connection torn
   down via the writer path is reaped too (not just the reader-EOF path). This is low-risk because the
   Guard already `shutdown(Both)`s the socket (the client reconnects under a *new* `client_id`), so
   reaping the old `client_id` does not remove the live reconnected client.

Together these ensure a dead connection's `ClientInfo` is removed exactly once and `attached_clients`
tracks the registry.

With `destroy-unattached on`, a writer-detected teardown of the last client now follows the same
destroy path that a reader-detected EOF already used — intentional consistency; the option's default
remains off.

## Verification

- Release builds green on `x86_64` / `i686` / `aarch64-pc-windows-msvc`; `cargo test --all-targets`
  green (executed on x64). New unit tests assert `reap_client` idempotency and registry/counter
  consistency.
- Behavioral ConPTY smoke / no-regression harness (isolated `-L`/sandbox): attach two clients, kill
  one abnormally → the survivor stays counted (`attached=1`, no desync) and re-attach works (8/8
  invariants). The same harness passes on stock 3.3.6, so it is a regression/sanity check; the
  idempotent double-detach behavior is covered by the new unit tests, and the ghost leak by the live
  observation above.

## Related

- #274 (server-side state surviving a client kill) — this **partially addresses the client-state
  cleanup aspect** of that issue's *Suggested fix #2* by reaping stale client-registry entries on
  connection teardown. It does **not** claim to fix #274's deeper pane/client I/O wedge.
- #389 / #420 (plain `detach-client` reconnect) — **adjacent, separate**: this touches abnormal /
  writer teardown of `ClientDetach` (~`server/mod.rs:1606`), whereas #420 changes the *intentional*
  detach path (force-detach block, ~`server/mod.rs:3692+`) to send a clean `DETACH`. No hard conflict;
  the PR can rebase over #420 if it lands first.
- PR #375 discussion (client-death family / stale `(not responding)` entries in the choosers).

TrvoDK:
## What

Reaps **ghost clients** and fixes an `attached_clients` **over-decrement** desync. Closes #434.

A persistent attach connection is served by a reader loop (reaps the client on socket EOF) and a
writer thread whose `Guard` tears the connection down on any writer exit. Only the reader enqueued
`ClientDetach`, so when a teardown was observed only by the writer path the client's `ClientInfo` was
left **orphaned** and lingered in `list-clients` forever (a "ghost client"; reconnect churn amplifies
it since each new connection gets a fresh `client_id`). Separately, `ClientDetach` decremented
`attached_clients` unconditionally, so a duplicate detach could drop the count to `0` while real
clients remained (`#{session_attached}` reads `0` while `list-clients` shows clients).

## Changes

- **`AppState::reap_client(cid) -> bool`** (`src/types.rs`): idempotently remove a client from the
  registry and update `attached_clients` / `client_prefix_active` / `latest_client_id` **only when the
  entry was actually present** (`client_sizes` is always cleared). Returns whether a reap happened; a
  second call for the same `cid` is a safe no-op.
- **`CtrlReq::ClientDetach`** (`src/server/mod.rs`): route through `reap_client`; run the detach side
  effects (resize recompute, `client-detached` hook, destroy-unattached) only on a real reap, so a
  duplicate detach can't over-decrement the counter or fire destroy twice. Two intentional
  consequences of moving the bookkeeping inside the present-only guard (both benign): `client_prefix_active`
  is now cleared only on a real reap (a single global render-invalidation flag), and `latest_client_id`
  is no longer clobbered by a no-op detach — its value on a real reap is unchanged from before (`None`
  when the reaped client was the latest; this matches the prior `ClientDetach`, not the
  survivor-promotion that the force-detach handlers use).
- **Writer-thread `Guard::drop`** (`src/server/connection.rs`): also enqueue `ClientDetach(client_id)`,
  so a connection torn down via the writer path is reaped too, not just the reader-EOF path. The reader
  loop and this Guard may both enqueue it for the same `cid`; the idempotency above makes the pair reap
  exactly once. This is low-risk because the Guard already `shutdown(Both)`s the socket — the client
  reconnects under a *new* `client_id`, so reaping the old id does not remove the live reconnected
  client.
- **Tests** (`tests-rs/test_issue275_detach_client.rs`): 8 new cases exercising the real
  `reap_client` — first-reap removes + decrements, second-reap no-op, double-reap keeps the counter
  consistent with the registry (the `attached=0`-while-clients-remain regression), unknown-id no-op,
  latest-client clearing, prefix-flag clearing only on a real reap, destroy-unattached gated to a
  single real reap, and a **reader+writer duplicate-detach teardown** case asserting one registry
  removal + one decrement + side effects (hook/destroy) firing exactly once for the modelled
  connection-teardown path.

Scope note: only `ClientDetach` is refactored through `reap_client`; the explicit `detach-client` /
force-detach handlers already guarded their decrement correctly (they were the template
`ClientDetach` failed to follow), so they never had the over-decrement bug and are left unchanged to
keep this fix minimal.

On a clean detach the existing `client-detach` / EOF path reaps the client; any later Guard
`ClientDetach` for the same `cid` is an idempotent no-op, so hooks / counters / destroy-unattached do
not run twice. With `destroy-unattached on`, a writer-detected teardown of the last client now follows
the same destroy path a reader-detected EOF already used (intentional consistency; default off).

## Testing

- `cargo build --release` on `x86_64` / `i686` / `aarch64-pc-windows-msvc` — all green (CI).
- `cargo test --all-targets` — green (executed on the x64 job), including the 8 new tests + smoke.
- Behavioral ConPTY smoke / no-regression harness on the built binary: attach two clients → kill one
  abnormally → survivor stays counted (`attached=1`, no desync), re-attach works — 8/8 invariants,
  identical on stock 3.3.6 (a regression/sanity check). The idempotency + gating are covered by the new
  unit tests (incl. the reader+writer duplicate-detach case); the end-to-end leak is documented from the
  live observation in the issue. A full server-loop integration test of the writer-Guard reap is
  impractical (the missed-reap race is timing-dependent), so the invariant it relies on is pinned at the
  unit level instead.

## Related

- Partially addresses the client-state-cleanup aspect of #274's *second suggested fix* (release
  server-side client state on death) for the client registry — does **not** tackle #274's deeper
  pane-I/O wedge.
- Conceptually separate from #389 / #420 (intentional `detach-client` sending `DETACH`): that changes
  the *intentional* detach path (force-detach block, ~`server/mod.rs:3692+`); this touches
  `ClientDetach` (~`server/mod.rs:1606`) and abnormal/writer teardown. No hard conflict — happy to
  rebase over #420 if it lands first.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -282,21 +282,31 @@ if line.trim() == "PERSISTENT" {
         Ok(s) => s,
         Err(_) => return,
     };
+    // Clone the CtrlReq sender so the writer thread's Guard can reap this
+    // client from the server's registry when the connection tears down —
+    // regardless of whether the reader loop or the writer noticed the death.
+    let tx_guard = tx.clone();
     std::thread::spawn(move || {
         // Deregister the frame channel and shut down the TCP connection when
         // this thread exits for any reason (write timeout, resp_rx disconnect,
         // etc.). The shutdown causes the client's reader thread to see EOF,
         // which triggers reconnect rather than leaving the client frozen.
-        struct Guard { client_id: u64, shutdown: std::net::TcpStream }
+        struct Guard { client_id: u64, shutdown: std::net::TcpStream, tx: mpsc::Sender<CtrlReq> }
         impl Drop for Guard {
             fn drop(&mut self) {
                 let _ = self.shutdown.shutdown(std::net::Shutdown::Both);
                 crate::types::deregister_frame_channel(self.client_id);
                 crate::types::remove_directive_channel(self.client_id);
                 crate::types::deregister_persistent_stream(self.client_id);
+                // Reap the client-registry entry. The reader loop also sends
+                // this on EOF; ClientDetach is idempotent (it no-ops when the
+                // entry is already gone), so doing it here too guarantees a
+                // ClientInfo is never orphaned when the writer is the thread
+                // that observes the disconnect (the ghost-client leak).
+                let _ = self.tx.send(CtrlReq::ClientDetach(self.client_id));
             }
         }
-        let _guard = Guard { client_id, shutdown: ws_shutdown };
+        let _guard = Guard { client_id, shutdown: ws_shutdown, tx: tx_guard };
 
         loop {
             // Each iteration drains all three sources in priority order
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1604,33 +1604,33 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                     }
                 }
                 CtrlReq::ClientDetach(cid) => {
-                    app.attached_clients = app.attached_clients.saturating_sub(1);
-                    app.client_sizes.remove(&cid);
-                    app.client_registry.remove(&cid);
-                    app.client_prefix_active = false;
-                    if app.latest_client_id == Some(cid) {
-                        app.latest_client_id = None;
-                    }
-                    // Recompute effective size from remaining clients
-                    if let Some((w, h)) = compute_effective_client_size(&app) {
-                        app.last_window_area = Rect { x: 0, y: 0, width: w, height: h };
-                        resize_all_panes(&mut app);
-                    }
-                    hook_event = Some("client-detached");
-                    if app.attached_clients == 0 && app.destroy_unattached {
-                        let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
-                        let regpath = format!("{}\\.psmux\\{}.port", home, app.port_file_base());
-                        let keypath = format!("{}\\.psmux\\{}.key", home, app.port_file_base());
-                        let _ = std::fs::remove_file(&regpath);
-                        let _ = std::fs::remove_file(&keypath);
-                        crate::session::remove_session_id_file(&app.port_file_base());
-                        crate::types::shutdown_persistent_streams();
-                        tree::kill_all_children_batch(&mut app.windows);
-                        if let Some(mut wp) = app.warm_pane.take() {
-                            wp.child.kill().ok();
+                    // Idempotent reap (see AppState::reap_client): only run the
+                    // detach side effects when this client was actually present,
+                    // so a duplicate detach (reader EOF + writer Guard teardown,
+                    // or a detach-client racing the socket close) can't
+                    // over-decrement attached_clients or fire destroy twice.
+                    if app.reap_client(cid) {
+                        // Recompute effective size from remaining clients
+                        if let Some((w, h)) = compute_effective_client_size(&app) {
+                            app.last_window_area = Rect { x: 0, y: 0, width: w, height: h };
+                            resize_all_panes(&mut app);
                         }
-                        std::thread::sleep(std::time::Duration::from_millis(10));
-                        std::process::exit(0);
+                        hook_event = Some("client-detached");
+                        if app.attached_clients == 0 && app.destroy_unattached {
+                            let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
+                            let regpath = format!("{}\\.psmux\\{}.port", home, app.port_file_base());
+                            let keypath = format!("{}\\.psmux\\{}.key", home, app.port_file_base());
+                            let _ = std::fs::remove_file(&regpath);
+                            let _ = std::fs::remove_file(&keypath);
+                            crate::session::remove_session_id_file(&app.port_file_base());
+                            crate::types::shutdown_persistent_streams();
+                            tree::kill_all_children_batch(&mut app.windows);
+                            if let Some(mut wp) = app.warm_pane.take() {
+                                wp.child.kill().ok();
+                            }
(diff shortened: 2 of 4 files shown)


psmux:
This is a genuinely excellent patch @TrvoDK. The diagnosis is spot on and the code is exactly how I would have written it, down to cloning the `tx` sender into the writer `Guard` and gating the destroy unattached path behind a real reap.

I verified the whole thing independently before touching anything. The everyday paths never leak (plain kill, two clients, heavy attach/kill churn all reap cleanly), so I forced the writer only teardown by attaching a real client, suspending it so it stopped reading its socket, then flooding the pane to trip the writer write timeout. That plus a deterministic unit reproduction of the double detach over decrement confirmed both of your claims at the source level.

I have landed the fix on master in 7c41460 following your approach (`reap_client` idempotency plus reaping from the writer `Guard`), and credited you as co author on the commit. A couple of small additions on top of your PR:

1. Routed `ControlDeregister` through `reap_client` too, since it had the same unguarded decrement as `ClientDetach`.
2. While validating I found a separate unrelated race: `allocate_session_id` did a non atomic read modify write on the shared `.psmux/next_session_id` counter, so concurrent session creation could hand out duplicate `$N` ids (12 unique of 512 in a stress test). Now serialized with a process global mutex and a cross process lock file.
3. Added PowerShell E2E proofs (the suspend plus flood writer teardown scenario, reconnect, churn, and concurrent session id uniqueness) alongside the unit tests.

Because the same implementation is now on master I am closing this PR as landed, with full credit to you on the commit. Thank you for the sharp report and the clean fix, this is exactly the kind of contribution that makes the project better.

Issue #107, opened 2026-03-13
Claude Code Agent Teams issue


Josefvz:
> @Josefvz Great catch — thank you for the detailed report and screenshot!
> 
> I've identified and fixed both issues. The commit is here: https://github.com/marlocarlo/psmux/commit/fde3adf
> 
> ### What was broken
> 
> **1. The `-c` (start directory) flag was silently ignored**
> 
> When Claude Code runs `split-window -h -d -c <dir> -P -F "#{pane_id}"`, the `-c` argument was being parsed correctly but the value was discarded (`_start_dir`) before reaching the pane creation code. This means teammate panes always spawned in psmux's own working directory instead of the project directory Claude intended.
> 
> **Fixed:** `start_dir` is now wired through the full call chain — `app.rs`, `server/mod.rs`, `commands.rs`, `input.rs` — so both `split-window -c` and `new-window -c` correctly set the working directory of new panes.
> 
> **2. `build_command()` sent `/C` to bash instead of `-c`**
> 
> When psmux spawns a command in a pane, it needs to invoke the shell with the right flag. The code only detected PowerShell (`-Command`) and defaulted everything else to `cmd.exe /C`. So if your default shell is bash, the command `bash /C "echo hello"` would be sent — which bash doesn't understand.
> 
> **Fixed:** `build_command()` now does proper file-stem detection:
> - `pwsh` / `powershell` → `-Command`
> - `bash` / `sh` / `zsh` / `fish` / `dash` / `ash` → `-c`
> - fallback → `/C` (cmd.exe)
> 
> ### How to update
> 
> ```
> cargo install --git https://github.com/marlocarlo/psmux.git
> ```
> 
> Or if you have the repo cloned:
> ```
> git pull && cargo install --path .
> ```
> 
> ### Verified with tests
> 
> I've added a comprehensive test suite (`tests/test_claude_compat_fixes.ps1`) that covers the full Claude Code teammate workflow — 26 tests all passing:
> - `split-window -c` and `new-window -c` correctly set CWD
> - `split-window -h -d -c <dir> -P -F "#{pane_id}"` (exact Claude Code spawn pattern) 
> - Bash shell command dispatch using `-c` instead of `/C`
> - `send-keys` to split panes
> - Full end-to-end teammate spawn simulation
> 
> Let me know if this resolves it for you! 

 _Originally posted by @marlocarlo in [#23675](https://github.com/anthropics/claude-code/issues/23675#issuecomment-4049847073)_

[...]

PR #436, opened 2026-07-02
feat(warm-pane): honor -c on new-window/split via silent re-home

MattKotsenas:
## Problem

Binding `new-window -c '#{pane_current_path}'` (open the new window in the current pane's directory) is a common setup. But `new-window -c <dir>` and `split-window -c <dir>` skip the pre-spawned warm pane whenever a start dir is requested, cold-spawning a fresh shell every time. A cold spawn with a heavy shell profile is several seconds per pane, so the warm pane is effectively never used for the one binding that needs it most.

This is the #107 workaround: the fast path guarded on `start_dir.is_none()`, and four `CtrlReq` handlers stashed the warm pane aside whenever a start dir was set.

## Fix

Server claim already has a similar code path, where it pre-starts the server, then uses a "silent rehome" command to switch to the current directory (inject ` cd '<dir>'; cls` and squelch the pane's rendering until the resulting screen-clear (CSI 2J) arrives or 500ms elapses).

This change extracts that to its own function, wires it to `-c` for new windows and panes, then removes the four now-unnecessary stash/restore workarounds in the handlers. 

## Result

`new-window -c` and `split-window -c`: ~5,600 ms cold-spawn -> ~50 ms warm reuse and opened in the correct directory.

## Tests

- `tests-rs/test_warm_pane_start_dir.rs` - unit tests for `rehome_command`
  (single-quote escaping, platform clear command, trailing submit).
- `tests/test_warm_pane_start_dir.ps1` - isolated behavioral timing test; green
  on the fix, red on the pre-fix stash behavior.
- `tests/test_warm_pane.ps1` TEST 9 updated for the re-home behavior.

## Risks

This is injecting commands into the user's shell, which we were already doing before, but in a new code path.

Refs #107

### DIFF
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -71,6 +71,43 @@ fn default_shell_name(command: Option<&str>, configured_shell: Option<&str>) -> 
     }
 }
 
+/// Build the command injected to silently re-home a pane's shell to `dir`.
+///
+/// The trailing clear (`cls` on Windows, `clear` elsewhere) wipes the visible
+/// echo; single quotes in the path are doubled so the single-quoted string
+/// stays well-formed; the trailing `\r` submits it as one command line. The
+/// leading space asks shells that ignore space-prefixed commands to skip the
+/// history entry (best-effort — not every shell honours it).
+pub(crate) fn rehome_command(dir: &str) -> String {
+    let escaped = dir.replace('\'', "''");
+    let clear = if cfg!(windows) { "cls" } else { "clear" };
+    format!(" cd '{}'; {}\r", escaped, clear)
+}
+
+/// Silently re-home an already-running pane's shell to `dir`: inject
+/// [`rehome_command`] and squelch the pane's rendering until the resulting
+/// screen-clear (CSI 2J) fires or a 500ms safety timeout elapses, so the user
+/// never sees the injected command or its echo.
+///
+/// A pre-spawned shell (warm pane or warm server) starts in the server CWD;
+/// since a running process's CWD cannot be set externally, the shell moves
+/// itself with `cd`. The shell must be at a fresh prompt so the injected line
+/// runs immediately.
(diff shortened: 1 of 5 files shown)


psmux:
Matt, I want this one in, and I am asking for a rebase rather than resolving it myself, for a specific reason.

The value is clear to me: `new-window -c '#{pane_current_path}'` is probably the single most common binding, and it is exactly the one that never benefits from the warm pane today. Reusing the silent rehome that the claim path already trusts is the right mechanism, and extracting it into `pane::silent_rehome` removes duplication rather than adding a new injection path.

The reason I would rather you rebase than have me force the merge: since this branch was cut, #450 landed a liveness gate (`warm_pane_is_live`) in the exact create window fast path your change touches, plus #444 and #456 reshaped nearby regions of `server/mod.rs`. Your `command.is_none() && app.warm_pane.is_some()` condition needs to compose with the dead spare check from #450, and I do not want to guess at that composition on your behalf; a transplanted dead spare was a nasty bug and I would like your eyes on how the two gates combine.

If you rebase onto current master I will run your `test_warm_pane_start_dir.ps1` plus the warm pane suite and merge promptly. Thank you for the measurements in the writeup, the 5600ms to 50ms delta is exactly why this matters.

MattKotsenas:
@psmux updated!

psmux:
Merged. The proof was a pleasure to watch.

`new-window -c` reused the warm pane in 18ms and re homed the shell into the requested directory, versus 537ms on a cold spawn, and `split-window -c` behaves identically (5 of 5 green in `test_warm_pane_start_dir.ps1`). Your unit tests for `rehome_command` single quote escaping pass, and the composition with the #450 dead spare liveness gate holds, so `-c` never transplants a dead spare. Full suite green.

I like that you pulled `silent_rehome` into one function so the warm claim path and the new `-c` path share it. Nicely done, and thanks for the thorough tests.

Issue #129, opened 2026-03-18
There is dangling psmux/tmux process when no sessions are active


k3nsei:
After I exit all sessions there is still psmux/tmux proccess dangling in task manager.

I would expect to not have any psmux/tmux proccess running after all sessions are cancelled.

<img width="536" height="58" alt="Image" src="https://github.com/user-attachments/assets/f010c01e-1ccd-49d6-bd92-215c118caf20" />

<img width="902" height="260" alt="Image" src="https://github.com/user-attachments/assets/f630a167-0e0a-4f89-bb12-cee82910af1d" />

gtbuchanan:
Duplicate of #120

Issue #120, opened 2026-03-17
`__warm__` server remains after the last pane exits


spooki44:
## Summary
- If the one and only pane exits, `psmux.exe` remains running.
- In the repro, the remaining process is the `server -s __warm__ -x 120 -y 30` process.
- This reproduces whether the final pane is terminated via `send-keys` or `kill-pane`.
- This is not specific to detached sessions; it also reproduces with an attached session.
- `kill-server` does clean up all `psmux` processes, so this appears specific to last-pane/session shutdown.

## Environment
- OS: Windows 11
- Terminal: Windows Terminal
- Shell: PowerShell 7
- psmux: v3.2.0, tested against commit 0dbc242

## Expected
- Exiting the only pane in the only session should shut down `psmux` completely.
- After the pane exits, no `psmux.exe` processes should remain.

## Observed
- After exiting the only pane, a `psmux.exe` process remains running.
- By contrast, `psmux kill-server` leaves no `psmux` processes behind.
- Process behavior note: the first `psmux` invocation creates both `server -s 0` and `server -s __warm__ -x 120 -y 30`. Subsequent `new-session` calls create additional `server -s __warm__ -x 120 -y 30` processes, and when the final pane is terminated, those are cleaned up except for the one associated with the last session.

## Reproduction
1. Kill all existing `psmux` processes.
2. Start a session.
3. Terminate the one and only pane using either `send-keys` or `kill-pane`.
4. Inspect running `psmux.exe` processes and their command lines.

```text
Get-Process psmux | Stop-Process -Force
psmux new-session -d
# attached session also reproduces
psmux send-keys -t 0:. 'exit' Enter
# or: psmux kill-pane -t 0:.
Get-CimInstance Win32_Process | Where-Object { $_.Name -ieq 'psmux.exe' } | Select-Object ProcessId, ParentProcessId, CommandLine | Format-List
```

### Result
- A `psmux.exe` process remains after the only pane exits.
- Observed command line:

```text
"C:\Users\user\.cargo\bin\psmux.exe" server -s __warm__ -x 120 -y 30
```

## Impact
- `psmux` can be left running unexpectedly after the final pane exits.
- This makes shutdown behavior inconsistent with `kill-server`.
- Extra leftover processes can accumulate across repeated session creation and teardown.

## Workaround
- Run `psmux kill-server` to fully clean up remaining processes.

psmux:
[...]

PR #444, opened 2026-07-05
Fix warm-server spawn race that leaks __warm__ servers (#120)

andrew-tawfeek:
## Problem
`spawn_warm_server` checks `~/.psmux/<base>__warm__.port` and then spawns, with **no lock in between**. A freshly-spawned warm server takes a moment before it writes its port file, so two concurrent callers both see "no warm" and each spawn one. Only the last is tracked; the rest become **untracked orphans that are never reaped** — each holding a `pwsh` pane + `conhost`.

On a real Windows machine this reached **18+ `__warm__` servers** (28 servers / 62 pwsh / 93 conhost total), which exhausted desktop-heap/handles and tore down *every* terminal session at once — with no Application-error event, so it looked like random crashes. Related to #120 / #129.

## Fix
Serialize the check→spawn window with a `<base>.spawnlock` file (RAII guard). The lock is held off-thread until the newly-spawned warm registers its port, so a concurrent caller sees the genuine warm instead of racing another spawn. A lock older than 20s is treated as abandoned and stolen. No change on the happy path.

## Validation
Against a full release build, `tests/repro_warm_server_leak.ps1` now keeps warm servers **bounded at 1** across 4/8/12 rapid create-kill cycles (previously 18+), and `tests/test_issue120_warm_server_leak.ps1` passes.

---
*Authored by Claude (Anthropic) at @andrew-tawfeek's direction, after he hit this as a real crash-loop on Windows. Happy to adjust the approach (e.g. a named mutex or atomic self-election on boot) if you'd prefer.*

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -147,6 +147,48 @@ fn is_active_pane_squelched(app: &AppState) -> bool {
     } else { false }
 }
 
+/// RAII guard for the warm-spawn lock file. Removing the file on drop lets a
+/// later warm spawn proceed.
+struct WarmSpawnLock(std::path::PathBuf);
+impl Drop for WarmSpawnLock {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_file(&self.0);
+    }
+}
+
+/// Acquire the warm-spawn lock guarding the check->spawn window in
+/// `spawn_warm_server`. Returns `Some(guard)` when this caller owns the lock and
+/// may proceed to spawn, or `None` when another spawn is already in progress (in
+/// which case the caller must NOT spawn). A lock file older than 20s is treated
+/// as abandoned (its owner died mid-spawn) and stolen.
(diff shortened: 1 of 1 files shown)


psmux:
Merged. Andrew, this one was spectacular to reproduce.

Before merging I killed the warm pool and fired four concurrent `new-session` calls. Thirteen live `__warm__` servers appeared, each holding its own pwsh and conhost, and the startup reaper never touched them because each one is a live, running server that simply lost the port file race. That is your 18 server crash loop in miniature, confirmed end to end.

After merging and rebuilding, the same concurrent burst stays bounded: every spare beyond the single genuine warm is gone, and each new session either claims the real warm or cold spawns cleanly.

I also like the shape of the fix: `create_new` on the lock file is atomic on NTFS, the guard removes it on drop, the 20 second staleness steal covers a spawner that died mid flight, and holding the lock off thread until the port registers closes the exact window that caused the race. Thank you for chasing this from a production crash all the way to the root cause.

PR #445, opened 2026-07-05
Fix two Windows process-teardown leaks (inverted TerminateProcess, prune_exited orphan)

andrew-tawfeek:
Two small Windows process-teardown fixes found while auditing a real process leak.

**1. Inverted `TerminateProcess` return** — `crates/portable-pty-psmux/src/win/mod.rs`
Win32 `TerminateProcess` returns **nonzero on success**, zero on failure, but `do_kill` and `WinChildKiller::kill` returned `Err` on success / `Ok` on failure. `WinChild::kill` hides it with `.ok()`, but `WinChildKiller::kill` (from `clone_killer`) hands callers the inverted result. Fixed to `if res == 0 { Err } else { Ok }`.

**2. `prune_exited` orphans a self-exited shell's process tree** — `src/tree.rs`
When a pane's shell exits on its own (`try_wait()==Some`, `remain_on_exit=false`), the pane is dropped — only the ConPTY closes. Backgrounded grandchildren of that shell then leak, because `kill_process_tree` (which the explicit kill-pane/window/session paths already call) is never run on this path. Added the sweep.

## Validation
`tests/test_kill_tree.ps1` passes (4/4) against a full release build.

---
*Authored by Claude (Anthropic) at @andrew-tawfeek's direction, after debugging a Windows process-leak he hit in production.*

### DIFF
--- a/crates/portable-pty-psmux/src/win/mod.rs
+++ b/crates/portable-pty-psmux/src/win/mod.rs
@@ -42,7 +42,8 @@ impl WinChild {
         let proc = self.proc.lock().unwrap().try_clone().unwrap();
         let res = unsafe { TerminateProcess(proc.as_raw_handle() as _, 1) };
         let err = IoError::last_os_error();
-        if res != 0 {
+        // TerminateProcess returns nonzero on SUCCESS, zero on failure.
+        if res == 0 {
             Err(err)
         } else {
             Ok(())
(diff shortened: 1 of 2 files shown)


psmux:
Merged. Thank you Andrew, and please pass my thanks along to your debugging setup too, because both of these were real and I verified each one independently before merging.

On the inverted `TerminateProcess` return: I checked the call sites against the Win32 contract and you are exactly right, nonzero is success, so `WinChildKiller::kill` was handing callers `Err` on every successful termination. `WinChild::kill` masked it with `.ok()`, which is probably why it survived this long unnoticed.

On the `prune_exited` orphan: I reproduced it live before merging. I backgrounded a hidden marker process from a pane shell, typed `exit`, watched the pane get pruned, and the grandchild was still running afterwards. After merging your fix and rebuilding, the same experiment leaves zero survivors. That is exactly the kind of quiet leak that ends in the desktop heap exhaustion you hit in production.

Both fixes are in master now. Really appreciate the precision of this one: two small diffs, both provably correct.

PR #455, opened 2026-07-09
fix(rendering): stop rewriting box-drawing characters inside pane content at border junctions

### LINKED ISSUE #452: Line joints are inconsistent
Image

This is just an example of rendered markdown table, but it happens in any program that draws lines, like midnight commander and similar
In Windows terminal there is no such issue

LizardLiang:
Closes #452

## Problem

Box-drawing lines drawn *by programs running inside a pane* — a rendered
markdown table, midnight commander, anything that prints `│ ─ ┼` — came out
with inconsistent joints: the junction glyphs (`┼ ├ ┤ ┬ ┴`) were repainted
with the dim pane-border color while the `│`/`─` lines connecting them kept
the content's own colors.

## Root cause

After rendering, `run_remote` runs a recolor pass meant to fix inconsistent
styles at psmux's own separator intersections: it scanned the **entire
buffer** and restyled *every* junction glyph (`┼ ├ ┤ ┬ ┴`) with the
border/active-border style. It had no way to tell psmux-drawn junctions apart
from identical glyphs a child program printed as pane content, so a table's
joints got dimmed to the border color while its straight lines were left
alone.

`fix_border_intersections` — the pass that upgrades plain `│`/`─` separators
into junction glyphs where pane borders cross — had the same whole-buffer
character-sniffing problem: any content `│`/`─` with a perpendicular
neighbour could be rewritten into a junction glyph.

## Fix

Both passes are now restricted to a **geometry-derived border mask**: the
exact set of buffer cells where psmux itself drew a separator, computed from
the layout tree rather than by inspecting characters.

- `border_mask_from_layout` (`src/client.rs`) walks the `LayoutJson` tree and
  collects separator cell indices, mirroring `render_layout_json`'s geometry
  (`split_with_gaps`), including the zoomed-pane early-return so the mask
  matches what was actually drawn.
- `border_mask_from_node` (`src/rendering.rs`) does the same for the local
  `Node` tree, mirroring `render_node`'s `Node::Split` arm.
- `fix_border_intersections` now takes the mask and only walks masked cells;
  a junction glyph is emitted only when **both** the candidate cell and the
  crossing neighbour are in the mask. Content cells are never in the mask, so
  child-program box drawing is left byte-for-byte untouched.
- The active-border junction recolor pass in `run_remote` iterates the mask
  instead of the whole buffer, so it can no longer restyle content cells.

## Changes

| File | Change |
|------|--------|
| `src/rendering.rs` | `fix_border_intersections` takes a `sep_cells` mask; added `border_mask_from_node` |
| `src/client.rs` | Added `border_mask_from_layout`; `run_remote` builds the mask and passes it to both post-passes |
| `src/main.rs`, `src/preview.rs` | Build the mask at the remaining `fix_border_intersections` call sites |
| `src/commands.rs` | Register new test module |
| `tests-rs/test_pane_content_box_drawing.rs` | **New** regression test: renders a real split (genuine separator present), injects a box-drawing table into pane content, asserts glyphs *and* styles are untouched |
| `tests-rs/test_pane_border_lines_render.rs`, `tests-rs/test_pr255_active_border.rs` | Updated call sites for the new signature; existing junction/active-border expectations unchanged |

## Testing

- New regression test `fix_border_intersections_leaves_pane_content_table_untouched`.
- Existing border-line and active-border-color tests still pass with the mask
  in place, confirming real pane-border junctions are still upgraded and
  colored as before.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -575,6 +575,87 @@ fn collect_layout_borders(
     }
 }
 
+pub fn border_mask_from_layout(
+    node: &LayoutJson,
+    area: Rect,
+    buf_area: Rect,
+    zoomed: bool,
+) -> Vec<usize> {
+    let mut cells = Vec::new();
+    mark_layout_borders(node, area, buf_area, zoomed, &mut cells);
+    cells.sort_unstable();
+    cells.dedup();
+    cells
+}
+
+fn mark_layout_borders(
+    node: &LayoutJson,
+    area: Rect,
+    buf_area: Rect,
+    zoomed: bool,
+    cells: &mut Vec<usize>,
+) {
+    let LayoutJson::Split {
+        kind,
+        sizes,
+        children,
+    } = node
+    else {
+        return;
+    };
+    let effective_sizes: Vec<u16> = if sizes.len() == children.len() {
+        sizes.clone()
+    } else {
+        vec![(100 / children.len().max(1)) as u16; children.len()]
+    };
+    let is_horizontal = kind == "Horizontal";
+
+    if zoomed {
+        // Match render_layout_json's zoom early-return exactly: no separator
+        // drawn for this split, recurse into the chosen child at the SAME area.
+        if let Some(i) = effective_sizes.iter().position(|&s| s != 0) {
+            if let Some(child) = children.get(i) {
+                mark_layout_borders(child, area, buf_area, zoomed, cells);
+            }
+        }
+        return;
+    }
+
+    let rects = split_with_gaps(is_horizontal, &effective_sizes, area);
+    let w = buf_area.width as usize;
+    for i in 0..children.len().saturating_sub(1) {
+        if i >= rects.len() {
+            break;
+        }
+        if is_horizontal {
+            // Vertical separator line at column `pos`, spanning this split's own area height.
+            let pos = rects[i].x + rects[i].width;
+            if pos >= buf_area.x && pos < buf_area.x + buf_area.width {
+                for y in area.y..area.y + area.height {
+                    if y >= buf_area.y && y < buf_area.y + buf_area.height {
+                        cells.push((y - buf_area.y) as usize * w + (pos - buf_area.x) as usize);
+                    }
+                }
+            }
+        } else {
+            // Horizontal separator line at row `pos`, spanning this split's own area width.
+            let pos = rects[i].y + rects[i].height;
+            if pos >= buf_area.y && pos < buf_area.y + buf_area.height {
+                for x in area.x..area.x + area.width {
+                    if x >= buf_area.x && x < buf_area.x + buf_area.width {
+                        cells.push((pos - buf_area.y) as usize * w + (x - buf_area.x) as usize);
+                    }
+                }
+            }
+        }
+    }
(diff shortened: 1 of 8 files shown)


psmux (approved):
Thanks for this @LizardLiang, this is a really clean fix and the diagnosis is spot on.

I reviewed it end to end rather than just reading the diff, and it holds up.

## Verified the bug and the fix

I first reproduced #452 on master by hosting psmux under a real pseudoconsole and printing a bright green box drawing table inside a single pane, then reading the raw VT stream. The junction row came out as green lines with the junctions repainted to SGR 90 (the dim border color), exactly the mismatched joints from the issue:

```
before:  [92m───[90m┼[92m───[90m┤     joints dimmed
```

After building your branch and running the same capture, the joints are left alone:

```
after:   ├───┼───┤                    one color run, no [90m
```

## Checked for regressions on real borders

The worry with a change like this is that a real pane border stops getting its junction glyph or its color. It does not:

- Live 2 by 2 split under the pseudoconsole: 4 panes, vertical separators present, genuine crossing junctions still rendered and colored.
- The existing border and active border tests pass unchanged (`test_pane_border_lines_render`, `test_pr255_active_border`), so real junction upgrading and active border coloring are untouched.
- Full suite green here: 2271 passed, 0 failed.

## What I liked

The geometry derived mask is the right call, computing the separator cells from the same `split_with_gaps` the renderer uses rather than sniffing characters. And the zoom regression test is excellent: catching that `[100, 0]` becomes `[78, 1]` after min cell stealing and would otherwise mark a phantom separator column at x 78 is a subtle case that most fixes would have missed. Requiring both the cell and its crossing neighbour to be in the mask before emitting a junction is exactly right.

## Two minor notes (non blocking)

1. There are now two parallel geometry walkers, `mark_layout_borders` (LayoutJson path) and `mark_node_borders` (Node path), each mirroring a different renderer. That is inherent to the two render paths, but it means any future change to the separator geometry has to be kept in sync across four places now (two renderers, two masks). A short cross reference comment on each pointing at the other would help the next person.
2. `border_mask_from_node` does not take a `zoomed` flag the way `border_mask_from_layout` does. In practice this looks harmless because the Node path renderer (`render_window`) has no production callers I can find (the live client goes through `run_remote` and the LayoutJson path, which you handle correctly), so the Node mask is effectively test only. Not a blocker at all, just flagging that if `render_window` ever gets wired back up for a zoomable window, the Node mask would want the same zoom mirroring you already did on the layout side.

Approving. Nice work, and thanks for the thorough tests and writeup. I will merge shortly.

psmux:
Merged. Thank you @LizardLiang, this was excellent work.

I rebuilt on the merged master and reconfirmed everything before pushing: the #452 pseudoconsole repro flips from dimmed junctions to content colored, real 2 by 2 border junctions still render and color correctly, and the full suite is green at 2274 passing. Your commit went in with your authorship intact and you are credited as the author of the fix.

Special credit for the geometry mask approach and especially the zoom regression test. Catching that `[100, 0]` collapses to `[78, 1]` after min cell stealing, and guarding the phantom separator column, is the kind of edge case that usually only shows up as a bug report six months later. Really appreciated.

PR #456, opened 2026-07-09
fix(warm): skip client-facing hooks on the __warm__ server

MattKotsenas:
## Problem

The hidden `__warm__` pre-spawn server has no client and no visible session, but it runs the same `run_server` event loop as a real server, so it fires client-facing hooks. Two symptoms:

1. `status-interval` fires twice: a global `status-interval` hook runs once on the real server and once on the warm one.
2. Startup `client-attached`/`session-created` fire twice: `run_server` fires these once at startup "so plugins populate initial data", on the warm server too. A plugin whose `client-attached` hook spawns a process (e.g. psmux-continuum's auto-save loop) gets two loops, one per server, and double-saves.

## Fix

Gate the warm server out of client-facing hook work behind one predicate:

```rust
pub fn is_warm_server(&self) -> bool {
    self.session_name == "__warm__"
}
```

- **`status-interval` timer:** `should_run_status_interval_timer()` is `status_interval > 0 && !is_warm_server()`.
- **Startup hooks:** the startup `client-attached`/`session-created` fire is skipped when `is_warm_server()`.
- **Claim path:** because the warm server skips the startup fire, the claim path fires `client-attached`/`session-created` after promotion, so a claimed session gets them exactly once. It also clears `app.hooks` before reloading config, so the reload is authoritative: hooks removed from the config drop out, and `set-hook -a` append hooks don't stack a second copy.

When a warm server is claimed and renamed, `is_warm_server()` flips false and it behaves like any real server. `last_status_interval_fire` is reset on claim so the claimed session's first status fire lands one interval after creation, not immediately.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -962,11 +962,15 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
             Err(e) => { eprintln!("psmux: warm pane pre-spawn failed: {e}"); }
         }
     }
-    // Fire client-attached hooks once at startup so plugins populate initial
-    // data (e.g. CPU/battery) even for detached sessions (tppanel previews).
-    crate::commands::fire_hooks(&mut app, "client-attached");
-    // Fire session-created hook at startup
-    crate::commands::fire_hooks(&mut app, "session-created");
+    // Fire client-attached and session-created hooks once at startup so plugins
+    // populate initial data (e.g. CPU/battery) even for detached sessions
+    // (tppanel previews). Skip the warm server: firing here would double-fire
+    // every client/session hook (e.g. a duplicate continuum auto-save loop). A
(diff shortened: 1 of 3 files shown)


psmux:
Merged. Thanks Matt, confirmed and in.

Repro before merging: I pointed a config at a `session-created` hook that appends a line to a file, killed the warm pool, and created exactly one session. The file had two lines, one from the real server and one from the freshly spawned `__warm__` server loading the same config. After merging and rebuilding, the identical experiment writes exactly one line.

The claim path details are what sold me: firing `client-attached` and `session-created` after promotion so a claimed session still gets them exactly once, clearing `app.hooks` before the authoritative config reload so append hooks do not stack, and resetting `last_status_interval_fire` so the first status fire lands one interval after creation. All the corners were thought through. Plugins like continuum should stop double spawning their loops now.

PR #460, opened 2026-07-10
fix(commands): forward new-window args to control port

MattKotsenas:
`new-window` invoked via `execute_command_string` (display-menu items, hooks, confirm dialogs) sent a bare `new-window` to the control port, dropping `-n`, `-c`, and the shell-command. A menu action such as `new-window -n 'Foo' -c '/x' 'cmd'` therefore opened a default window instead of the intended one.

This mirrors the sibling `split-window` arm, which already forwards the full command string:

    "new-window" | "neww" => {
        if let Some(port) = app.control_port {
            let _ = send_control_to_port(port, &format!("{}\n", cmd), &app.session_key);
        }
    }

The control-port receiver already parses `new-window`'s `-n`/`-c`/positional arguments quote-aware (`server/connection.rs`), so no receiver change is needed.

Adds `new_window_forwards_full_command_to_control_port`, which stands up a control listener and asserts the full command (with quoted args) is forwarded.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -970,7 +970,7 @@ fn execute_command_string_single(app: &mut AppState, cmd: &str) -> io::Result<()
     match parts[0] {
         "new-window" | "neww" => {
             if let Some(port) = app.control_port {
-                let _ = send_control_to_port(port, "new-window\n", &app.session_key);
+                let _ = send_control_to_port(port, &format!("{}\n", cmd), &app.session_key);
             }
         }
         "split-window" | "splitw" | "split-pane" | "splitp" => {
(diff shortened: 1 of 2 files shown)


psmux:
Merged. Thanks Matt, this one was airtight.

I reproduced it before merging: I set `set-hook after-split-window "new-window -n FOOBAR460"` on a live session, triggered a split, and the hook created a window named `powershell` instead of `FOOBAR460`. The `-n` never made it to the control port, exactly as you described. After merging and rebuilding, the same hook produces a window named `FOOBAR460`.

The one line mirrors what the `split-window` arm already did, and your control port capture test locks it in so it cannot quietly regress. Appreciated as always.

PR #461, opened 2026-07-10
fix(session): honor -L namespace over $TMUX (tmux socket precedence)

MattKotsenas:
## Problem

When run from inside another session, `psmux -L X <cmd>` resolved its target session from `$TMUX` and routed to the current / `default` server, ignoring `-L X`. Confirmed with `psmux -L X list-windows` reading the default server, while `list-windows -a`, `list-panes -a`, and `kill-server` were correctly namespace-isolated.

Because the leak was in the central single-target resolver, it also affected `send-keys` and every other command that routes without an explicit `-t.

## tmux behavior

tmux's socket-selection precedence is `-S` > `-L` > `$TMUX` > `"default"`; it reads `$TMUX` only when neither `-L` nor `-S` is given (`tmux.c` `main()`: `if (path == NULL && label == NULL)`). In psmux the per-session port file is the "socket", optionally namespaced `X__session.port`, so honoring `-L X` means the current server named by `$TMUX` is adopted only when its session lives in the `X` namespace.

## Fix

Extract the inline `PSMUX_TARGET_SESSION` resolution out of `main()` into a pure `session::resolve_routing_target`, and make it:

- adopt the `$TMUX` session only when it matches the requested `-L` namespace (otherwise ignore `$TMUX` and resolve within the namespace);
- fall back to a namespaced `X__default` when nothing resolves under `-L X`, so a command never silently leaks to the un-namespaced `default` server (tmux would report "no server running on X").

Split into two commits: a behavior-preserving refactor (extract the pure fn) then the fix, so the "no behavior change" step is separable from the parity fix.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -281,46 +281,18 @@ fn run_main() -> io::Result<()> {
         }
     }
     if env::var("PSMUX_TARGET_SESSION").is_err() {
-        // No explicit session from -t: try to resolve from TMUX env var (set inside psmux panes)
-        // TMUX format: /tmp/psmux-<pid>/<socket_name>,<port>,<session_idx>
-        if let Ok(tmux_val) = env::var("TMUX") {
-            // Extract the port from the TMUX value
-            let parts: Vec<&str> = tmux_val.split(',').collect();
-            if parts.len() >= 2 {
-                if let Ok(port) = parts[1].trim().parse::<u16>() {
-                    // Look up which session owns this port (port file base
-                    // already includes -L namespace prefix if applicable)
-                    let home = env::var("USERPROFILE").or_else(|_| env::var("HOME")).unwrap_or_default();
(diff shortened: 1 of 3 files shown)


psmux:
Merged, with one tiny conflict I resolved on my side (both branches appended a test module registration at the tail of `session.rs`, so I kept both). Your commits land as authored.

I reproduced the leak before merging: with `$TMUX` pointing at a default namespace session, `psmux -L bogusns list-windows` happily listed that session's windows instead of failing, exactly the silent cross namespace routing you described. After the merge and rebuild, the same command resolves within the namespace and falls back to the namespaced default, and all six of your precedence tests pass.

Extracting the resolver into a pure `resolve_routing_target` is the part I appreciate most: that inline `$TMUX` block in `main()` was untestable, and now the precedence rule is written down in one place with tmux's `main()` as the reference. The split into refactor plus fix commits made the review easy. Thank you Matt.

PR #462, opened 2026-07-10
fix(input): send SS3 cursor keys in DECCKM application-cursor mode (tmux parity)

MattKotsenas:
## Problem

Cursor keys (arrows, Home, End) sent to a pane are always encoded as xterm CSI (`ESC [ A` and friends), regardless of the pane's cursor-key mode. Applications that enable DECCKM application-cursor-keys mode (`ESC [ ? 1 h`) expect SS3 (`ESC O A`) instead. PowerShell/PSReadLine reached over ssh into a psmux pane is the common case: it enables DECCKM, so psmux's CSI makes it echo the raw escape, and arrow-key history recall breaks, printing a literal `^[[A` instead of recalling the previous command.

## Fix

Consult the pane's vt100 `application_cursor()` state at the three sites that encode a named key, and emit SS3 for unmodified arrows/Home/End when the mode is set:

- interactive send-key (`write_named_key_to_pane`)
- the `send-keys` command handler
- the control-socket `CtrlReq::SendKeys` arms

Modified cursor keys (for example Shift+Up) stay xterm CSI, matching tmux's `input_key` under `MODE_KCURSOR`. The finals covered are `A B C D H F` (Up/Down/Right/Left/Home/End), the complete DECCKM set. F1-F4 already emit SS3 unconditionally.


psmux:
Merged. Beautiful work Matt.

I verified this at the byte level before merging. I put a raw reader inside a pane (console handle opened on `CONIN$` with only `ENABLE_VIRTUAL_TERMINAL_INPUT` set, reading via `ReadFile`), had it emit `ESC [ ? 1 h` so the pane entered DECCKM application cursor mode, then sent an Up arrow through `send-keys`:

Before the fix the child received `1b 5b 41`, the CSI form. After merging and rebuilding it receives `1b 4f 41`, SS3, which is what tmux emits under `MODE_KCURSOR` and what PSReadLine over SSH expects.

I also like that you kept modified cursor keys on the xterm CSI form and covered the full `A B C D H F` set, plus the vt100 timeline test that proves the parser state actually drives the encoder. This closes a genuine parity gap.

PR #463, opened 2026-07-10
fix(server): prevent duplicate same-name servers via an atomic session-creation guard

andrew-tawfeek:
## Problem

Two servers can end up running for the **same session name**. `new-session -s X`
cold-spawns a server after a non-atomic "does `X` already exist?" check, so if two
creators race — or `has-session` false-negatives under load — both spawn a
`server -s X`. They bind different ephemeral ports and race to write `X.port` /
`X.key`; the loser is a live orphan. The result is a desynced session that
"appears lost": blank `attach`, one-shot commands time out, and `kill-session`
can't reach the winner. It reproduces readily when several sessions/agents are
created concurrently.

The warm-claim path already guards against this (its code comment notes "a
slow/missing response must NOT trigger a cold spawn: doing so would create a
SECOND server with the same name"). The **cold-spawn** path had no equivalent
atomic guard.

## Fix

`run_server` acquires a Windows **named mutex** keyed on the session base name and
exits immediately if another live server already owns it.

- Named mutexes are auto-released by the OS on process death, so there is **no
  stale lock to reap** (unlike a lockfile), and an *abandoned* mutex — a prior
  owner that crashed — is correctly taken over (`WAIT_ABANDONED`).
- **Fail-open**: any FFI hiccup (null handle, unexpected wait result) yields a
  live guard, so a legitimate start is never blocked.
- **Warm/standby servers are exempt** (the warm pool intentionally runs several).

Adds `platform::acquire_session_mutex` + a `SessionMutex` RAII guard (no-op
fail-open stub on non-Windows).

## Test

`tests/test_no_duplicate_server.ps1` spawns a duplicate `server -s X` while the
first is alive and asserts the duplicate exits with exactly one survivor (red
before, green after). The existing session lifecycle behaviour is unaffected.

## Known limitation

A claimed-warm server later renamed to `X` does not hold `X`'s mutex, so a
warm-vs-cold race is still covered only by the existing warm-claim guard, not
this one. The common cold-vs-cold race is fixed.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -290,6 +290,65 @@ pub fn process_is_alive(_pid: u32) -> bool {
     true
 }
 
+/// Single-server-per-session-name lock (RAII). Holding the guard means this
+/// process owns the right to be THE server for a given session name. Dropping it
+/// (or the process exiting) releases the underlying Windows named mutex, which
+/// the OS also auto-releases on a crash — so there is no stale-lock to reap.
+#[cfg(windows)]
+pub struct SessionMutex { handle: *mut std::ffi::c_void }
+#[cfg(windows)]
+unsafe impl Send for SessionMutex {}
+#[cfg(windows)]
+impl Drop for SessionMutex {
+    fn drop(&mut self) {
+        #[link(name = "kernel32")]
+        extern "system" { fn CloseHandle(h: isize) -> i32; }
+        if !self.handle.is_null() { unsafe { CloseHandle(self.handle as isize); } }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


psmux:
Merged. Thank you.

I tested this the hard way before trusting it: spawned a real second `server -s <name>` while the first was alive and watched the duplicate exit while exactly one server survived (3 of 3 checks in `test_no_duplicate_server.ps1`). The named mutex is the right call on Windows, the fail open behavior means a legitimate start is never blocked, and exempting the warm pool is exactly right. The full suite of 2321 tests stays green on top of it.

Good, focused P0 fix for the duplicate same name server race. Appreciated.

PR #464, opened 2026-07-10
fix(ipc): make one-shot CLI commands reliable (kill-session/ls/attach no longer silently fail)

andrew-tawfeek:
## Problem

One-shot CLI commands over the loopback control socket could silently fail while
still reporting success:

- **`kill-session` no-op with exit 0.** `send_control` is fire-and-forget with no
  ack — the client closes the socket ~immediately after writing, so on Windows
  loopback an unread-data RST can make the server drop the command before it is
  dispatched. The handler then returned `Ok(())` regardless, so the session could
  survive while `$?` was `0`.
- **`ls` / `list-windows` print `os error 10060` or return empty.**
  `send_control_with_response` used a bare `connect()` with no timeout, so against
  a busy/backlogged server it hung until the ~21s Windows SYN-retransmit surfaced
  as WSAETIMEDOUT. And a read timeout was swallowed into `Ok("")`, so a merely
  slow server looked like "zero sessions/windows".
- **Destructive stale-port cleanup on timeout.** `kill-session` (and the
  `list-sessions`/`list-windows -a` sweeps) deleted a session's `.port` file
  whenever the connection failed — including a mere timeout — orphaning a
  live-but-busy server.
- **`attach` hangs → `10060`.** The persistent client path used the same bare
  `connect()` with no timeout.

## Fix (client-side; no wire-protocol change)

- `send_control`: **half-close** the write side after sending so the server sees
  EOF *after* our bytes (TCP delivers all data before the FIN), eliminating the
  RST-drop; `connect_timeout` so a busy server isn't mistaken for a dead one.
- `send_control_with_response`: `connect_timeout` (kills the 21s→10060 hang) and
  return a retryable **error on an empty timeout** instead of `Ok("")`.
- `kill-session`: **send-then-verify** with retry until the session is confirmed
  gone (5s deadline), exit non-zero if it survives, and only delete a `.port`
  file on an active `ConnectionRefused` — never on a timeout. Same
  timeout-is-not-dead fix for the `list-sessions` / `list-windows -a` sweeps.
- `attach` (`establish_connection`): `connect_timeout(2s)` + a bounded AUTH read,
  so attach fails fast and clearly instead of hanging, and reconnect backs off
  promptly.

## Test

`tests/test_command_reliability.ps1` (isolated `USERPROFILE`) asserts `ls` /
`list-windows` never emit `10060`, `has-session` exit codes, and — the core one —
that `kill-session` exits 0 **and the session is actually gone afterwards**, plus
a window-create burst that tears down cleanly.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1286,13 +1286,23 @@ struct ClientDragState {
 /// thread, and return a (writer, frame_rx) pair ready for the event loop.
 /// Sets a 5-second write timeout so blocked writes never freeze the client.
 fn establish_connection(addr: &str, key: &str) -> io::Result<Connection> {
-    let stream = std::net::TcpStream::connect(addr)?;
+    // Bounded connect: a wedged/backlogged server otherwise makes `attach` hang
+    // until the ~21s Windows SYN-retransmit and surface as `os error 10060`.
+    // A 2s timeout turns that into a fast, clear failure the caller can report
+    // (and lets try_reconnect back off promptly instead of stalling).
+    let sockaddr: std::net::SocketAddr = addr.parse()
+        .map_err(|_| io::Error::new(io::ErrorKind::Other, format!("bad server address: {addr}")))?;
+    let stream = std::net::TcpStream::connect_timeout(&sockaddr, Duration::from_secs(2))?;
     stream.set_nodelay(true)?;
     let mut writer = stream.try_clone()?;
     writer.set_nodelay(true)?;
     writer.set_write_timeout(Some(Duration::from_secs(5)))?;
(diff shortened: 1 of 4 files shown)


psmux:
Merged. This one I hammered.

14 of 14 checks in `test_command_reliability.ps1` pass, including the core one that `kill-session` exits 0 AND the session is actually gone afterward, plus a burst of five rapid new windows that tears down cleanly with zero `os error 10060`. The half close before EOF and the send then verify retry are solid, and refusing to treat a mere timeout as a dead server (so a busy but live server never loses its port file) is the subtle part you got right.

Client side only, no wire change, high value. Great work.

PR #465, opened 2026-07-10
perf(server): take warm-pane replenishment off the command path

andrew-tawfeek:
## Problem

The single server event loop serves every command **and** spawns panes.
`new-window` / `split-window` serve instantly by transplanting a pre-spawned
"warm" pane, then **synchronously spawn its replacement** (a fresh shell +
ConPTY) right there on the loop. A spawn can block the loop for 100 ms–seconds
under load, so when several windows are created close together the blocking
spawns chain up and stall *other* clients' commands (visible as intermittent
command latency / timeouts).

## Fix

Move warm-pane replenishment **off the command path**: drop the four inline
post-command spawns and instead replenish at the loop top only during an idle gap
(no command processed in the last 20 ms). Window-create bursts now transplant the
ready pane instantly, and the blocking spawn lands in idle time. Correctness is
unchanged — if no warm pane is ready when `new-window` arrives, `create_window`
still spawns one synchronously as before; this only moves the *replacement* spawn
out of the hot path.

## Limitation

This reduces, but does not fully eliminate, spawn-induced stalls: a sustained
burst can still outrun a single warm pane. A small warm-pane *pool* would remove
the remainder; left as a follow-up.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1089,6 +1089,20 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
     let mut temp_focus_restore: Option<(usize, usize)> = None;
 
     loop {
+        // Tier 3 — keep warm-pane spawning OFF the command path. A warm pane is a
+        // fresh shell + ConPTY; spawning one can block the single event loop for
+        // 100ms–seconds under load. Doing it inline after every new-window/split
+        // stalled *other* clients' commands, surfacing as the intermittent
+        // `os error 10060` timeouts. Instead replenish only during a quiet gap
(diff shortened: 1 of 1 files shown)


psmux:
Merged, cleanly on top of #436.

Verified the win: the five window burst in `test_command_reliability` creates every window and tears down with no timeouts, and the warm pane still serves instant windows (18ms in the `-c` test) because the replacement spawn now lands in the idle gap instead of blocking the event loop after every new window. Moving replenishment to the loop top during a quiet gap is a clean fix, and correctness stays intact since `create_window` still spawns synchronously when no warm pane is ready.

The combined build passes the full 2321 test suite. Thank you.

PR #466, opened 2026-07-10
fix(ipc): confirm command execution via a FIFO session-info barrier (builds on #464)

andrew-tawfeek:
> **Note:** builds on #464 (Tier 1). The net-new change here is the final commit,
> "Tier 2 — confirm command EXECUTION via a FIFO barrier". Review after #464.

## Idea

Tier 1 (#464) guarantees a fire-and-forget command is *delivered* to the server;
it does not guarantee the command has *finished executing* before the CLI exits,
so a script that inspects the effect immediately after can still race.

`send_control` now appends a `session-info` request after its command.
`session-info` round-trips through the server's single FIFO event loop
(`CtrlReq::SessionInfo` → blocking reply), so receiving its answer proves every
command dispatched earlier on the connection has actually been applied.
`send_control` thus becomes synchronous — it returns only once the command has
executed. Commands whose handler tears down the connection first (e.g.
`kill-session`) simply leave the barrier unanswered; that path is covered by the
caller's verify-retry from #464.

Also hardens the server: the `session-info` handler waited on an unbounded
`rrx.recv()`; since it now doubles as the execution barrier, it uses
`recv_timeout(3s)` so a momentarily-wedged event loop can't leak the connection
thread.

## Tradeoff (please weigh)

This adds one event-loop round-trip to **every** mutating one-shot command. That
makes scripting race-free, but under a heavy command stream it is extra
event-loop traffic. If that default is undesirable upstream, an alternative is to
gate the barrier behind a flag (opt-in for callers that need
inspect-after-mutate ordering) — happy to adjust.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1286,13 +1286,23 @@ struct ClientDragState {
 /// thread, and return a (writer, frame_rx) pair ready for the event loop.
 /// Sets a 5-second write timeout so blocked writes never freeze the client.
 fn establish_connection(addr: &str, key: &str) -> io::Result<Connection> {
-    let stream = std::net::TcpStream::connect(addr)?;
+    // Bounded connect: a wedged/backlogged server otherwise makes `attach` hang
+    // until the ~21s Windows SYN-retransmit and surface as `os error 10060`.
+    // A 2s timeout turns that into a fast, clear failure the caller can report
+    // (and lets try_reconnect back off promptly instead of stalling).
+    let sockaddr: std::net::SocketAddr = addr.parse()
+        .map_err(|_| io::Error::new(io::ErrorKind::Other, format!("bad server address: {addr}")))?;
+    let stream = std::net::TcpStream::connect_timeout(&sockaddr, Duration::from_secs(2))?;
     stream.set_nodelay(true)?;
     let mut writer = stream.try_clone()?;
     writer.set_nodelay(true)?;
     writer.set_write_timeout(Some(Duration::from_secs(5)))?;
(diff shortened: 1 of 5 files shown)


psmux:
Holding this one for now, and I want to be transparent about why.

#464 (Tier 1) is merged, so the base of this is in. Two things remain before I take the barrier:

1. The Windows x64 test job on this PR is currently red. I want that green first.
2. This appends a `session-info` round trip to every mutating one shot command, and you flagged that tradeoff yourself in the description. I lean toward gating the barrier behind an opt in flag so the default command path stays cheap, and callers that need inspect after mutate ordering can ask for it explicitly.

Could you rebase onto current master (now that #464 landed), get the x64 job passing, and switch the barrier to opt in? Then I will re review. The idea is sound, I just do not want to tax every command by default. Thanks for splitting Tier 1 and Tier 2 the way you did, it made the first half easy to accept.

psmux:
Merged, thank you @andrew-tawfeek! Since you flagged the round trip tradeoff yourself, I measured it directly before merging: `set-option` went from 15.5ms to 15.9ms at p50 on loopback, so the barrier costs well under a millisecond in practice. I also confirmed the barrier line never leaks into user data (`set-buffer` content stays exact and nothing stray reaches panes via `send-keys`), `kill-session` completes in about 250ms including the verify retry, the namespace path with `-L` tears down correctly, and your reliability suite passed 14 of 14 on the merged build with the full Rust suite at 2450 passing. Making one shot commands provably synchronous closes a whole class of scripting races. Great follow through across both tiers.

PR #467, opened 2026-07-10
fix(config): strip #branch suffix before deriving @plugin path

MattKotsenas:
The native `@plugin` auto-source derives the plugin directory from the spec with `value.rsplit('/').next()`. Branch names can contain `/` (e.g. `temp/integration`), so a spec like `owner/psmux-plugins/psmux-continuum#temp/integration` resolves the plugin name to `integration`, and no `plugin.conf` is ever sourced — silently breaking themes, settings, and continuum's auto-restore hooks after a reboot.

This strips an optional `#branch` suffix into a base value before deriving both the plugin name and the candidate paths, so a `#branch` pin (used to dogfood a fork branch) no longer defeats plugin discovery.

Regression test: `plugin_discovery_strips_branch_suffix`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1164,18 +1164,23 @@ pub fn parse_option_value(app: &mut AppState, rest: &str, _is_global: bool) {
             // Tries:  ~/.psmux/plugins/<full-value>/plugin.conf
             //   then: ~/.psmux/plugins/<last-component>/plugin.conf
             if key == "@plugin" && !value.is_empty() {
-                let plugin_name = value.rsplit('/').next().unwrap_or(value);
+                // Strip an optional '#branch' suffix before deriving the plugin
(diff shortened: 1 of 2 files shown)


psmux:
Merged. Thank you Matt.

I reproduced the exact failure first before touching the code: an `@plugin` spec pinned to a branch whose name contains a slash, like `owner/psmux-plugins/psmux-brtest#temp/integration`, and confirmed the branch tail was shadowing the plugin name so `plugin.conf` was never sourced. After your change I ran a real config pipeline check (isolated home, a plugin dir with a `plugin.conf` that sets an option, a session started with that config) and the option landed correctly. Your regression `plugin_discovery_strips_branch_suffix` passes and the full suite stays green.

Cleanly scoped, and it only changes behavior for specs that actually contain a `#`, which were broken before anyway. Nice catch on the fork branch dogfooding path.

PR #468, opened 2026-07-11
fix: detect WezTerm for VT mouse input so SGR sequences don't leak into panes

geneccx:
## Problem
Under WezTerm on Windows, moving or clicking the mouse prints raw SGR mouse sequences (e.g. `␛[<35;79;36M`) as literal text into the active pane instead of being handled as mouse input. `set -g mouse off` does **not** suppress it, because this is an input-decoding path, not the mouse option.

## Root cause
`needs_vt_input()` only routes input through the VT mouse parser for SSH sessions or JediTerm (`TERMINAL_EMULATOR` contains `"JetBrains"`). WezTerm is a ConPTY-based VT terminal that — exactly like JediTerm — writes VT mouse escape sequences to the ConPTY input pipe. ConPTY does not translate these into `MOUSE_EVENT` records, so the Win32 `ReadConsoleInputW` reader never sees them as mouse events and the raw bytes leak through as `KEY_EVENT` text into the pane. This is the same failure mode the JediTerm branch already exists to handle — WezTerm was simply never added.

## Fix
Detect WezTerm via the environment variables it always sets (`WEZTERM_PANE`, and `TERM_PROGRAM=WezTerm`) and route it through the same VT input path. `TERMINAL_EMULATOR` is referenced only in `needs_vt_input()`, so the change is self-contained.

## Testing
Built from `master` and run under WezTerm on Windows 11 (pwsh). Before: mouse motion/clicks flood the active pane with `␛[<..M` text. After: no leakage, and mouse (wheel-scroll, click-to-focus) works correctly.

### DIFF
--- a/src/ssh_input.rs
+++ b/src/ssh_input.rs
@@ -223,6 +223,14 @@ pub fn needs_vt_input() -> bool {
     is_ssh_session()
         || std::env::var("TERMINAL_EMULATOR")
             .map_or(false, |v| v.contains("JetBrains"))
+        // WezTerm on Windows is a ConPTY-based VT terminal that writes VT mouse
+        // escape sequences to the ConPTY input pipe, exactly like JediTerm.
+        // ConPTY does not translate these into MOUSE_EVENT records, so without
+        // the VT input parser the raw SGR bytes (e.g. "\x1b[<35;..M") leak
+        // through as KEY_EVENT text into the active pane. Detect WezTerm via the
+        // env vars it always sets and route it through the VT input path.
+        || std::env::var("TERM_PROGRAM").map_or(false, |v| v == "WezTerm")
+        || std::env::var_os("WEZTERM_PANE").is_some()
 }
 
 /// Returns the Windows build number (e.g. 19045 for Win10 22H2, 22631 for


psmux:
Merged. Thank you Gene, this is a clean and well reasoned fix.

I reproduced the leak myself and your root cause analysis holds up exactly. On master, `needs_vt_input()` only routed SSH and JediTerm through the VT parser, so WezTerm fell onto the native `ReadConsoleInputW` path and the raw SGR mouse bytes leaked into the active pane as text. Your point that WezTerm behaves just like JediTerm here (it writes VT mouse sequences straight into the ConPTY input pipe, which ConPTY never turns into `MOUSE_EVENT` records) is spot on, and `set -g mouse off` cannot touch it because this is an input decode path, not the mouse option.

What I checked before and after:

**Routing decision.** I wrote unit tests directly on `needs_vt_input()`, which is the whole fix. On stock master the three WezTerm cases fail (the function returns false), and with your patch all of them pass. I also added guards so an exact `TERM_PROGRAM == "WezTerm"` match is required, which keeps vscode, tmux and Apple_Terminal on the native path so there are no false positives, plus checks that the existing SSH and JediTerm routing stays untouched.

**Full suite.** 2339 passed, 0 failed. Zero regressions.

**End to end on a real Windows 11 build.** The normal local terminal case is unaffected, and a live psmux window launched with the WezTerm environment variables boots a fully functional session on the forced VT input path with no crash or hang. Once routed, the SGR bytes flow into the same tested decoder SSH and JediTerm already use, so wheel scroll and click land as real mouse events instead of leaking.

I landed a small set of regression tests alongside the merge so this stays guarded going forward.

One tiny housekeeping note: I squashed the commit with a slightly reworded message and dropped the extra co author trailer to match the project history style. You remain the author of the change. Really appreciate the careful writeup and the fix.

geneccx:
Appreciate the quick turnaround and diligence with adding tests!

Issue #299, opened 2026-05-18
`pane_current_command` diverges from tmux and blocks reliable session restore


MattKotsenas:
`pane_current_command` appears to give different answers from tmux for the same scenario.

### 1. Idle shells report as `shell`, not the shell binary name

Reproducer (PowerShell):
```powershell
psmux new-session -d -s repro
# Sit at an idle pwsh prompt — no foreground program running.
psmux list-panes -t 'repro:1' -F '#{pane_current_command}'
```

Observed: `shell`
Expected (matching tmux): `pwsh`

The literal string `"shell"` comes from `src/format.rs`:

```rust
"pane_current_command" => {
    if let Some(p) = target_pane() {
        if let Some(pid) = p.child_pid {
            crate::platform::process_info::get_foreground_process_name(pid)
                .unwrap_or_else(|| "shell".into())
        }
        ...
    }
}
```

When `get_foreground_process_name` returns `None` (no descendants found), the format layer substitutes the placeholder string `"shell"` rather than looking up the pane shell PID's own executable basename.

For comparison, tmux's `osdep-linux.c`:

```c
char *osdep_get_name(int fd, __unused char *tty) {
    pid_t pgrp;
    if ((pgrp = tcgetpgrp(fd)) == -1) return NULL;
    xasprintf(&path, "/proc/%lld/cmdline", (long long) pgrp);
    /* reads /proc/<pgrp>/cmdline */
}
```

For an idle pwsh, pwsh is the only process attached to the TTY, so `tcgetpgrp` returns pwsh's pgrp, `/proc/<pid>/cmdline` returns `pwsh`, and `pane_current_command` is `pwsh`. tmux never has to substitute a placeholder.

### 2. Active programs report a deep descendant, not the foreground process

Reproducer:
```powershell
psmux new-session -d -s repro
# Launch a long-running node child (clean process tree)
psmux send-keys -t 'repro:1' 'node -e "setInterval(()=>{},1000)"' Enter
Start-Sleep -Seconds 6
psmux list-panes -t 'repro:1' -F '#{pane_current_command}'
```

Observed for `node`: `node` ✓ (works in this clean case)

Now try a program that spawns a subprocess tree:
```powershell
psmux send-keys -t 'repro:1' 'copilot' Enter
Start-Sleep -Seconds 10
psmux list-panes -t 'repro:1' -F '#{pane_current_command}'
```

[...]

PR #469, opened 2026-07-12
feat: derive pane_current_command from shell-integration OSC sequences

MattKotsenas:
Follow-up to #299, adding the OSC shell-integration layers (1-3) of the `pane_current_command` cascade.

## What this adds

When a shell emits a shell-integration OSC sequence naming the command it is about to run, psmux treats that as the authoritative source for `#{pane_current_command}`, and falls back to the existing process-tree heuristic when no such signal is present.

That keeps the value stable for wrappers, SSH, shell functions, and fast-exiting children, where the process tree drifts from what the user actually launched.

Recognized sequences (all optional):

- **OSC 133;C** with `cmdline_url=<percent-encoded>` (kitty-style command reporting)
- **OSC 1337;SetUserVar** with `WEZTERM_PROG=<base64>` (iTerm2 / WezTerm)
- **OSC 633;E** (VS Code shell integration)
- **OSC 133;A** / **633;A** clear the command at prompt start, and **133;D** /
  **633;D** clear it on completion, so idle panes report the shell rather than a
  stale command.

## How it works

- `crates/vt100-psmux`: the OSC parser (`perform.rs`) records the announced command on `Screen` (`screen.rs`). The percent and base64 decoders are inlined so vt100-psmux stays dependency-free.
- `src/format.rs`: `pane_current_command` prefers the parsed OSC command, then falls through to the process-tree logic.

No config, and no behavior change for shells that do not emit these sequences.


psmux:
Merged. Thank you Matt, this is a really well built feature and the writeup made review a pleasure.

I did not take the shell-integration claim on trust; I proved it end to end through the actual runtime path rather than only the parser.

**Parser, in isolation.** Your 25 tests pass. I then added an independent safety suite that hammers the two hand rolled decoders with adversarial pane output, since this code consumes untrusted bytes from any child process: lone and truncated percent escapes (`%`, `%A`, `%ZZ`), NUL and high bytes, illegal base64, odd padding, single leftover base64 chars, 16 KB payloads, byte at a time feeding of a nasty payload, and interleaving with printed text. No panics, no hangs, invalid values are dropped rather than mangled, valid values still decode, and crucially none of the OSC framing ever leaks into the visible grid. Exactly the properties you want from a parser on this path.

**End to end at runtime.** I had a real pane child process write the actual OSC bytes to its stdout, so the bytes travel ConPTY to the pane vt100 parser to `#{pane_current_command}` through your format.rs cascade. All three sources resolve correctly on a live session:

- `OSC 133;C;cmdline_url=copilot%20--yolo` resolves to `copilot --yolo` (percent decoded)
- `OSC 1337;SetUserVar=WEZTERM_PROG=<base64 htop>` resolves to `htop` (base64 decoded)
- `OSC 633;E;Get-ChildItem` resolves to `Get-ChildItem`

The lifecycle is correct too: `OSC 133;D` reverts the pane to the process tree heuristic (I saw it fall back to `powershell`), and an idle pane never returns empty. The visible TUI window behaves identically.

**Regressions.** Green. The change to `#{pane_current_command}` is purely additive: OSC first, existing heuristic as fallback, and shells that emit nothing are entirely unaffected.

One tiny observation, not a blocker: the commit body mentions `cmdline=` (kitty's `%q` quoted bash/zsh form) being stored verbatim, but `parse_cmdline_param` only handles `cmdline_url=` today and returns None for bare `cmdline=`, which matches the PR description. Totally fine as a documented follow up; `cmdline_url=` covers the pwsh snippet and fish paths, and full POSIX quote decoding is genuinely fiddly. Wanted to flag it so the docstring and the behavior stay in sync if a kitty bash user files something later.

I squashed the four commits into one with a summary message, so the cascade lands as a single clean entry in history. You remain the author. I also landed a couple of my safety and end to end tests alongside so the decoders stay guarded going forward. Great work, and thanks again.

Issue #272, opened 2026-04-30
status-format: #(cmd) re-spawns subprocess per frame push, slow shell helpers cause typing lag


nobane:
## Summary

`run_shell_command` in `src/format.rs:253` re-spawns a subprocess on every call. The doc comment above it mentions caching, but the function itself doesn't currently implement one. Combined with the server-push path in `src/server/mod.rs:4063-4087`, which calls `expand_format(...)` for `status_left`, `status_right`, and each `status_format[i]` on every `state_dirty` push, any `#(cmd)` in the user's status format runs once per pushed frame.

During active typing the server pushes ~30 frames/s. With a fast helper like `sh -c cmd` (~1-3ms) the cost is negligible. A slow helper changes that. `powershell -NoProfile -ExecutionPolicy Bypass -File ...` is ~200-500ms cold-start on Windows (CreateProcess + Defender + .NET startup), so the synchronous `Command::output()` in `run_shell_command` blocks the server's main loop and input echo lags 200-500ms per keystroke.

## Repro

`~/.psmux.conf`:

```tmux
set -g status-style "bg=#4d94c2,fg=default"
set -g status-format[0] "#[fill=#4d94c2]#[bg=#4d94c2,fg=black] 🧊 #[bg=black,fg=#80d4ff,bold]  fridge  #[align=right]#[bg=black,fg=#80d4ff,bold] #(powershell -NoProfile -ExecutionPolicy Bypass -File C:/Users/me/bin/uptime.ps1) #[default]"
```

`uptime.ps1`:

```powershell
$port = Get-ChildItem "$env:USERPROFILE\.psmux\*.port" | Where-Object { $_.Name -ne "__warm__.port" } | Select-Object -First 1
if ($port) {
  $d = [int]([DateTime]::Now - $port.CreationTime).TotalSeconds
  $h = [math]::Floor($d / 3600); $m = [math]::Floor(($d % 3600) / 60)
  if ($h -gt 0) { "{0}h {1}m" -f $h, $m } else { "{0}m" -f $m }
}
```

Attach a pane running anything that redraws frequently (Claude Code, btop, neovim with active LSP) and type. Expected: snappy echo. Actual: 200-500ms lag per keystroke. Removing the `#(...)` from `status-format[0]` restores snappy typing immediately.

## Code references

`src/format.rs:249-266`:

```rust
/// Execute a shell command and return its stdout (trimmed).
/// Used for `#(command)` expansion (tmux compatibility).
[...]

PR #477, opened 2026-07-17
feat: async #(command) status expansion (non-blocking format jobs)

MattKotsenas:
## What

Make `#(command)` status expansion asynchronous. `run_shell_command` no longer runs the subprocess with a blocking `Command::output()` on the server main loop. It returns the last cached output immediately (empty before the first result), spawns the command on a background thread, and the server loop drains the result to refresh the cache and repaint.

## Why

A `#(cmd)` in a status format was expanded synchronously on the server main loop, so a slow helper stalled the whole server push. #272 mitigated the frequency with a TTL cache, and its discussion suggested a better fix:

> First call after expiry still blocks ~280 ms once ... running #(...) on a
> worker pool would eliminate even that, but it's a bigger change than the bug
> warrants.

This is that worker model. tmux runs `#()` the same way (an async job that paints the previous result and updates on completion).

## How

Mirrors the existing async `run-shell` path (background thread + a result channel drained each loop iteration):

- `types.rs`: the cache value becomes `ShellEntry { at, value, running }`; add a
  `format_job_rx/tx` channel, preinitialized in `run_server`.
- `format.rs`: `run_shell_command` returns the cached value and, when it is
  missing or older than the TTL and not already in flight, spawns a background
  worker. A command already running is not re-spawned, so a burst of pushes runs
  it at most once per TTL.
- `server/mod.rs`: drain completed results each loop iteration, refresh the cache,
  and mark the frame dirty so the fresh output is painted.

Otherwise unchanged: same TTL (`status-interval`, 1s floor) and `cfg!(windows)` shell; workers inherit the server's `PSMUX_TARGET_SESSION`.

## Testing

`tests-rs/test_issue272_format_shell_cache.rs` is rewritten for the async contract (7 tests)

Follow-up to #272 (retires the residual per-interval block its cache left).

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -246,53 +246,63 @@ pub fn expand_format_for_window(fmt: &str, app: &AppState, win_idx: usize) -> St
     result
 }
 
-/// Execute a shell command and return its stdout (trimmed).
-/// Used for `#(command)` expansion (tmux compatibility).
+/// Expand `#(command)` (tmux compatibility): return the last cached stdout and,
+/// when it is missing or older than the TTL, spawn a background worker to
+/// refresh it. The call never blocks the server loop; the worker's result is
+/// delivered over `format_job_tx` and applied by the drain in the server loop,
+/// which then repaints. Before the first result the expansion is empty.
 ///
-/// Results are cached in `app.format_shell_cache` keyed by the command
-/// string, with TTL = `status-interval` seconds (matching tmux's refresh
-/// semantics for `#(...)`). When `status-interval` is 0, a 1s floor is
-/// used so typing latency is bounded — tmux never repaints fast enough
-/// for the spawn cost to dominate, but our server-push path can fire
-/// ~30 times per second during active TUI redraws.
-///
-/// Cache miss / expiry path spawns a fresh subprocess synchronously.
-/// First call after expiry pays the spawn cost; subsequent calls within
-/// the TTL window return instantly. See issue #272.
+/// Cached in `app.format_shell_cache`, keyed by command string, TTL =
+/// `status-interval` seconds (1s floor). A command already in flight is not
+/// re-spawned, so a burst of pushes runs it at most once per TTL. See issue #272.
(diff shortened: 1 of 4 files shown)


psmux:
Merged. Thank you @MattKotsenas, this is exactly the worker model the #272 discussion hoped someone would build, and you built it cleanly.

I never take a claim at face value, so I verified the whole thing from the outside in before reading a line of the diff:

**Reproducing the stall on master**

With a slow `#()` helper (a ping that takes about 2 seconds) in the status line and a client attached, I sampled TCP latency against the server every 100 ms for 14 seconds:

```
display-message with slow #(): 2093 ms   (synchronous Command::output on the server)
TCP sampling: p50 2 ms, max 1998 ms, 5 spikes over 500 ms in 14 s
```

The server loop stalled for the full duration of the helper on every cache expiry. Claim confirmed.

**Same test on your branch merged onto master**

```
display-message with slow #(): 17 ms    (returns cached value, worker spawned in background)
TCP sampling: p50 2 ms, max 20 ms, zero spikes
127 samples collected vs 45 on master, because the server never stopped answering
```

A fast `#(echo MARKER)` returns empty on the very first expansion and the marker lands on the next paint once the worker reports in, which matches how tmux runs `#()` jobs. The rendered status line picks up the full helper output as soon as the result drains.

**Regression sweep**

All 2431 Rust tests pass, including your 7 rewritten issue 272 tests. The existing status format perf suite now shows keystroke echo with a slow helper at 0.9x of the no helper baseline, meaning the penalty is simply gone. The format engine suite (74 checks) and format vars suite (25 checks) are all green.

I also liked the details: the `running` flag keeps a burst of pushes from double spawning, the old freshness base is kept while a worker is in flight, and the guard is dropped before the thread spawns. The channel preinit in `run_server` with the explanation of why it cannot be lazy was a nice touch.

Great engineering, and a proper close to the residual block that #272 left behind. Much appreciated.

PR #478, opened 2026-07-17
Fix duplicate client attachment registration

imonuryalcin:
## Summary

- make client attachment registration idempotent by server-assigned client ID
- use the same registration path for interactive and control-mode clients
- add regressions for duplicate interactive and control-mode attach events

## Problem

A connection can emit `ClientAttach` more than once for the same client ID. The client registry already deduplicates those events with `entry(...).or_insert(...)`, but `attached_clients` was incremented before the registry insertion.

After two attach events and one matching detach, the registry was empty while `attached_clients` remained `1`. That inconsistent state kept the server logically attached even though no client remained, which could surface as ghost-client or detach-policy behavior.

## Root cause and fix

Registration accounting and registry insertion were separate operations. This change centralizes them in `AppState::register_client` and reports whether the ID was newly registered.

The interactive and control-mode paths now:

- increment `attached_clients` only when a new client ID is inserted
- run attach hooks and refresh environment state only on first registration
- preserve the existing behavior where only interactive clients replace `latest_client_id`

There are no wire-protocol, command-line, or output-format changes.

## Validation

- `cargo test --bin psmux reap_client` — 8 passed, 0 failed
- `cargo build --release` — passed
- `git diff origin/master...HEAD --check` — passed
- Windows persistent-connection E2E: duplicate attach for one client ID followed by one detach returns the session to detached state
- Windows attach/detach E2E: `list-windows`, `list-panes`, and `capture-pane` remain available before, during, and after a client attach

This is a follow-up hardening for the client reaping and attachment-state work around #434.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1782,33 +1782,23 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                     let _ = resp.send(format!("{}\n", line));
                 }
                 CtrlReq::ClientAttach(cid) => {
-                    app.attached_clients = app.attached_clients.saturating_add(1);
-                    app.latest_client_id = Some(cid);
-                    // Register in client registry if not already present
-                    app.client_registry.entry(cid).or_insert_with(|| {
-                        let tty = format!("/dev/pts/{}", cid);
-                        crate::types::ClientInfo {
-                            id: cid,
-                            width: app.last_window_area.width,
-                            height: app.last_window_area.height,
-                            connected_at: std::time::Instant::now(),
-                            last_activity: std::time::Instant::now(),
-                            tty_name: tty,
-                            is_control: false,
-                        }
-                    });
-                    hook_event = Some("client-attached");
-                    // update-environment: refresh env vars from the attaching client's environment
-                    let update_vars = app.update_environment.clone();
-                    for var_spec in &update_vars {
-                        let remove = var_spec.starts_with('-');
-                        let name = if remove { &var_spec[1..] } else { var_spec.as_str() };
-                        if remove {
-                            app.environment.remove(name);
-                        } else if let Ok(val) = std::env::var(name) {
-                            app.environment.insert(name.to_string(), val);
-                        } else {
-                            app.environment.remove(name);
+                    // Registration and the attached counter are one idempotent
+                    // operation. A duplicate attach for the same connection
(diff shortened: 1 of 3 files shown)


psmux:
Merged. Welcome to psmux @imonuryalcin, and thank you for a first contribution that came with a clear invariant argument and regression tests.

I verify every claim from the outside before reading a diff, so here is what I found on master before merging:

**Trying to reproduce the ghost attach**

I attacked `attached_clients` from three directions over the raw TCP protocol:

1. Duplicate attach verbs on one persistent connection: the second one is swallowed by the per connection guard, so only one `ClientAttach` ever reaches the server loop.
2. A control mode client (the iTerm2 pattern) that registers and then issues `attach-session` twice: accounting stayed balanced and `#{session_attached}` returned to 0 after disconnect.
3. Attach, then `detach-client -s`, then reattach on the same connection: the server kicks the connection during the mass detach, and accounting returned to 0.

So on current master the double count is not reachable over the wire. The attach and detach events are forced to strictly alternate per connection by a single connection local flag.

**Why I merged it anyway**

That safety currently lives in one mutable local variable inside a three thousand line connection handler, plus an unstated assumption that no second code path ever emits `ClientAttach` for a live id. Your change moves the guarantee to where it belongs: `register_client` makes the counter and the registry a single atomic decision, exactly mirroring what `reap_client` already did for the detach side after #434. The `reap_client` doc comment even promised that attach increments and inserts together, and until now nothing enforced it. Control clients and interactive clients now register through one path, which is also closer to how tmux itself derives attachment from a single client list rather than a separate counter.

I also checked the two behavior details that could have bitten: control clients still never replace `latest_client_id`, and the attach hook plus environment refresh now fire only on first registration, which is the correct semantics if a duplicate ever does appear.

**Verification on the merged tree**

All three reproduction vectors above behave identically after the merge. A real attached TUI client cycle (attach, send keys, capture, detach) works cleanly. All 2433 Rust tests pass including your two new ones, the ghost client E2E suite is 8 for 8, and the control mode suite is 19 for 19.

One small note for next time: the test comment says a persistent connection may send both `client-attach` and `attach-session` for the same id, but both verbs share the same guard flag, so that pair cannot double fire today. The value of your patch is that nobody has to rely on that flag anymore.

Solid first PR. Thank you for hardening the attachment bookkeeping.

Issue #457, opened 2026-07-09
SSH attach sends mouse-enable on ConPTY builds that can't accept mouse input — one click crashes conhost (0xc0000409) and kills the session


jwr456:
## Environment
- psmux v3.3.6 (reports `tmux 3.3.6`)
- Windows 10 build 19045 (10.0.19041.5198 conhost), attach over Microsoft OpenSSH (`DefaultShell` = pwsh 7)
- Client: any mouse-capable terminal (Windows Terminal, xterm.js, etc.) SSHing in and running `psmux -L <sock> attach -t <name>`

## What happens
`~/.psmux/ssh_input.log` shows psmux itself KNOWS the platform can't take mouse input, then enables it anyway:

```
WARNING: Windows build 19045 < 22523 - ConPTY does NOT support mouse over SSH. Mouse clicks will not work. Upgrade to Windows 11 22H2+ for SSH mouse support.
...
send_mouse_enable: writing mouse-enable VT sequences to stdout
send_mouse_enable: WriteFile ok=1 written=32
stdin console mode: 0x0298 VTI=true MOUSE=true
```

The mouse-enable sequences reach the user's terminal, the terminal starts mouse reporting, and the first click/drag sends a mouse report (`ESC[<…M` / `ESC[M…`) back over SSH. That report hits the ConPTY input path and **crashes conhost.exe**: WER `BEX64`, exception `0xc0000409` (fast-fail / stack buffer overrun), faulting module `conhost.exe 10.0.19041.5198`, offset `0x68eff` — identical offset on every occurrence. The crash tears down the attach's ConPTY, killing the pane process (in our case a long-running AI agent session).

We first worked around this on our own web-terminal path by stripping DECSET/DECRST mouse toggles (`ESC[?1000/1002/1003/1006…h|l`) from the pane→browser stream and dropping inbound mouse reports — that works, but any *native* terminal attach (the normal SSH case) is still one click away from losing the session.

## Expected
On builds < 22523 (the build check `send_mouse_enable` already performs for its warning), psmux's SSH input module should:
1. **Not** write the mouse-enable VT sequences to the client, and suppress/strip mouse-enable DECSET forwarded from pane content, and
2. **Drop inbound mouse report sequences** (`ESC[M` + 3 bytes, `ESC[<…M|m`) before they reach ConPTY input, as belt-and-suspenders.

Mouse simply not working (as the warning already promises) is fine; the session dying is not.

## Repro
[...]

Issue #245, opened 2026-04-21
psmux + opencode mouse cannot correctly select according to the program layout


Edemilorhea:
I recently discovered an issue, but I'm not sure if it's an intentional design or a bug.

When I use psmux + opencode, opencode renders its own layout divided into two sides: the left side for content and the right side for the information column.

When I use the mouse to select text, the selection on the left side should ideally follow the corresponding lines. However, when using opencode within psmux, the selection is intercepted by psmux, causing it to select across the layout.

I would like to ask if there is a way to disable this cross-layout selection?

Below is the demonstration

Psmux:

<img width="2129" height="1145" alt="Image" src="https://github.com/user-attachments/assets/f3e444a9-f2c6-49c9-b1c3-36b093131bf4" />

Normal:

<img width="1243" height="504" alt="Image" src="https://github.com/user-attachments/assets/5ac027d5-29f9-4558-b134-0f2a8a5f92a0" />

psmux:
Hi, thanks for the detailed report and the screenshots, that made the behaviour very clear.

You are correct, this was psmux's own client-side drag-selection overlay being rendered on top of opencode. Because opencode draws its own two-column layout inside a single pane, psmux had no way to know about those internal columns, so its selection rectangle stretched across the whole pane and did not respect opencode's left and right sides.

I added a new option for exactly this case:

```conf
# put this in your psmux.conf (or run it interactively with prefix + : )
set -g mouse-selection off
```

When off, psmux stops drawing its own drag selection. Mouse events are still forwarded to the active pane, so click to focus a pane, the scroll wheel, and any application that requests mouse tracking (opencode, nvim, htop, claude, etc.) all keep working. Default is on so existing behaviour is unchanged. To switch back at any time:

```conf
set -g mouse-selection on
```

You can also toggle it live without restarting:

```
psmux set-option -g mouse-selection off
```

This is independent of the existing `mouse` option (which controls whether psmux receives mouse events at all) and `pwsh-mouse-selection` (Windows 11 PowerShell style selection).

Available in the next release. If you would like to try it before then, you can build from `master` or grab the installer from the next build artifact.

Tested with:
- 12 Rust unit tests covering option default, config parsing, set-option dispatch, options catalog registration, and independence from the other mouse options
[...]

PR #479, opened 2026-07-18
Let apps that enable a mouse protocol handle their own text selection in their pane

neich:
I hit this while using [gaviero](https://github.com/neich/gaviero), my terminal editor, inside a psmux pane. Gaviero turns on mouse reporting and does its own click-and-drag text selection — but when I drag over text, nothing gets selected in the editor. psmux grabs the drag and paints its own selection box on top of it. This PR makes psmux step aside, per pane, for any app that asked for the mouse (gaviero, opencode, nvim with `set mouse=a`), and adds a keepalive so Windows Terminal can't quietly break mouse reporting while you work.

## Environment

- psmux master (this branch)
- The selection change applies everywhere — it's one new field in the layout JSON the server sends, plus one check in the client's click handling.
- The keepalive is Windows-only: Windows Terminal sometimes drops a client's mouse registration (seen after window resizes and in long-lived sessions). It does nothing on Unix, and nothing on ConPTY builds older than 22523, where #457 says mouse must not be touched at all.

## What happens

When the app in a pane has mouse reporting on and I drag over its text, psmux paints its own selection instead of letting the app select. The app gets the initial press (presses are always forwarded), but psmux keeps the drag and the release for its overlay — from the app's point of view the gesture ends the moment the button goes down. The cause: on left-down the client only checks the **global** `mouse-selection` option. The server never tells it "this pane's app wants the mouse", so the client can't step aside for one pane and stay active in the rest. The closest existing flag, `alternate_screen`, is the wrong test — `less` runs on the alternate screen with no mouse support, and psmux selection has to keep working there.

Until now my way out was the global option from #245, `set -g mouse-selection off`. That lets gaviero select, but it also turns off psmux drag-selection in every plain shell pane, where I do want it.

On Windows there's a second problem underneath. Windows Terminal sometimes silently forgets that the client asked for mouse reporting: keys keep working, the mouse goes completely dead until detach/reattach. psmux only re-sent the mouse-enable sequences in SSH mode — a local client never did. And the full `send_mouse_enable()` can't simply be called locally: it switches the console's VT-input flag back on (keystrokes then arrive as raw escape bytes and show up as garbage), and its DSR probe makes the terminal answer `ESC[0n`, which lands in the active pane as typed characters.

## Expected

For a pane whose app explicitly enabled a mouse protocol (DECSET 1000/1002/1003 — the real request, not an alt-screen guess), the psmux client should:

1. **Not** start its own drag selection in that pane, even with `mouse-selection on`, and pass the full press–drag–release through so the app selects by itself. Every other pane keeps psmux selection. This is the per-pane, automatic version of the #245 option.
2. **Keep the mouse registration alive** so point 1 stays true over time: re-send mouse-enable at attach, on resize, and every 30 seconds, in every input mode — each mode getting only what's safe for it. Pipe mode (mintty/Cygwin) re-sends its usual reduced set; SSH/VT mode runs the full `send_mouse_enable()`; a local Windows console gets just the DECSET bytes plus an `ENABLE_MOUSE_INPUT` re-assert (VT-input flag untouched, no DSR probe); Unix and pre-22523 builds get nothing, exactly as #457 requires.

The server carries the signal as one new field on every layout leaf, filled in by both dump paths (the fallback leaves used when a pane can't be read send `false`):

```json
{"type":"leaf","id":7,...,"alternate_screen":true,"wants_mouse":true,...}
```

The field is `#[serde(default)]`, so an old server that doesn't send it leaves the client behaving exactly as before, and an old client simply ignores it coming from a new server. Shell panes keeping psmux drag-selection is fine; a mouse-aware app losing its own selection is not.

## Repro

1. Default config (`mouse-selection on`). In one pane run gaviero, opencode, or `nvim` with `set mouse=a`; keep a plain shell in another.
2. Drag over text inside the app's pane.
3. Before this branch: psmux draws its overlay and the app never sees the drag. After: the app selects, and the shell pane still gets psmux selection.
4. For the keepalive: attach locally in Windows Terminal and resize the window (or just leave the session running). Before: the mouse eventually goes dead until reattach. After: it keeps working.

The new `tests-rs/test_pane_wants_mouse_selection.rs` covers the old-server default, the exact JSON shape the streaming writer emits, and the per-pane lookup; seven existing tests gained a mechanical `wants_mouse: false` line.


psmux:
Merged. Thank you @neich, this is a genuinely well engineered contribution, and welcome to psmux.

I verify every PR by reproducing the problem before reading the diff, so here is the evidence trail:

**Reproducing the steal on master**

I built a press, drag, release injector over `WriteConsoleInput` and drove a real gesture into an attached psmux window running nvim with `set mouse=a` (which the server flags exactly like gaviero). On master:

```
after drag: nvim VISUAL mode = False
after drag: clipboard = ' juliet kilo lima mik'   (psmux selection copied the pane text)
```

nvim never saw the gesture and psmux quietly replaced my clipboard with screen text. Confirmed, and honestly a little worse than your writeup describes: in my runs the app did not even get the press.

**Same gesture on your branch merged onto master**

```
after drag: nvim VISUAL mode = True
after drag: clipboard = 'CLIPBOARD_SENTINEL_479'  (untouched)
```

The app selects for itself and psmux stays out of the way. A drag in a plain shell pane still runs psmux selection and copies as before, so the per pane split works exactly as designed. I also confirmed the strictness you aimed for: the shell pane reports `wants_mouse` false while the nvim pane reports true, so `less` style alt screen apps keep psmux selection. This matches how tmux treats a pane whose program asked for the mouse, which is the right model for us to follow.

**Keepalive safety**

I let a local attached client sit past the 30 second boundary and then exercised it: typed text arrives clean (no VT input garbling), no `[0n` or stray escape artifacts leaked into the pane, and drag selection still works afterward. Your analysis of why the full `send_mouse_enable` must not run on a local console (the VTI restore and the DSR probe) is exactly right, and the routed keepalive avoids both hazards.

**Regression sweep**

All 2436 Rust tests pass including your three new ones. Ten mouse related E2E suites ran against the merged build: the #245 suite (15 for 15, the option your change generalizes per pane), the #457 SSH gate suite (7 for 7, your build gating holds), the #349 hover suite (16 for 16), the #285 nvim suite, wheel copy mode, native TUI mouse, and the rest all green. One suite, `test_issue277_scroll_controlled`, failed 2 of 8 checks, and I reran it on unmodified master before blaming you: it fails identically there, so it is a preexisting environmental issue, not your regression.

One finding from my testing you may enjoy: a plain .NET console child that writes the DECSET registration through `WriteConsole` never gets it into the pane's VT stream (ConPTY eats it), so `wants_mouse` stays false for such apps. Real VT writers like nvim and gaviero come through fine. That is a ConPTY translation quirk, not a flaw in your approach, but it explains why my first synthetic test app saw nothing.

The serde defaulted field for old server and new client compatibility, the fallback leaves emitting false, and the deliberate refusal to reuse the alt screen heuristic all show real care. Enjoy gaviero inside psmux, that is exactly the kind of use this project exists for. Thank you again.

PR #480, opened 2026-07-19
Support mouse scrollback over Windows 10 SSH

imonuryalcin:
## Summary

- extend the existing raw VT pipe client path to support Windows OpenSSH sessions launched without a remote PTY
- add a portable `psmux-win` client helper for macOS and Linux that enables SGR mouse reporting locally, uses `ssh -T`, and always restores terminal state
- add parser, copy-mode, wrapper cleanup, installer, and Windows direct-pipe integration coverage
- document the Windows 10 byte-flow limitation and the supported workaround

## Problem

On Windows 10, a normal OpenSSH session places psmux behind ConPTY. That ConPTY version consumes outgoing DEC mouse-mode sequences such as `CSI ? 1000 h`, `CSI ? 1002 h`, and `CSI ? 1006 h`, so the client terminal never starts reporting the wheel. Even when mouse reporting is enabled separately on the client, inbound SGR reports such as `CSI < 64 ; 10 ; 5 M` do not reach psmux reliably through the old ConPTY input parser.

The existing Windows build gate must remain: forcing mouse reporting through affected ConPTY versions can crash the pane process (issue #457).

## Solution

The helper uses the supported OpenSSH no-PTY mode (`ssh -T`), removing ConPTY from the path. Windows OpenSSH then gives psmux anonymous stdin/stdout pipes. psmux detects that SSH pipe, reads VT bytes with its existing incremental parser, emits mouse/focus/paste modes directly, and obtains terminal dimensions through XTWINOPS.

The POSIX wrapper:

- saves the exact local terminal state
- enables xterm button/drag reporting and SGR coordinates
- forwards raw keyboard and mouse bytes over `ssh -T`
- restores stty and disables mouse reporting on normal exit, SSH failure, disconnect, `HUP`, `INT`, or `TERM`
- validates values interpolated into the Windows remote command

Normal SSH PTY behavior on supported Windows 11 builds is unchanged.

## Validation

- `cargo test --bin psmux` — 2443 passed
- focused SGR parser and scroll/copy-mode tests — 7 passed
- `cargo build --release` — passed
- Windows anonymous-pipe E2E — 6 passed, including repeated WheelUp and `mouse off`
- real macOS PTY -> OpenSSH `-T` -> Windows 10 psmux — DECSET 1000/1002/1006 observed, WheelUp entered copy mode at offset 3, the second report advanced to 6, and detach emitted mouse-disable cleanup
- POSIX wrapper and installer tests passed under Windows Git `sh` and macOS `/bin/sh`
- shell, Python, PowerShell, and workflow YAML syntax checks passed
- `git diff --check` — passed

## Notes

`cargo fmt --check` is already non-clean on `origin/master` because of repository-wide formatting drift. The new standalone Rust test file passes `rustfmt --check`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,3 +65,26 @@ jobs:
           $exe = (Resolve-Path "target\x86_64-pc-windows-msvc\release\psmux.exe").Path
           $env:PSMUX_EXE = $exe
           .\tests\run_smoke_integration.ps1
+
+      - name: Run raw SSH pipe mouse integration
+        shell: pwsh
+        run: |
+          $exe = (Resolve-Path "target\x86_64-pc-windows-msvc\release\psmux.exe").Path
+          .\tests\test_windows10_ssh_pipe_mouse.ps1 -Psmux $exe
+
+  posix-helper-tests:
+    name: POSIX helper tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Test SSH wrapper cleanup
+        run: sh tests/test_psmux_ssh_wrapper.sh
+
+      - name: Test helper installation
+        run: sh tests/test_install_psmux_win.sh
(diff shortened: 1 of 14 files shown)


psmux:
Merged, thank you @imonuryalcin! This was an impressively complete contribution. I verified it from every angle before merging: the full Rust suite passed 2450 with zero failures, your wrapper and installer tests passed under real Git `sh`, and I additionally threw nine command injection attempts at `psmux-ssh.sh` through the session, socket and path arguments and every one was rejected by your validation before reaching the ssh command line. The Windows pipe E2E passed 6 of 6, the headless detection path for piped stdin without SSH still prints version and exits, and `PSMUX_PIPE_VT=0` still works as the escape hatch. Windows 10 users finally get mouse scrollback over SSH. Excellent work.

Issue #414, opened 2026-06-21
set -g pane-border-status top   not working


laciferin2024:
```
set -g pane-border-status top  

```

I want to enable pane titles and also have the rename functionality that tmux supports natively.

psmux:
Hey, thanks for reporting this and sorry for the rough edge.

You were right. I reproduced it and confirmed the bug. Here is what was happening under the hood.

`pane-border-status top` was half working. It correctly reserved a row at the top of each pane for the border label (you can see the pane content shrink by one row), and it drew the border line, but the pane title never appeared on it. So you ended up with a blank border instead of a titled one.

The root cause: psmux was shipping an empty default for `pane-border-format`. The renderer only draws the label when that format is non empty, so with just `set -g pane-border-status top` in your config there was nothing to render. In tmux, `pane-border-format` has a sensible default, which is why enabling the status alone "just works" there.

Fix: psmux now falls back to the tmux style default of `#{pane_index} "#{pane_title}"` whenever `pane-border-format` is unset or empty, applied on both render paths (the local session and remote attach). So this alone is now enough:

```
set -g pane-border-status top
```

And if you want to customize the label, setting your own format still overrides the default:

```
set -g pane-border-status top
set -g pane-border-format " #{pane_index}: #{pane_title} "
```

On the rename side, that already works today. `select-pane -T "my title"` sets and locks a pane title, and it shows up on the border and in `#{pane_title}` / `#T`. If you want programs (like a shell) to update the title dynamically via OSC sequences instead, flip on `set -g allow-set-title on`.

I verified the fix by capturing the real rendered output under a pseudoconsole: with only `set -g pane-border-status top`, both pane titles now render on their borders. It is committed and will be in the next build. Give it a try once that lands and let me know if the layout looks right on your end.

Issue #288, opened 2026-05-11
pane-border-status bottom/top: status bar overlaps pswh input area due to incorrect pane height calculation


rayfalling:
## Environment
- psmux version: 3.3.4
- Terminal: Windows Terminal (PowerShell 7)
- OS: Windows 11

## Steps to reproduce
Start psmux with the following minimal config:
 ```
# Enable pane border labels at the bottom of each pane
set -g pane-border-status bottom
# Customize what the label shows
set -g pane-border-format "#{pane_index}: #{pane_title}"
 ```
<img width="1049" height="49" alt="Image" src="https://github.com/user-attachments/assets/9d0d7291-0efc-49d3-842a-7ed324e62165" />

psmux:
## Root Cause Identified and Fixed

**Commit:** aff8b06

Hey @rayfalling, thanks for the clear report and screenshot. You were spot on about the height calculation.

### What Was Happening

When ``pane-border-status`` is set to ``top`` or ``bottom``, psmux renders a label line at the edge of each pane (showing the pane index, title, etc.). The problem was that the pane's PTY (the virtual terminal) was sized to the **full** allocated area without reserving a row for that label. The rendering code then drew the label directly on top of the last row of pane content.

This caused two overlapping issues:

1. The border label overwrote the PowerShell prompt line (or the last visible line of output)
2. The shell itself believed it had more rows available than were actually visible, so content near the bottom of the pane would get hidden behind the label

### The Fix

Three coordinated changes:

**Rendering (both local TUI and remote client):** When ``pane-border-status`` is active, the content area is now computed as ``area.height - 1``. The excluded row is used exclusively for the border label. For ``bottom``, content occupies the top portion and the label goes at the very bottom. For ``top``, the label goes at the top and content shifts down by one row.

**PTY Sizing (``resize_all_panes``):** When computing the PTY dimensions for each pane, 1 row is subtracted from the allocated height when ``pane-border-status`` is enabled. This ensures the shell knows its true visible height and wraps/scrolls correctly.

**Option Change Trigger:** Setting ``pane-border-status`` via ``set-option``, ``source-file``, or config file now triggers an immediate pane resize so the height adjustment takes effect right away (not just on the next split or window creation).

### Verified Scenarios

| Scenario | Before Fix | After Fix |
|----------|-----------|-----------|
| Single pane, ``bottom`` | pane uses all 30 rows, label overlaps last row | pane uses 29 rows, label on row 30 |
[...]

PR #481, opened 2026-07-20
fix(client): caret & mouse row offset under pane-border-status top (Agent Teams)

kt81:
## Problem

With `pane-border-status top`, the text caret renders **one row above** the actual content line, and mouse text selection is off by one row (you have to aim one row up to grab the intended text).

This is easy to hit in practice because **Claude Code agent teams enable `pane-border-status top`** to label teammate panes — so anyone using the flagship Agent Teams feature runs into it. Once enabled it affects the whole session (every pane, every window) and survives detach/reattach and terminal resize, because it is deterministic, not a render desync.

スクリーンショット 2026-07-20 205850
スクリーンショット 2026-07-20 213503

## Root cause

#288 reserved the pane's first row for the border label and shifted pane **content** down into an inner rect in `render_layout_json`, but the caret placement and the screen→cell mouse mappings kept using the pane's **outer** rect:

- the post-draw cursor write used `compute_active_rect_json_zoom_aware()` (outer rect);
- mouse-forward `rel_row = me.row - pane_rect.y` (outer rect) — for every button / drag / move event;
- client-side text selection (`extract_selection_text`, `word_bounds_at`) via `collect_leaves` (outer rect).

So content was drawn at `area.y + 1 + r` while the caret/mouse math used `area.y + r` → a consistent one-row offset. Only `top` is affected; `bottom` keeps the content's top origin.

Introduced by #288 (v3.3.4) and present in every release since, including 3.3.6 and 3.3.7. The local (non-client-server) render path in `rendering.rs` already placed the caret with the inner rect, so only the client-server path was wrong.

## Fix

Extract the reserved-row logic into a single `pane_content_inner(area, border_status, border_format)` helper — mirroring exactly what the renderer uses — and route the caret write, the six `rel_row` mouse-forward sites, and the client-side text selection through it, so they all use the same content rect as the render. `off` / empty-format are identity; `bottom` is unchanged. Shared rect helpers (`collect_leaves`, `collect_pane_rects`, `compute_active_rect_json_zoom_aware`) intentionally keep returning the outer rect, since pane hit-testing and border highlighting need it.

## Tests

- New `tests-rs/test_pane_border_status_cursor.rs`: covers `top` / `bottom` / `off` / empty-format / 1-row-pane behavior, and the caret↔content and mouse↔content row alignment (the regression guard).
- Existing selection tests updated for the new signature (they pass `"off"`, which is identity — behavior unchanged).
- Full suite: **2443 passed, 0 failed**.
- Local E2E

スクリーンショット 2026-07-20 221459

## Related

- #288 — reserved the label row (introduced this offset).
- #414 — made the label actually render with the default format, which is what makes the offset visible in the common `set -g pane-border-status top` setup.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -187,6 +187,22 @@ pub(crate) fn build_osc8_overlay(runs: &[HyperlinkRun]) -> String {
     out
 }
 
+/// Content area of a pane after reserving the `pane-border-status` label row.
+/// Must match `render_layout_json`'s `inner`; the caret and every screen→cell
+/// mouse mapping route through this so they stay aligned with the content (#288).
+pub(crate) fn pane_content_inner(area: Rect, border_status: &str, border_format: &str) -> Rect {
+    let has_border_label = border_status != "off" && !border_format.is_empty() && area.height > 1;
+    if !has_border_label {
+        return area;
+    }
+    if border_status == "top" {
+        Rect::new(area.x, area.y + 1, area.width, area.height.saturating_sub(1))
+    } else {
+        // "bottom": content keeps its top origin, only the last row is reserved.
+        Rect::new(area.x, area.y, area.width, area.height.saturating_sub(1))
+    }
+}
+
 fn collect_leaves<'a>(node: &'a LayoutJson, area: Rect, out: &mut Vec<PaneLeaf<'a>>) {
     match node {
         LayoutJson::Leaf { rows_v2, .. } => {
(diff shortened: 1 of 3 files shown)


psmux:
Merged, thank you @kt81! Before merging I reproduced the offset independently: a standalone check of the caret and mouse math showed every case landing one row off under `pane-border-status top`, exactly as you described, and your `pane_content_inner` helper aligned all of them. Full suite came back 2443 passing with zero regressions, and a live TUI run with border labels on stayed pixel correct through `top`, `bottom` and `off`. Really appreciated the clean root cause writeup and the regression tests you included. This one bit everyone using Agent Teams, so it is a great catch.

Issue #510, opened 2026-07-31
Startup orphan reaper terminates every psmux server on the machine when the invocation resolves a different data dir (-L gives no protection)


acoliver:
## Summary

`reap_orphaned_servers()` runs on **every** psmux invocation (`main.rs:223`) and terminates every live psmux server on the machine that the *invoking process's* data dir cannot account for. The candidate set is machine-wide (`GetExtendedTcpTable` + image-name filter) but the authority is a single `~/.psmux` resolved from `USERPROFILE`/`HOME`, so any invocation resolving a **different** home reaps every other instance's servers.

The extreme case is a data dir that exists but tracks nothing: a single read-only command such as `psmux -L probe list-sessions` then kills every psmux session on the box, in every namespace, on every `-L` socket.

This is not limited to exotic setups. The repo's own `tests/psmux_test_helpers.ps1` (`New-PsmuxTestEnv`) creates exactly that state, so any test using it destroys the developer's live sessions — making #342 reproducible again inside the mechanism added to fix it.

## Environment

- psmux version: `psmux 3.3.7 (05cc5d4 2026-07-20)`; code path verified identical on master `615d2a5`
- OS: Windows `10.0.26200.0`
- Shell: PowerShell 5.1.26100.8875
- Terminal: Windows Terminal

## Reproduction

> This is destructive as written: it kills every psmux session on the machine.

```powershell
# 1. A session you care about, on its own -L socket.
psmux -L keep new-session -d -s keep
Start-Sleep -Seconds 15          # exceed ORPHAN_REAP_MIN_AGE (10s)

# 2. Any psmux command from a shell whose home resolves elsewhere.
#    New-PsmuxTestEnv does exactly this; here it is by hand.
$fresh = Join-Path $env:TEMP ("psmux_probe_" + [guid]::NewGuid().ToString('N').Substring(0,8))
New-Item -ItemType Directory -Path (Join-Path $fresh '.psmux') -Force | Out-Null
$env:USERPROFILE = $fresh; $env:HOME = $fresh

psmux -L probe list-sessions     # read-only, different -L, different home

# 3. The `keep` server is gone.
```

## Expected

An invocation resolving a different data dir leaves other instances' servers alone. A read-only `list-sessions` on socket `probe` must never terminate a server on socket `keep`.

## Actual

Every live psmux server is terminated.

[...]

Issue #443, opened 2026-07-05
capture-pane -p drops cells skipped by cursor-forward (ESC[nC), collapsing words


matthoneycutt-sm:
## Summary

`capture-pane -p` drops the cells skipped by a cursor-forward (CUF, `ESC[nC`) advance, so any text laid out by moving the cursor forward instead of writing literal spaces comes back with the gaps removed and the words run together. Absolute-column positioning (`ESC[nG` / CHA) captures correctly. psmux backfills those cells with spaces, so the bug is specific to CUF.

This bites TUI apps that position columns/words via relative cursor advance (Claude Code's renderer is one), whose captured pane text becomes hard to read because adjacent words collapse.

## Environment

- psmux version: 3.3.5 (Windows x64, winget install); changelog for 3.3.6 shows no related fix
- OS: Windows 11
- Shell: PowerShell 7.5.2

## Reproduction

From a detached session, emit three words separated by a 4-column cursor-forward advance rather than literal spaces, then capture:

```powershell
psmux new-session -d -s captest
# words separated by ESC[4C (cursor-forward 4), not spaces:
psmux send-keys -t captest '$e=[char]27; Write-Host ("WORDA${e}[4CWORDB${e}[4CWORDC${e}[4CWORDD")' Enter
psmux capture-pane -t captest -p
```

## Expected

The four columns skipped by each `ESC[4C` render as spaces (as they do on screen):

```
WORDA    WORDB    WORDC    WORDD
```

## Actual

The skipped cells are dropped entirely; the words collapse:

```
WORDAWORDBWORDCWORDD
```

## Contrast — absolute positioning works

The same layout via absolute-column moves (`ESC[nG`) captures correctly, which isolates the defect to the CUF path:

```powershell
psmux send-keys -t captest '$e=[char]27; Write-Host ("COLA${e}[10GCOLB${e}[20GCOLC")' Enter
psmux capture-pane -t captest -p
# -> COLA     COLB      COLC   (spaces preserved)
```

## Likely cause

A CUF advance appears to move the cursor without materializing the skipped cells as spaces in the grid. The capture serializer then emits only written cells, so interior never-written cells are omitted rather than rendered as spaces. Absolute positioning presumably backfills the gap, which is why it round-trips correctly.

## Related

This is distinct from #433 (scrollback not captured for full-screen-redraw TUIs), which is about history population. This issue is about the content of a single captured line being lossy for CUF-positioned text.

psmux:
[...]

PR #512, opened 2026-07-31
fix(#443): route every capture and copy path through the blank-cell backfill

acoliver:
Fixes the remainder of #443.

## Overview

de119a0 fixed the three `capture-pane` text paths (`-p`, ranged, `-e`), but the other serializers still used the inline `push_str(cell.contents())`, so a cell the cursor merely skipped over contributed nothing and adjacent words still fused. The boundary is *written vs never-written* cells rather than CUF vs CHA, so these collapsed on both:

| Path | Reached by |
| --- | --- |
| `capture_active_pane` | `capture-pane` with **no `-p`** — the paste-buffer variant of the command in the report, readable via `save-buffer` |
| `yank_selection` | copy-mode `y`, `M-w`, mouse drag-release — Char, Line **and** Rect |
| `copy_end_of_line` | copy-mode `D` |

- Extract `capture_row_text(screen, row, cols)` over the existing `push_capture_cell` and route those three plus the three already-fixed paths through it, so the space backfill and the wide-glyph continuation skip have exactly one implementation and cannot drift apart again.
- `search_copy_mode` is deliberately left alone: it indexes matches by grid column, so it wants one char per column *including* the wide continuation.

## Audit

I checked all 35 `cell.contents()` sites rather than only the ones named in the issue. Everything else is already correct and is unchanged: `format.rs` (the five row builders behind `#{C:pattern}`, `mouse_line`, `mouse_word`, `copy_cursor_word`, `copy_cursor_line`), `layout.rs` (four renderers plus two has-content checks), `rendering.rs` (`w == 0` emits a space), `popup.rs`, `window_ops.rs`, and `copy_mode`'s `read_row_text`.

Two observations left out of this PR on purpose:

- `util.rs:68,78` (`infer_title_from_prompt`) has the same unguarded pattern, but the function is dead code — no callers anywhere in `src/**` or `tests-rs/`. The `types.rs:119,131` comments referring to it look stale. Flagging rather than touching it.
- `#{cursor_character}` (`format.rs:1520`) yields `""` when the cursor sits on a never-written cell. It is a single-cell read, so it cannot fuse words; whether it should be `" "` is a behaviour call for you.

## Tests

`tests-rs/test_issue443_blank_cell_capture.rs` (registered from `copy_mode.rs`) drives the real functions over a PTY-backed pane tree — **no server and no session**, so it is safe to run on a live machine. All six behavioural tests reproduce the reported `CUFACUFBCUFC` collapse before the change.

The clipboard-touching tests serialize on a mutex and restore the prior clipboard: `yank_selection` and `copy_end_of_line` call `copy_to_system_clipboard`, and concurrent `GlobalLock` against another thread's `EmptyClipboard` corrupts the heap under the default multithreaded runner.

`tests/test_issue443_cuf_capture.ps1` gains `capture-pane` (no `-p`) + `save-buffer` coverage and is moved onto `New-PsmuxTestEnv` + a dedicated `-L` namespace.

## Validation

- `cargo test --bin psmux tests_issue443` — 14 passed
- `cargo test --bin psmux copy_mode::` — 30 passed
- `cargo test --bin psmux` for `issue244` (11), `named_buffers` (22), `issue413` (4), `issue210` (37), `issue155`, `vt100` — all passed
- `cargo check --all-targets` — clean
- `tests/test_issue443_cuf_capture.ps1` — 9/9 on a real session
- Confirmed the default psmux session list was unchanged before and after validation, and that no test session leaked into it.

## One thing to decide

The ps1 is additionally gated behind `PSMUX_ALLOW_DESTRUCTIVE_TESTS`, because `New-PsmuxTestEnv` alone is **not** sufficient to make it safe — see #510, which I hit while validating this branch: it terminated the live session I was working in. That gate is a workaround for #510, not part of this fix. If you would rather keep the two separate, drop the second commit (`test(#443): isolate the capture script...`) and this PR still stands on the first.

The gate auto-detects CI (`CI` / `GITHUB_ACTIONS`) rather than requiring the opt-in there, so the nightly `integration.yml` run still executes it. Requiring the opt-in in CI would report a green pass for a test that never ran.

---

### DIFF
--- a/src/copy_mode.rs
+++ b/src/copy_mode.rs
@@ -461,34 +461,23 @@ pub fn yank_selection(app: &mut AppState) -> io::Result<()> {
             match sel_mode {
                 crate::types::SelectionMode::Rect => {
                     let c0 = anchor.1.min(pos.1); let c1 = anchor.1.max(pos.1);
-                    let mut line = String::new();
-                    for c in c0..=c1 {
-                        if let Some(cell) = parser.screen().cell(r, c) { line.push_str(&cell.contents().to_string()); } else { line.push(' '); }
-                    }
+                    let line = capture_row_text(parser.screen(), r, c0..c1.saturating_add(1));
                     text.push_str(line.trim_end());
                     if !is_last { text.push('\n'); }
                 }
                 crate::types::SelectionMode::Line => {
-                    let mut line = String::new();
-                    for c in 0..cols {
-                        if let Some(cell) = parser.screen().cell(r, c) { line.push_str(&cell.contents().to_string()); } else { line.push(' '); }
-                    }
+                    let line = capture_row_text(parser.screen(), r, 0..cols);
                     text.push_str(line.trim_end());
                     text.push('\n');
                 }
                 crate::types::SelectionMode::Char => {
                     if total_lines == 1 {
                         let c0 = anchor.1.min(pos.1); let c1 = anchor.1.max(pos.1);
-                        for c in c0..=c1 {
-                            if let Some(cell) = parser.screen().cell(r, c) { text.push_str(&cell.contents().to_string()); } else { text.push(' '); }
-                        }
+                        text.push_str(&capture_row_text(parser.screen(), r, c0..c1.saturating_add(1)));
                     } else {
                         let line_start = if is_first { top_col } else { 0 };
                         let line_end   = if is_last  { bot_col } else { cols.saturating_sub(1) };
-                        let mut line = String::new();
-                        for c in line_start..=line_end {
-                            if let Some(cell) = parser.screen().cell(r, c) { line.push_str(&cell.contents().to_string()); } else { line.push(' '); }
-                        }
+                        let line = capture_row_text(parser.screen(), r, line_start..line_end.saturating_add(1));
                         text.push_str(line.trim_end());
                         if !is_last { text.push('\n'); }
                     }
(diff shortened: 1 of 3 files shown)


psmux:
Merged in 58f339f, thank you for finishing what de119a0 started and for auditing all 35 `cell.contents()` sites instead of just the ones named in the issue. That audit table is what let me merge this with confidence.

Before merging I reproduced the claim from the real user side rather than trusting it. One finding worth recording for anyone who reads this later: the collapse is reachable only from the client side. `psmux capture-pane` without `-p` from the CLI goes through the TCP server path, which de119a0 already fixed, so a CLI reproduction shows the gaps preserved and makes the bug look nonexistent. The proof needed real keystrokes into an attached window: with the payload script on screen, copy mode plus `?CUFA`, `V`, `y` yanked `CUFACUFBCUFCCUFD` on the pre merge build. WSL tmux 3.4 yanking the same line via `send-keys -X` preserves the gaps, so this was a genuine parity break in every copy path you listed.

Verified after the merge, same keystroke path: `y` and `D` both now produce `CUFA    CUFB    CUFC    CUFD`. Your 14 unit tests pass and the revised script passes 9 of 9 including the new buffer and save-buffer coverage.

Three follow up commits landed on top:

* **51d9b42** updates the safety gate header. Its rationale described the #510 machine wide reaper kill as live, but PR #514 merged here just before this one (c257249), so the hazard it names is fixed. The gate stays for two narrower reasons now stated in the header: anyone running the script against a pre #514 binary, and the ~10s redirected home self shutdown you documented. Your CI auto detect reasoning is kept verbatim, it was correct.
* **1d5833a** hardens the test harness. One full suite run saw two of the new buffer tests fail while every isolated run passes; the harness spawns a real ConPTY per test and this file added 14 more to the parallel pool, so creation can fail transiently under load. PTY creation now retries with backoff and reports which stage failed instead of a bare `.expect`. No product code involved.
* Full suite after all of the above: 2642 passed, zero failed, on repeated runs.

On your other open items: `util.rs` `infer_title_from_prompt` is confirmed dead (no callers) and will go in a housekeeping pass rather than this merge, and `#{cursor_character}` stays `""` for now since it cannot fuse words and its tmux behaviour deserves its own check.

The PTY backed unit tests plus the keystroke level proof against a real attached window make a good pair for this class of bug. Good work, and thanks again.

Issue #448, opened 2026-07-05
Harden server cleanup: reap live orphaned servers and store PID (not just port) in ~/.psmux


andrew-tawfeek:
Follow-up to the spawn-race fix in #444. Two structural gaps remain in server-lifecycle cleanup:

1. **No reaper for *live* orphaned servers.** `cleanup_stale_port_files` only deletes registry files for servers proven *dead* (connection refused / AUTH-rejected / pre-boot mtime). A live-but-orphaned server (a duplicate, or a crashed client's headless server) keeps running; only a manual `kill-server` reaps it.
2. **Registry stores the TCP port but no PID.** A wedged server that stopped listening but hasn't exited can't be targeted by identity at all.

**Suggestion:** also write `<base>.pid`, and add a startup pass that terminates same-name/duplicate server *processes* whose PID isn't the tracked one. That bounds the process count regardless of how a duplicate arose. #444 stops the main way duplicates were created; this cleans up any that still slip through.

Context: on a real machine this reached **28 servers / 62 pwsh / 93 conhost (18 orphaned)**, which exhausted Windows desktop-heap/handles and tore down every terminal session at once (no crash event).

---
*Reported by Claude (Anthropic) at @andrew-tawfeek's direction, from an audit of that incident.*

psmux:
Confirmed both gaps, and both are now fixed on master in b20bc77.

Rather than trust the audit at face value I reproduced each claim on a live build first.

**Gap 2 (no PID in registry):** verified instantly. A running session had `<base>.port`, `<base>.key`, and `<base>.sid` but no `<base>.pid` anywhere.

**Gap 1 (no reaper for live orphans):** reproduced tangibly. I took a live server, deleted its registry files so it became an untracked live process, then triggered a full startup cleanup pass. The orphan kept running and was now completely untargetable (no port file left to reach it), exactly the accumulation you saw in the incident.

**What changed**

1. Every server now writes `<base>.pid` (its OS process id) alongside the port file, in `ensure_session_registry_files`. It gets re anchored on rename and claim, re ensured every 5s, and torn down wherever the rest of the registry is (session teardown, stale cleanup, and the panic hook).

[...]

Issue #509, opened 2026-07-31
No stable per-namespace server identity: #{pid} tracks the newest session's server process, so creating a session looks like a server restart to supervisors


acoliver:
## Summary

On a `-L <namespace>` socket, `psmux` runs **one `psmux.exe server` process per session** (plus the internal `__warm__` helper), and `display-message -p '#{pid}'` resolves against the **most recently created session's** server process rather than a single namespace-level server. Creating a session therefore changes the value of `#{pid}` for the whole namespace.

There is consequently no stable identity a supervisor can use to answer *"is this the same psmux server I was talking to a moment ago, or did it restart?"* — which is the standard health check for a program that manages long-lived sessions. Under tmux, `#{pid}` is exactly that identity, because there is one server per socket.

This is a parity/observability gap rather than a crash. It is filed separately from #448 (which is about psmux's own cleanup bookkeeping storing port instead of PID) and from #485 (`display-message` not reporting active session data), because neither asks for a stable per-namespace server identity. It is likely the same underlying architecture as #459 (unbounded `psmux.exe` growth).

## Environment

- psmux `3.3.7 (05cc5d4 2026-07-20)`, winget `marlocarlo.psmux`
- Windows 11, PowerShell 7
- Sessions created detached on a private `-L` namespace

## Reproduction

```powershell
$ns = "repro$(Get-Random -Max 99999)"

psmux -f NUL -L $ns new-session -d -s alpha
psmux -f NUL -L $ns display-message -p '#{pid}'      # -> 20696

psmux -f NUL -L $ns new-session -d -s bravo
psmux -f NUL -L $ns display-message -p '#{pid}'      # -> 21228

psmux -f NUL -L $ns new-session -d -s charlie
psmux -f NUL -L $ns display-message -p '#{pid}'      # -> 20912
```

Three different values, one namespace, no server restart. The value is stable while no session is created (4 consecutive probes all returned `20912`), so this is not flapping — it tracks session creation.

`#{session_name}` shows what it is binding to:

```powershell
psmux -f NUL -L $ns display-message -p '#{pid}|#{session_name}'
# -> 20912|charlie      (the most recently created session, not the one I asked about)
```

And the process table shows why — there is no single server:

```
pid=19020  psmux.exe server -s __warm__ -L repro54593 -x 120 -y 30
pid=20696  psmux.exe server -s alpha    -L repro54593
pid=21228  psmux.exe server -s bravo    -L repro54593
[...]

PR #514, opened 2026-07-31
fix(#510): reap only servers this data dir can positively claim

### LINKED ISSUE #510: Startup orphan reaper terminates every psmux server on the machine when the invocation resolves a different data dir (-L gives no protection)

acoliver:
## Overview

Fixes #510. The startup reaper enumerates candidate servers machine-wide
(loopback listeners filtered by image name) but draws its authority from a
single `~/.psmux` resolved from `USERPROFILE`/`HOME`. Anything that registry
did not account for was classified as an orphan and terminated, so any
invocation resolving a different home reaped every other instance's live
servers.

The defect was the polarity: a server was reaped when we could not prove it
was ours, rather than only when we could prove it was.

## The fix

Servers now write an ownership marker into their own data dir
(`~/.psmux/servers/<pid>`), carrying the same `pid:creation_filetime` identity
gate as the `.pid` sentinel. `select_orphan_pids` honours a marker only when
the live process's creation time still matches, so an absent or stale marker
means "not ours" instead of "orphan".

The marker is deliberately independent of the per-session registry lifecycle,
because the states #448 exists to clean up - a spawn-race duplicate, a wiped
registry - are exactly the ones where `.port`/`.pid` are already gone. It is
reconciled against the process table on each pass rather than removed on
shutdown, since a server can exit down several routes and a marker missed by
any of them would linger forever. A lingering marker is inert: it names either
a dead PID or, after reuse, a process whose creation time no longer matches.

**Note on a weaker fix that was considered and rejected.** Refusing to reap
only when the registry is empty is not sufficient. A redirected-HOME harness
repopulates its registry the moment it creates its own session, which puts
every foreign server back in scope on the very next invocation. Attribution
has to be per candidate.

## Compatibility

A server started by an older build writes no marker, so it is never reaped.
That is the safe direction (under-kill, not over-kill), and it means orphan
reaping only resumes for servers started by a build containing this change.

## Validation

- 14 new unit tests (`tests-rs/test_issue510_reaper_attribution.rs`) covering
  the foreign-server rule, the identity gate, marker parsing, and both the
  harness and wiped-registry scenarios end to end over real files.
- The 11 existing #448 unit tests are updated, not weakened: every scenario
  keeps its original intent and now supplies the marker its server would have
  written. `orphan_with_no_registry_reference_is_reaped` still asserts an
  orphan is reaped.
- TDD: the first test failed against the pre-fix code with
  `a server this data dir cannot claim must be left alone, got [1000]`.
- Verified on a live machine that the exact operation from the issue - a
  read-only command under a throwaway home - now leaves all three live servers
  running, and that a real server does write its marker.
- `tests/test_issue510_foreign_server_survives.ps1`: 6/6 PASS.
- Confirmed the default psmux session list was unchanged, no registry files
  were left behind, and no pre-existing psmux server was terminated.

## One thing found while validating

psmux servers under a redirected `USERPROFILE`/`HOME` shut themselves down
after ~10s on Windows. This is pre-existing and unrelated to the reaper (it
reproduces with the reaper compiled out), but it is worth knowing: a test
built on `New-PsmuxTestEnv` cannot keep a session alive past that, which is
why the E2E here keeps its victim in the normal data dir under a unique `-L`
namespace and runs only the intruder under a throwaway home. Happy to file
this separately if it is news.

### DIFF
--- a/src/paths.rs
+++ b/src/paths.rs
@@ -104,6 +104,25 @@ pub fn port_file(session: impl AsRef<str>) -> String {
     format!("{}\\{}.port", psmux_dir(), session.as_ref())
 }
 
+/// Directory holding one ownership marker per server process this data dir
+/// started (issue #510).
+///
+/// The startup reaper enumerates candidate servers machine-wide, but its
+/// registry only describes one data dir. Without a per-process record of
+/// "this one is mine" it cannot tell an orphan of its own from another
+/// instance's healthy server, and killing on that ambiguity destroyed live
+/// sessions belonging to a different USERPROFILE/HOME. A server writes its
+/// marker into its OWN data dir, so ownership is self-declared and needs no
+/// inspection of other processes.
+pub fn server_marker_dir() -> String {
+    format!("{}\\servers", psmux_dir())
+}
+
+/// Path to one server process's ownership marker (see `server_marker_dir`).
+pub fn server_marker_file(pid: u32) -> String {
+    format!("{}\\{}", server_marker_dir(), pid)
+}
+
 /// Fallible variant of [`port_file`]: `None` when no data directory can be
 /// determined. Lets a call site early-exit on the same condition without having
 /// to know that `port_file` routes through `psmux_dir`.
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -127,6 +127,14 @@ fn ensure_session_registry_files(app: &AppState) {
         let _ = std::fs::write(&pid_path, &pid_value);
     }
 
+    // Claim this process for this data dir (issue #510). Keyed by PID and kept
+    // outside the per-session registry on purpose: the `.pid` entry above is
+    // removed with its session, but the reaper still needs to recognise a
+    // server as its own AFTER that happens - a spawn-race duplicate, or a
+    // registry wipe, is exactly the case it must clean up. Re-ensured here so
+    // the claim self-heals if the file is deleted underneath a live server.
+    crate::session::write_server_marker(self_pid);
+
     // .port goes LAST: it is the readiness beacon (see comment above).
     if std::fs::read_to_string(&port_path)
         .map(|s| s.trim() != port_value)
--- a/src/session.rs
+++ b/src/session.rs
@@ -177,6 +177,46 @@ pub fn remove_session_pid_file(port_file_base: &str) {
     let _ = std::fs::remove_file(&pid_path);
 }
 
+/// Record that server process `pid` belongs to THIS data dir (issue #510).
+///
+/// Keyed by PID rather than by session, and deliberately independent of the
+/// session registry lifecycle: the whole point is to still identify a server as
+/// ours after its `.port`/`.pid` entries are gone, which is exactly the state a
+/// spawn-race duplicate or a registry wipe leaves behind. Body is the same
+/// `pid:creation_filetime` as the `.pid` sentinel so a reused PID cannot
+/// inherit a dead process's claim.
+pub fn write_server_marker(pid: u32) {
+    let dir = crate::paths::server_marker_dir();
+    if std::fs::create_dir_all(&dir).is_err() {
+        return;
+    }
+    let creation = crate::platform::process_kill::process_creation_time(pid).unwrap_or(0);
+    let _ = std::fs::write(
+        crate::paths::server_marker_file(pid),
+        format_pid_file_contents(pid, creation),
+    );
+}
+
+/// Server processes `psmux_dir` claims, as `pid -> recorded creation filetime`.
+///
+/// The PID is read from the file body rather than its name so a truncated or
+/// hand-copied marker cannot assert a claim over an arbitrary PID.
+pub fn read_owned_server_pids(psmux_dir: &Path) -> std::collections::HashMap<u32, Option<u64>> {
+    let mut owned = std::collections::HashMap::new();
+    let Ok(entries) = std::fs::read_dir(psmux_dir.join("servers")) else {
+        return owned;
+    };
+    for entry in entries.flatten() {
+        let Ok(body) = std::fs::read_to_string(entry.path()) else {
+            continue;
+        };
+        if let Some((pid, creation)) = parse_pid_file_contents(&body) {
+            owned.insert(pid, creation);
+        }
+    }
+    owned
+}
+
 /// Resolve a tmux session ID (`$N`) to the port file base name of the
 /// session that owns that ID. Returns `None` if no session has that ID.
 pub fn resolve_session_by_id(id: usize) -> Option<String> {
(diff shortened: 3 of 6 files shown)


psmux:
Merged in c257249, with the E2E strengthened in 6f8288f (marker written on disk by a real server, and pruned after death, with the prune check waiting for full process exit so it does not race `kill-server` teardown).

This was an excellent PR for a genuinely dangerous defect. The polarity inversion is exactly right, the identity gate on the marker closes the pid reuse hole, and the note explaining why the empty registry guard is insufficient saved me from suggesting it. I reproduced the incident on master before merging (one read only probe under a throwaway home killed an aged victim on another `-L` socket) and verified on the merged build that the victim survives repeated foreign passes while a registry wiped server of our own is still reaped. Full suite 2628 passed, zero failed.

Note that the merge needed a small conflict resolution against the #509 work that landed just before it, since both add a re-ensured block to `ensure_session_registry_files`. Both blocks are kept.

Longer response on the remaining layering questions in the issue thread.

Issue #485, opened 2026-07-22
display-message doesn't display active session data


pantaryl:
## Summary

`display-message -p '#S'` (and other variables) reports information for the last session/pane created or attached, instead of the active session. This makes identification of current target (session:window.pane) extremely difficult.

## Environment

- **psmux version:** 3.3.7 (05cc5d4 2026-07-20), chocolatey, windows-x64
- **OS:** Windows 11 (10.0.26200)
- **Shell:** PowerShell 7

## Steps to Reproduce

```powershell
# terminal 1
psmux new-session -s repro
tmux display-message -p '#S'    # should show 'repro'

# terminal 2
psmux new-session -s repro_other
tmux display-message -p '#S'    # should show 'repro_other'

# now go back to terminal 1, which is still attached to the 'repro' session
tmux display-message -p '#S'    # should show 'repro', but it shows 'repro_other'
Prefix + d # to detach
tmux attach -t repro
tmux display-message -p '#S'    # should show 'repro', does
```

## Expected (tmux behavior)

- My understanding based on the articles I've read is that calling `tmux display-message -p '*'` from within an attached session should return variables for that session, not the last one that was created or attached to.
- I am not positive at this time whether doing `Prefix + :` and then `display-message -p '#S'` shows the expected value.

## Impact

Trying to write code to uniquely identify the current session doesn't work, and then trying to get identifying info for targeting other sessions is also difficult. We could theoretically parse the `TMUX` environment variable and then parse the pid/sid files, but that seems more complicated than necessary.

psmux:
Thanks for the really clear repro, @pantaryl. You nailed it, and it turned out to be a genuine bug. Fixed now in 061ac56.

## What was happening

I reproduced it exactly as you described. With two sessions alive, running `tmux display-message -p '#S'` from inside the pane of the first one reported the second (last created) session instead of the current one.

The cause is on our side and specific to how psmux starts panes fast. psmux keeps a small pool of pre started "warm" shells and transplants one into your session when you create it, so new sessions come up instantly. Those warm shells are spawned by an internal `__warm__` server and inherit an environment variable, `PSMUX_TARGET_SESSION=__warm__`, that names the server they were born in. Once a shell is running, its environment block is frozen (there is no way to rewrite a live process's env on Windows), so that value sticks around inside your real pane after the transplant.

[...]

PR #523, opened 2026-08-01
fix(#509): give each -L namespace a stable server identity

### LINKED ISSUE #509: No stable per-namespace server identity: #{pid} tracks the newest session's server process, so creating a session looks like a server restart to supervisors

acoliver:
## Overview

Fixes #509.

psmux runs one `psmux.exe server` process per session, so `#{pid}` — and its
alias `#{server_pid}` — resolve to whichever session's server answered the
request. Creating a session changes the value for the whole namespace, and a
supervisor following the tmux idiom reads that as *"the server restarted"*,
marking every already-running session as having lost its server.

This adds `#{server_instance}`, a namespace-scoped identity:

- **Minted** by the first server in a `-L` namespace, stored under
  `<data dir>/instances/`.
- **Read, never re-minted,** by every later server in that namespace — so the
  reported value does not depend on which server replied.
- **Re-minted only when no peer is still alive,** so a genuine restart remains
  visible to a watchdog.

Peer liveness reuses the existing `.pid` identity gate (`confirms_identity`): a
pid whose recorded creation time no longer matches has been recycled and is
somebody else's process, so it cannot keep a dead namespace's token alive.

## Design notes

**Reclamation is self-healing, not explicit.** A server can exit via
`exit-empty`, `kill-session`, or a crash, so nothing reliably runs on the way
out. Rather than deleting the token on exit, it is replaced by the next server
that finds no live peer.

**No pruning pass, deliberately.** Wrongly deleting a *live* namespace's token
would cause the next re-ensure to mint a fresh one and signal a restart that
never happened — the exact defect being fixed. A stale token is a few bytes and
is reused whenever that namespace name is used again; accumulation is bounded by
distinct namespace names, exactly as it already is for `.port`/`.key`/`.sid`/`.pid`.

**`#{pid}` / `#{server_pid}` are unchanged.** They correctly report the
answering server, and changing them would break anyone relying on current
semantics. The docs no longer describe them as "PID of the psmux server", state
that they are session-scoped, and point supervisors at `#{server_instance}`.

**Out of scope:** making unattached `display-message` resolve against the
namespace rather than the newest session (direction 2 in the issue). That
overlaps #485 and changes existing behaviour, so it is left as a separate call
for the maintainer.

**Filename isolation:** `-L` values come from the user and may contain
characters that are illegal in a filename, or that collide once sanitised
(`a/b` and `a_b` both become `a_b`). The identity file name is a readable
prefix plus a fixed-seed hash of the *full* namespace name, so isolation does
not depend on the prefix.

## Validation

- **The issue's own three-session reproduction**, against the built binary:

  ```
  after creating alpha    #{pid}=24892   #{server_instance}=b644f0a347fa5e14
  after creating bravo    #{pid}=16088   #{server_instance}=b644f0a347fa5e14
  after creating charlie  #{pid}=25344   #{server_instance}=b644f0a347fa5e14
  ```

  `#{pid}` still varies (behaviour preserved); `#{server_instance}` holds constant.

- **21 new unit tests** (`tests-rs/test_issue509_namespace_instance.rs`) covering
  namespace membership (including each namespace's warm helper, and the default
  namespace not adopting `-L` entries), the first-server decision under dead and
  recycled pids, mint-once/keep semantics, re-mint after a genuine restart,
  namespace independence, and that reading never mints. Peer liveness is
  injected, so there is no OS process enumeration and the tests are
  deterministic and cross-platform.

- **7 end-to-end checks** (`tests/test_issue509_namespace_instance.ps1`): the
  reproduction, a real kill-and-restart producing a new token, two namespaces
  not sharing an identity, and an unknown namespace reporting nothing. Runs in a
  unique `-L` namespace and tears down only what it created — never a bare
  `kill-server`.

- **Full unit suite: 2589 passed, 0 failed.** `cargo check --all-targets` clean.

- Confirmed the default psmux session list was unchanged, and that no server
  outside the test namespace was affected.

---

### DIFF
--- a/docs/integration.md
+++ b/docs/integration.md
@@ -413,9 +413,22 @@ row per object. The full catalogue, including the human facing status bar variab
 | `#{client_pid}` | `32944` | PID of the attached client |
 | `#{client_key_table}` | `root` | Key table the client is currently in |
 | `#{version}` | `3.3.7` | psmux version, for capability gating |
-| `#{pid}` / `#{server_pid}` | `19004` | PID of the psmux server |
+| `#{pid}` / `#{server_pid}` | `19004` | PID of the server process that answered — **session-scoped**, see below |
+| `#{server_instance}` | `b644f0a347fa5e14` | Stable identity of the `-L` namespace — poll this to detect a real restart |
 | `#{socket_path}` | `C:\Users\me/.psmux/default` | Server discovery path |
 | `#{host}` / `#{host_short}` / `#{user}` | `BOX` / `me` | Host and user identity |
+
+> **Supervising a namespace.** Unlike tmux, psmux runs one server process per
+> session, so `#{pid}` (and its alias `#{server_pid}`) report whichever session's
+> server handled the request — creating a session changes the value even though
+> nothing restarted. A watchdog that polls `#{pid}` to answer *"is this still the
+> server I was talking to?"* will read every new session as a server restart.
+>
+> Poll `#{server_instance}` instead. It is minted by the first server in a `-L`
+> namespace, reported identically by every server in that namespace, and changes
+> only when the namespace has genuinely gone away and come back. An unknown or
+> not-yet-started namespace reports an empty value, which should be treated as
+> *unknown* rather than as a restart.
 
 Any option name also resolves inside `#{...}`, which is usually cheaper and more reliable than
 parsing `show-options` output:
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -417,7 +417,8 @@ bind-key -n C-h if-shell -F "#{pane_at_left}" "send-keys C-h" "select-pane -L"
 | `#{host}` / `#{hostname}` | Full hostname |
 | `#{host_short}` | Hostname up to the first dot |
 | `#{user}` / `#{username}` | Current user name |
-| `#{pid}` / `#{server_pid}` | PID of the psmux server |
+| `#{pid}` / `#{server_pid}` | PID of the server process that answered the request. psmux runs one server per session, so this is session-scoped and changes when a session is created — use `#{server_instance}` to identify the namespace |
+| `#{server_instance}` | Stable identity of the `-L` namespace. Constant while the namespace is up, whichever of its servers answers; changes only after a genuine restart. Empty for a namespace that has no server |
 | `#{version}` | psmux version, for example `3.3.7` |
 | `#{start_time}` | Server start time |
 | `#{socket_path}` | Path of the server's discovery files |
--- a/src/format.rs
+++ b/src/format.rs
@@ -1150,6 +1150,10 @@ pub fn expand_var(var: &str, app: &AppState, win_idx: usize) -> String {
                     .map(|d| d.to_string_lossy().into_owned())
                     .unwrap_or_default(),
                 "pid" | "server_pid" => std::process::id().to_string(),
+                "server_instance" => {
+                    crate::session::read_namespace_instance(app.socket_name.as_deref())
+                        .unwrap_or_default()
+                }
                 "version" => VERSION.to_string(),
                 "host" | "hostname" => hostname_cached(),
                 "host_short" => { let h = hostname_cached(); h.split('.').next().unwrap_or(&h).to_string() }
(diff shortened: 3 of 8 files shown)


psmux:
Merged in 794d4c0, thank you for a genuinely excellent contribution. The token design, the reuse of the `.pid` identity gate for peer liveness, and the hashed filename isolation were all sound, and the tests were a pleasure to review.

One follow up landed on top in 8b0ba69: the ensure also runs inside the 5 second registry self heal loop, and rerunning the startup first server decision there made a lone server (no live peers, for example a single session under `PSMUX_NO_WARM=1`) delete and remint its own token on every tick. I measured 4 distinct tokens in 16 seconds on this branch before the follow up. The fix caches the established token per server process and turns every later ensure into a pure restore, which also means a lost token file is recreated with the same identity instead of a fresh mint. Full details are in the issue thread and the commit message.

Issue #490, opened 2026-07-22
white spaces are not preserved when using psmux send-keys


fekir:
~~~~
psmux send-keys 'echo "a           a"' C-m;
~~~~

behaves as-if one executes

~~~~
psmux send-keys 'echo "a a"' C-m;
~~~~

and 

~~~~
psmux send-keys " cd <path>" C-m;
~~~~

behaves as-if one executes

~~~~
psmux send-keys "cd <path>" C-m;
~~~~

Since different shells can be configured not to store in the history commands prefixed by whitespaces, the second example changes what is store in the history file.

psmux:
Thanks, both examples reproduced exactly as you described, including the shell history angle with the leading space, which is a really good reason to get this byte exact.

Root cause: internally the `send-keys` arguments were joined with single spaces and then resplit on whitespace on the server side, so any run of spaces inside a quoted argument collapsed to one space and leading and trailing spaces vanished entirely.

Fixed in b6c96fd, refined in 6231fa0. The arguments now travel as separate tokens end to end: each argument is either matched as a named key (`Enter`, `C-m`, `Up`, ...) in its entirety or typed verbatim with every space intact. Verified with a `Read-Host` probe that echoes back exactly what the shell received:

* `psmux send-keys 'echo "a           a"' C-m` arrives with all eleven spaces,
* `psmux send-keys " cd <path>" C-m` keeps the leading space, so your history exclusion works, and
* trailing spaces and `send-keys -l` are byte exact too.

One small alignment with tmux while I was in there: a key name inside a single quoted argument (for example `send-keys "ls Enter"`) is now literal text, as it is in tmux. Pass the key as its own argument (`send-keys "ls" Enter`), which is what tmux expects as well.

On master now. Appreciate the precise examples!

PR #524, opened 2026-08-01
fix(send-keys): implement -H literal-byte injection

krhougs:
## Context

tmux's `send-keys -H` is a **byte** channel: each operand is one hexadecimal
byte value that reaches the pty verbatim. tmux flags those keys `KEYC_LITERAL`
and `input_key.c` writes them out as single bytes ("Literal keys go as
themselves (can't be more than 8 bits)"). The man page wording "hexadecimal
number representing a Unicode codepoint" does not describe what the code does.

Confirmed against real tmux 3.7b:

```console
$ tmux -L probe new-session -d -s hex 'cat > out.bin'
$ tmux -L probe send-keys -H -t hex 41 e4 b8 ad 0a
$ xxd out.bin
00000000: 41e4 b8ad 0a                             A....
```

Five operands in, exactly those five bytes out.

## Current behaviour

psmux accepts `-H` but never gives it a meaning, so the hex operands are echoed
into the pane as literal text. Typing `echo` through a client that uses `-H`
puts `65 63 68 6f` on screen instead.

```console
$ psmux new-session -d -s demo
$ psmux send-keys -H -t demo 65 63 68 6f
$ psmux capture-pane -p -t demo | tail -1
PS C:\Users\me> 65 63 68 6f
```

The same happens through control mode, which is how gateway-style clients
drive psmux:

```console
$ printf 'send-keys -H -t %%1 65 63 68 6f\n' | psmux -C attach-session -t demo
```

This is not limited to one client. **iTerm2 3.7 routes every C0 control byte
(0x00–0x1F) through `send -H`** when the server reports 3.0a or newer — see
`iTermTmuxSendKeysEncodingLiteralByte` in its `TmuxGateway.m`, and the 3.7
release notes ("Control bytes 0x00–0x1F are now sent through send-keys with
the -H flag on tmux 3.0a or newer"). psmux reports `tmux 3.3.7`, so it takes
that path, and Enter, Tab, Escape and every Ctrl chord arrive as hex text.
`docs/iterm2-control-mode.md` predates that iTerm2 change.

## Root cause

The flag is dropped at three separate layers, and the operands then fall
through to the key-name lookup:

| Layer | Behaviour |
|---|---|
| `main.rs` CLI parsing | getopt handles `-l/-R/-X/-t/-N`; `-H` hits the catch-all and is pushed onto the key list |
| `server/connection.rs` dispatchers | both the control-mode and plain-TCP `send-keys` arms filter out every `-`-prefixed argument, so `-H` disappears |
| `server/mod.rs` key dispatch | bare `41` matches no key name, so it lands in `send_text_to_active` and is written to the pane as text |

`decode_send_command` also refuses to coalesce these commands, because bare hex
is neither a `0x`-prefixed operand nor literal-mode text.

## Fix

Decode the operands where the flag is parsed and carry them to the pane as
bytes through a new `CtrlReq::SendBytes`, in all three entry points.

Valid UTF-8 still goes through `send_text_to_active`, so copy mode, popups,
menus and synchronized input keep their existing behaviour. Only a byte run
that is not valid UTF-8 — which happens when a caller splits a multi-byte
character across two chunked `send-keys` calls — is written straight to the
pane, since reassembling it there is not possible and a lossy decode would
corrupt the input.

Send-coalescing has to understand the flag too. iTerm2 splits an arrow key
into three differently-encoded sub-commands on one line
(`send -H … 1b`, `send -t … 0x5b`, `send -lt … A`), and it is the coalescing
that merges them into a single pty write so PSReadLine does not time out
between the ESC and the `[A`. `decode_send_command` now decodes `-H` operands,
and the merged line is re-emitted as `send -H <hex>` instead of
`send -l '<latin-1>'`.

That carrier change fixes a second, pre-existing bug. The old quoting mapped
each accumulated byte through `as char`, so anything the decoder had already
encoded to UTF-8 was encoded a second time: `send -t %1 0x4e2d` (中) arrived at
the pane as `ä¸­`. With a byte-exact carrier the round trip is gone, and `0xNN`
can consistently mean a codepoint while raw bytes arrive only through `-H`.

## Verification

Driven through control mode against a real pwsh 7.6.4 pane, the same way a
gateway client does it (`psmux -C attach-session` with commands fed to stdin):

| Case | Before | After |
|---|---|---|
| `send-keys -H … 65 63 68 6f` | `65 63 68 6f` | `echo` |
| `send-keys -H … e4 b8 ad` (中) | `e4 b8 ad` | `中` |
| `send -lt %1 abc` | `abc` | `abc` |
| `send -t %1 0xe9` (é) | `é` | `é` |
| `send -t %1 0x4e2d` (中) | `ä¸­` | `中` |
| iTerm2 arrow key, three sub-commands merged | `echo MARKER0d1b[A` | recalls `echo MARKER` |
| `send-keys -lt %1 "a     b"` (#490) | `a     b` | `a     b` |

The last two rows are the regression checks that matter here: the existing
coalescing behaviour and the #490 whitespace fix both still hold.

`tests-rs/test_send_keys_literal_byte.rs` pins the two pure functions the fix
rests on -- `decode_send_command` and `coalesce_send_commands` -- covering byte
decoding, the `-H` (byte) versus `0xNN` (codepoint) distinction, non-UTF-8
operands, the merged `send -H` carrier, target separation and malformed
operands. They start no server, per AGENTS.md.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -2908,6 +2908,47 @@ pub fn send_paste_to_active(app: &mut AppState, text: &str) -> io::Result<()> {
     Ok(())
 }
 
+/// Write raw bytes (decoded from `send-keys -H`) to whichever pane would
+/// receive input.
+///
+/// Valid UTF-8 is handed to `send_text_to_active` so every existing mode
+/// (copy, popup, confirm, menu, synchronized input) keeps its semantics.  A
+/// byte run that is not valid UTF-8 bypasses that and goes straight to the
+/// pane: callers that chunk their input can split a multi-byte character
+/// across two `send-keys` calls, and a lossy decode would corrupt it.
+pub fn send_bytes_to_active(app: &mut AppState, bytes: &[u8]) -> io::Result<()> {
+    if let Ok(text) = std::str::from_utf8(bytes) {
+        return send_text_to_active(app, text);
+    }
+    {
+        let win = &mut app.windows[app.active_idx];
+        if let Some(fi) = win.floating_focus {
+            if let Some(fp) = win.floating.get_mut(fi) {
+                let _ = fp.pane.writer.write_all(bytes);
+                let _ = fp.pane.writer.flush();
+                return Ok(());
+            }
+        }
+    }
+    if app.sync_input {
+        let win = &mut app.windows[app.active_idx];
+        fn write_all_panes(node: &mut Node, data: &[u8]) {
+            match node {
+                Node::Leaf(p) => { let _ = p.writer.write_all(data); let _ = p.writer.flush(); }
+                Node::Split { children, .. } => { for c in children { write_all_panes(c, data); } }
+            }
+        }
+        write_all_panes(&mut win.root, bytes);
+    } else {
+        let win = &mut app.windows[app.active_idx];
+        if let Some(p) = active_pane_mut(&mut win.root, &win.active_path) {
+            let _ = p.writer.write_all(bytes);
+            let _ = p.writer.flush();
+        }
+    }
+    Ok(())
+}
+
 pub fn send_text_to_active(app: &mut AppState, text: &str) -> io::Result<()> {
     // In clock mode, any input exits back to passthrough
     if matches!(app.mode, Mode::ClockMode) {
--- a/src/main.rs
+++ b/src/main.rs
@@ -1631,14 +1631,16 @@ fn run_main() -> io::Result<()> {
             "send-keys" | "send" | "send-key" => {
                 let mut literal = false;
                 let mut has_x = false;
+                let mut has_hex = false;
                 let mut keys: Vec<String> = Vec::new();
-                // Getopt-style parsing: -t consumes next arg, -l/-R/-X are flags
+                // Getopt-style parsing: -t consumes next arg, -l/-R/-X/-H are flags
                 let mut i = 1;
                 while i < cmd_args.len() {
                     match cmd_args[i].as_str() {
                         "-l" => { literal = true; }
                         "-R" => { keys.push("__RESET__".to_string()); }
                         "-X" => { has_x = true; }
+                        "-H" => { has_hex = true; }
                         "-t" => { i += 1; } // consume target value (already handled globally)
                         "-N" => { i += 1; } // repeat count, consume value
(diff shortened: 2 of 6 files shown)


psmux:
Merged. Thank you @krhougs, this is one of the best researched PRs this project has received.

Before touching the diff I reproduced every claim independently, and each one held up exactly as you described:

**The bug, confirmed on master**
* `psmux send-keys -H -t demo 65 63 68 6f` echoed the literal text `65 63 68 6f` into the pane.
* Real tmux 3.4 under WSL with the same operands delivered exactly the raw bytes `41 e4 b8 ad 0a` into `cat > out.bin`, byte for byte what you showed against 3.7b.
* Control mode `send -t %1 0x4e2d` arrived as `ä¸­` on master, your double encoding find was spot on. That one was a real preexisting bug hiding under the coalescing carrier and I am glad you caught it rather than papering over it.

**Your source citations checked out too.** I verified against the tmux tree: `input-keys.c` writes `KEYC_LITERAL` keys as single bytes with the exact comment you quoted, `cmd-send-keys.c` parses `-H` operands base 16 capped at `0xff`, and `key-string.c` treats `0xNN` as a codepoint through `wctomb`. Your read that the man page wording does not match the code is correct.

**Verification on your branch after merging master into it**
1. All seven of my end to end checks pass: CLI `-H` types `echo`, a `-H` byte run `e4 b8 ad` arrives as one CJK character, control mode `-H` works, `0x4e2d` now arrives as that same character, the three part iTerm2 arrow key coalesces into a history recall, the issue 490 space run regression guard holds, and a malformed operand is dropped without leaking into the pane.
2. Full Rust suite: 2680 passed, 0 failed.
3. Targeted suites for issues 490, 200 and 409 all green.

The distinction you drew between `-H` as a byte channel and `0xNN` as a codepoint channel, and rebuilding the coalescing carrier as pure ASCII `send -H`, is exactly the right design. The unit tests pinning `decode_send_command` and `coalesce_send_commands` are much appreciated and are now part of the suite.

Merged in 0888fd0. Thanks again for the depth here, from the iTerm2 gateway analysis down to the tmux source. Contributions like this are a pleasure to review.

PR #525, opened 2026-08-02
fix(input): dispatch copy-mode key tables on the live client/server path

Gerrrt:
# input

## Problem

`bind -T copy-mode-vi` / `-T copy-mode` bindings parsed, were stored in key_tables, and showed up in list-keys. However, nothing on the live path ever consulted them. The only lookup lived in `input::handle_key`, which has no callers (dead code from the pre-server architecture). The live path hardcodes copy-mode keys in `send_key_to_active` and `handle_copy_mode_char`.

The failure was near-invisible. The common rebinds (`v` begin-selection, `y` copy, `Escape` cancel) coincide with the hardcoded vi defaults, so the keys did something reasonable. What silently did not happen was the user's own action. A config that looks like:

`bind -T copy-mode-vi y send-keys -X copy-pipe-and-cancel "clip.exe"`

the selection was yanked to the internal buffer but never piped to clip.exe.

## Fix

- Extract the table lookup into copy_mode_binding / run_copy_mode_binding and call them from both live entry points before the built-in defaults so a bound key wins and an unbound key falls through unchanged.
- `send-keys -X` is queued back onto the server loop via a new AppState.control_tx (the copy-mode command implementation lives in the loop's own SendKeysX arm. Re-entering it or routing back out through a self-connection would deadlock). The command is rebuilt from the original string so a quoted argument with spaces survives.

## Tests

tests-rs/test_copy_mode_key_tables.rs covers table selection by `mode-keys`, key normalization (Shift/Ctrl), that `send-keys -X` is queued with arguments intact, and the fall-through path for unbound keys.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -51,6 +51,76 @@ fn write_mouse_event(master: &mut dyn std::io::Write, button: u8, col: u16, row:
     }
 }
 
+/// Look up a copy-mode key binding for `key`, honouring `mode-keys`.
+///
+/// The table name selection matches tmux: `mode-keys vi` uses `copy-mode-vi`,
+/// anything else uses `copy-mode`.
+pub fn copy_mode_binding(app: &AppState, key: (KeyCode, KeyModifiers)) -> Option<crate::types::Action> {
+    let table_name = if app.mode_keys == "vi" { "copy-mode-vi" } else { "copy-mode" };
+    let key_tuple = normalize_key_for_binding(key);
+    app.key_tables
+        .get(table_name)?
+        .iter()
+        .find(|b| b.key == key_tuple)
+        .map(|b| b.action.clone())
+}
+
+/// Run a resolved copy-mode binding.
+///
+/// Returns `true` when the binding was handled and the caller must NOT fall
+/// through to the built-in copy-mode key handling.
+///
+/// `send-keys -X <cmd>` is by far the most common action in these tables (it is
+/// how tmux-yank and every copy-mode rebind is written), and its implementation
+/// lives in the server loop's own `CtrlReq::SendKeysX` arm. Since this function
+/// is itself called from the loop, it queues the request rather than trying to
+/// re-enter that arm — see `AppState::control_tx`. Everything else goes through
+/// the normal action executor.
+pub fn run_copy_mode_binding(app: &mut AppState, action: &crate::types::Action) -> bool {
+    use crate::types::Action;
+    if let Action::Command(cmd) = action {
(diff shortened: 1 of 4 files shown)


psmux:
Merged. Thank you @Gerrrt, this was a sharp piece of archaeology.

You found something genuinely nasty: bindings that parse, store, and show up in `list-keys` while the live path never consults them, with the failure masked because the popular rebinds coincide with the hardcoded vi defaults. I confirmed it exactly as you described before reading a line of the diff:

**Reproduced on master**
* With `bind -T copy-mode-vi y send-keys -X copy-pipe-and-cancel "..."` registered and visible in `list-keys`, a real injected `y` keystroke (WriteConsoleInput into an attached client, not just CLI `send-keys`) performed the builtin yank into the internal buffer while the pipe command never ran. Silent, near invisible, precisely your clip.exe scenario.
* Direct `send-keys -X copy-pipe-and-cancel` from the CLI worked the whole time, which is what made the binding gap so hard to notice.

Your diagnosis of `input::handle_key` as dead code with the only table lookup in it matched my black box results, and the `control_tx` queueing design to avoid reentering the server loop is clean. Keeping the dead function annotated instead of bundling a 600 line deletion into this fix was the right call, as is the heads up on `switch-client -T`.

**One adjustment I made while verifying, commit b288372.** Rebuilding the queued command from the original bind string preserved the quote characters, and the `SendKeysX` arm hands everything after the copy mode command name to `pwsh -Command` verbatim. So for a pipe command with spaces, like `copy-pipe-and-cancel "cmd /c more > file"`, pwsh received a quoted string literal, evaluated it, and discarded it: the cancel half fired but the pipe never spawned. The TCP dispatcher never hit this because it joins tokens whose quote grouping was already stripped. I changed `run_copy_mode_binding` to mirror that dispatcher exactly (tokenize, drop flags and their operands, join), added a unit assertion that no quote character reaches the `SendKeysX` queue, and proved the end result with an injected keystroke piping the real selection text into a marker file. Your dispatch fix was correct; only the carrier needed to match what the arm consumes.

**Verification after the adjustment**
1. The injected `y` keystroke now runs the user's pipe with the selection on stdin, and real tmux 3.4 under WSL agrees on the semantics.
2. Bindings to non `send-keys` actions fire too (`Y` to `new-window`), unbound keys fall through to the builtins unchanged.
3. Full Rust suite: 2680 passed, 0 failed, including your ten table tests. Copy mode end to end suites all green (23, 14, 7 and 3 checks).

Merged in f86cec1 with your commit intact. Thank you for the careful writeup and for fixing the root cause instead of the symptom.

PR #526, opened 2026-08-02
fix(format): a modifier over an empty variable renders its value, not its name

Gerrrt:
# format

## Problem 

`expand_var` returns "" for two different cases: "this is not a variable" and "this is a variable whose value is currently empty." `expand_var_or_format` treated any empty result as the former and fell back to echoing the target as a literal. So a modifier over an `optional-but-real` variable printed the variable's own name:

```sh
#{b:pane_path}   →   "pane_path"     (when the shell hasn't sent an OSC 7 yet)
#{pane_path}     →   ""              (correct)
```

The bare form was always right, so the bug only surfaced once a modifier (`b:`, `d:`, `q:`, `=N:`, etc.) was involved and it failed in the worst way: not blank, not an error, but plausible-looking text sitting in the status bar. It bites anyone using `#{b:pane_path}` for a cwd segment before the first prompt has announced its path.

## Fix

Distinguish the two cases with an internal UNKNOWN_VAR sentinel (it contains a NUL, so it can never collide with real output):

- `expand_var_inner` returns the sentinel for a name that is not a variable.
- The public `expand_var` maps the sentinel back to "", preserving tmux parity for callers that expect unknown → empty.
- `expand_var_or_format` checks for the sentinel instead of `is_empty()`, so an `empty-but-real` variable flows through the modifier and only a genuinely unknown name falls back to the literal echo (which stays intentional. it's how `#{=3:some literal}-style` usage degrades).

## Tests

tests-rs/test_modifier_over_empty_var.rs covers the reported `#{b:pane_path}` case, the other modifiers over an empty variable, a modifier over a populated variable (still works), the still-intended literal fallback for genuinely unknown names, and that the sentinel never leaks into output.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -515,7 +515,7 @@ fn expand_expression(expr: &str, app: &AppState, win_idx: usize) -> String {
         return result;
     }
 
-    // Plain variable or option name
+    // Plain variable or option name. Unknown names render empty (tmux parity).
     expand_var(expr, app, win_idx)
 }
 
@@ -919,8 +919,14 @@ fn expand_var_or_format(target: &str, app: &AppState, win_idx: usize) -> String 
         if target.is_empty() || target.parse::<f64>().is_ok() {
             return target.to_string();
         }
-        let val = expand_var(target, app, win_idx);
-        if val.is_empty() && !target.is_empty() {
+        // Use the inner form so a REAL variable that is merely empty is
+        // distinguishable from a name that is not a variable at all. Testing
+        // `val.is_empty()` conflated them, so `#{b:pane_path}` rendered the
+        // literal text "pane_path" whenever no OSC 7 had arrived yet — every
+        // modifier over an optional variable echoed its own name instead of
+        // rendering nothing.
+        let val = expand_var_inner(target, app, win_idx);
+        if val == UNKNOWN_VAR {
             // Try as option
             if let Some(opt_val) = lookup_option(target, app) {
                 return opt_val;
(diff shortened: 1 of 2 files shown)


psmux:
Merged. Thank you Gerrrt, this is another really sharp catch and a beautifully scoped fix. Three quality PRs in a row now, and this one shows the same care as the last two.

Before touching your branch I reproduced the bug on current master exactly as you described, using a cmd.exe pane so no OSC 7 ever arrives and `pane_path` stays genuinely empty:

* `#{pane_path}` rendered empty (correct)
* `#{b:pane_path}` rendered the text `pane_path`
* `#{q:pane_path}` rendered `pane_path`
* `#{=3:pane_path}` rendered `pan`, a truncated variable name in the status bar
* and the one that convinced me this mattered even more than the report suggested: `#{?#{b:pane_path},HAS,NONE}` evaluated to `HAS`, so conditionals over a modified empty variable took the wrong branch entirely

I also verified parity against a live tmux 3.4: a modifier over an empty but real variable renders empty there, exactly what your sentinel gives us. One honest note for the record: tmux renders even genuinely unknown names empty, so the literal echo we keep for unknown names is a deliberate psmux extension. Your fix narrows that divergence to only the truly unknown case, which I think is the right call for now, and your test pins the boundary so it stays intentional.

Verification on my side:

* your 9 new Rust tests pass, and the full suite is green with 2689 passed and 0 failed
* I added `tests/test_pr526_modifier_empty_var.ps1`, an end to end suite covering the CLI path, the raw TCP server path, edge cases, a status bar cwd pill via `status-right`, and a live attached TUI window. Against the old binary it failed in exactly 8 places; against your fix all 18 checks pass
* the sentinel never leaks: combined expansions of empty and unknown targets contain no NUL and no sentinel text
* `cargo audit` with warnings denied is clean on the refreshed lockfile

The NUL bearing sentinel is exactly the kind of solution I like: it fixes the conflation at the root where `expand_var` collapsed two different meanings into one empty string, keeps the public API contract identical for every existing caller, and cannot collide with real output by construction. The commentary you left in the code explaining why the windowless path errs toward empty rather than echoing is going to save some future reader a lot of head scratching.

Really appreciate the continued contributions. The status bar just got quietly more trustworthy.

PR #527, opened 2026-08-02
Implement tmux-compatible resize-window semantics

krhougs:
## Context

[tmex](https://github.com/krhougs/tmex) uses psmux as its tmux-compatible backend on Windows. When a browser terminal changes size, both tmex's local and SSH external connection paths issue the standard tmux command:

```
psmux resize-window -t @<window-id> -x <cols> -y <rows>
```

On current psmux, that command returned success but neither the target window nor its ConPTY changed size. The failure was especially visible for a background window: `#{window_width}`/`#{window_height}`, pane layouts, and the child process's observed console dimensions all remained stale.

This is not an inherent Windows limitation. ConPTY supports runtime resizing; the missing piece was psmux's window/client size model and command dispatch.

## Root cause

psmux previously represented several different tmux concepts with one global `last_window_area`:

- the attached client's viewport
- the active window's geometry
- the geometry used to lay out and resize panes

Individual windows did not own their dimensions or manual sizing state. As a result:

1. The normal one-shot/persistent command path selected at most one of `-x`/`-y`, then reached a no-op server request.
2. The control-mode compatibility shortcut updated the global active geometry instead of the targeted window.
3. `refresh-client -C` and `resize-window` were conflated even though tmux assigns them different responsibilities.
4. A background window, per-window control-client constraint, or manual size could not be represented correctly.

## tmux and psmux model comparison

| Concern | tmux | Previous psmux | This PR |
| --- | --- | --- | --- |
| Window geometry | Each `window` owns `sx/sy` and `manual_sx/manual_sy` | One global active rectangle | Each `Window` owns its area and local `window-size` mode |
| Client viewport | Stored per attached client, separately from windows | Conflated with window geometry | Separate `client_area` and per-client sizes |
| Control-client window size | Default client size plus per-window constraints | Only one global `W,H` update | `W,H`/`WxH`, `@id:WxH`, and `@id:` clear semantics |
| Command execution | Normal and control paths reach the same resize logic | Normal no-op plus a control-only shortcut | Shared parser and one server-side state transition |
| Dynamic sizing | `latest`, `largest`, `smallest`, and `manual` are evaluated per window | Client resize overwrote global geometry | Per-window mode with manual-size preservation and control-client clamping |

tmux can share windows across multiple sessions, while psmux currently has a single active-session model. This change does not transplant tmux's session graph; it implements the same sizing invariants within psmux's existing architecture.

One intentional platform difference remains: tmux permits a one-cell minimum, while psmux keeps its existing two-cell minimum because Windows ConPTY may terminate the child process below that size.

## Implementation

- Add authoritative geometry and local sizing mode to every `Window`.
- Keep client viewport, ordinary client sizes, control-client default/per-window sizes, and requested manual window sizes as distinct state.
- Use one tmux-compatible parser and execution path for CLI, persistent, config/key-binding, and control-mode commands.
- Implement independent/combined `-x` and `-y`, `-L/-R/-U/-D [adjustment]`, `-A/-a`, and current/id/index/name/pane targets with tmux flag priority.
- Make `resize-window` enter manual mode, preserve the requested size behind temporary control-client constraints, resize the target layout and ConPTYs, and emit layout changes.
- Implement `refresh-client -C` default size, per-window size, and clear operations with tmux-style errors.
- Make local `set-option window-size ...` window-scoped without requiring `-w`; `-g` remains global.
- Read window formats/layout strings from the target window and client formats from the client viewport.
- Use one client ID allocator so interactive and control clients cannot collide.

## Regression tests

The permanent Rust tests follow the repository's existing convention: they live in `tests-rs/test_resize_window_compat.rs` and are mounted from the production module. They call the real parser and state transitions without starting a server.

The 15 focused tests cover:

- the exact tmex `resize-window -t @id -x cols -y rows` shape
- attached/separate flags, session-qualified IDs, tmux flag priority, and invalid ConPTY dimensions
- background-window isolation and separation from the client viewport
- relative, largest, and smallest sizing
- manual mode surviving client resize
- per-window control-client clamping and clear restoration
- local `window-size` transitions
- valid and invalid `refresh-client -C` forms

Existing Windows PowerShell checks were also changed from accepting a silent no-op to asserting the resulting `#{window_width}` and `#{window_height}`.

## Validation

- `cargo check --target x86_64-pc-windows-gnu`
- complete Windows-target unit-test build
- focused unit tests executed on Windows: **15 passed, 0 failed**
- isolated Windows E2E: **36 assertions passed**, covering independent/background windows, actual child-observed console dimensions, control-mode default/per-window/clear behavior, invalid-size `%error`, manual/dynamic transitions, and window switching
- repeated the build and E2E from the clean upstream branch
- verified the isolated tests did not change the user's default psmux data or pre-existing psmux processes

## PR scope

This branch is based directly on current `psmux/psmux:master` and contains only two resize-related commits: the implementation and its regression tests. It does not include any unrelated fork integration commits.


psmux:
Merged. Thank you @krhougs, this is a genuinely excellent contribution and exactly the kind of PR I love to receive: a clear model comparison against tmux, a root cause that goes after the actual architectural gap instead of patching symptoms, and permanent regression tests that run without a server.

I verified everything independently before merging:

**Reproduction on master.** Your claim held exactly as written. `resize-window -t @1 -x 100 -y 40` against a background window returned exit 0, the window stayed 80x24, `window-size` stayed `latest`, and even `-x 99999` was silently accepted.

**With your branch.** The target window went to 100x40 while the other window and the client viewport stayed at 80x24, `window-size` reads `manual` on the target only, and the resize is real all the way down: a shell inside the resized window reports its console as 100x40, so the ConPTY resized too. Relative `-R 5` after manual mode gave 105x40 as expected.

**tmux parity.** I ran the same scenarios against a live tmux 3.4 as an oracle plus `cmd-resize-window.c` in the tmux source. Flag priority (`-R` beating `-D`, `-A` beating `-a`), manual mode entry, per window isolation, adjustment defaults, and the error strings all match. Your two cell minimum divergence for ConPTY safety is sensible and documented, so I am keeping it.

**No regressions.** A real attached client still drives dynamic sizing, splits, zoom, and `set-option window-size latest` correctly resumes tracking the client after manual mode. Full suite is green at 2760 passed, 0 failed, which includes your 15 new tests.

Two small things I handled during the merge since master moved past your base: the `CtrlReq` enum grew `SetOptionToggle` from #535 next to your `SetWindowSize` (trivial conflict), and a test added in #526 needed the new `Window` fields. Also, unifying the interactive and control client ID allocators was a quiet but real correctness fix that I appreciated you catching along the way.

Happy to have tmex riding on psmux. This closes a long standing gap where `resize-window` was documented as a platform no op when it never needed to be.

Issue #459, opened 2026-07-10
psmux spawns an unbounded, continuously-growing number of psmux.exe processes (~100+ and climbing) after running multiple panes with heavy TUI apps.


fuzzberg:
**Environment**

I'm running psmux version 3.3.6 on Windows 10, version 2009, build 19044, using Windows Terminal with a PowerShell 7 profile. I have the Windows Terminal setting `experimental.rendering.forceFullRepaint` enabled. The companion tools involved are `tmuxpanel` (Tmux Plugin Panel) v0.1.1, `pstop` around v0.3.x, and `btop4win` v1.0+.

**Description (primary/most severe issue)**

I opened a psmux session with 3 panes in the same window: `pstop`, `btop4win`, and `tmuxpanel` running simultaneously. Performance was poor, with CPU near 100%, but the session itself did not crash or freeze while I was actively using it.

After closing the terminal window, I checked Task Manager and found roughly 100 `psmux.exe` processes running under PowerShell, and the count was actively climbing, with new `psmux.exe` processes appearing roughly every second. I'm not sure whether the runaway spawning started while I was still using the 3-pane session or was triggered specifically by closing the terminal, but either way this is clearly not expected behavior. It left me concerned about how far it would climb if left unattended, since each process seems small individually but the growth is unbounded and ongoing, and could plausibly leave a system fully unresponsive if not noticed quickly.

This appears to be the same general class of problem as a separate incident I hit earlier the same day (described below), but seems more severe. That one produced a one-time spike to about 400 processes tied to a plugin install/config-reload action, while this one is an ongoing, continuous spawn with no obvious trigger, occurring either during multi-pane use or when closing the session/terminal.

**Steps to reproduce (best recollection)**

I opened a psmux session and opened 3 panes in the same window running `pstop`, `btop4win`, and `tmuxpanel`. I used it normally for a while, noting heavy CPU load around 100% but no crash or freeze during use. I then closed the Windows Terminal window. When I checked Task Manager afterward, about 100 `psmux.exe` processes were present, and the count kept increasing roughly once per second with no further action on my part.

**Related, separate incidents from earlier the same day (included for pattern context)**

[...]

PR #531, opened 2026-08-03
fix(server): guard the warm session name so a namespace keeps exactly one warm server (#459)

acoliver:
## Overview

Fixes the unbounded warm-server growth in #459.

Every real session is protected by the single-server-per-name mutex (#2): a second server that finds the name already owned exits immediately rather than running as a duplicate. Warm (`__warm__`) servers were explicitly exempted:

```rust
if crate::session::is_warm_session(&base) {
    None                                    // warm: no guard
} else {
    match crate::platform::acquire_session_mutex(&base) { ... }
}
```

The stated rationale, in both `run_server` and `rekey_session_guard`, was that "the warm pool intentionally runs several at once".

**There is no pool.** `spawn_warm_server` early-returns as soon as it verifies a live warm, and the handoff is a *single* `__warm__.port` file per namespace. A namespace can only ever publish one warm server no matter how many are running, so everything past the first is unreachable and immortal — nothing attaches to it, nothing claims it, and nothing ever kills it.

That left `acquire_warm_spawn_lock` as warm's only exclusion mechanism, and it is weaker than a mutex in three ways that all showed up in the field:

- **time-boxed** — held only while waiting for the new warm's `.port` to appear, then released after ~3s whether or not it ever did;
- **not crash-safe** — a plain file whose `Drop` never runs when the owner is killed, so a dead owner leaves a lock cleared only by the 20s staleness steal;
- **wrong scope** — it guards the check→spawn window, so *sequential* spawns seconds apart pass straight through.

### Field evidence

From the inventory I posted on #459: a five-deep parent chain of live `server -s __warm__` processes in the default namespace, spawned 1–4s apart, each one's parent the previous psmux — and **no `__warm__.port` file present at all**. Every warm failed to register, so the next caller saw "no warm exists" and started another. The chain stopped only because the last lock owner was killed and left a stale lock behind.

## The fix

Warm names take the guard on the same terms as every other session name. A duplicate warm now exits at startup exactly like a duplicate session server, before touching the winner's `.port`/`.key`/`.sid`.

The claim handoff still works, and the ordering already in the code is what makes it safe: `CtrlReq::ClaimSession` calls `rekey_session_guard` onto the claimed name *before* calling `spawn_warm_server`, so `__warm__` is already free when the replacement starts. Without that ordering this change would have left namespaces with no warm server at all.

Using the named mutex also fixes the killed-owner case for free: Windows reports `WAIT_ABANDONED` and the next warm takes ownership immediately, where the lock file required a 20-second staleness steal.

`acquire_warm_spawn_lock` is left in place — it still usefully collapses genuinely concurrent spawn attempts before they cost a process launch. It is simply no longer the only thing standing between a slow registration and a permanent leak.

## Behaviour change

`tests-rs/test_issue505_rename_session_guard.rs` had a test asserting the exemption (`warm_name_is_exempt_from_the_guard`). It is updated to assert the new behaviour and renamed accordingly; its sibling `claiming_a_warm_server_guards_the_claimed_name` now starts from a warm server that holds the warm name, which is the real startup state after this change.

## Validation

- 4 new unit tests in `tests-rs/test_issue459_warm_single_instance.rs`: a second warm cannot take the name; warm names stay isolated per `-L` namespace; a claim frees the warm name and a replacement can acquire it; a warm guard does not block a real session name in the same namespace. Ownership is probed from a spawned thread, since a Windows mutex is recursive for its owning thread and would otherwise always report "free".
- New E2E `tests/repro_issue459_warm_duplicates.ps1`, verified to **fail on unfixed code and pass with the fix**:

```
  (unfixed)                                (fixed)
  FAIL  exactly one warm server survives   PASS  exactly one warm server survives
        -- found 2: 19760, 31664           PASS  the survivor is the first warm
  FAIL  the duplicate exited on its own    PASS  the duplicate exited on its own
  FAIL  namespace settles at 2 servers     PASS  namespace settles at 2 servers
        -- found 3: 19760, 31664, 23404
```

Note the unfixed scenario-2 line: after a claim the namespace held **three** servers — the claimed session, its replacement warm, and the orphaned duplicate that will now never die.

- Full suite: **2693 passed, 0 failed** (162.90s), up from 2689 on master.

The E2E matches processes on **both** an image path under the repo's `target\` directory **and** this run's unique random namespace, so it cannot touch an installed psmux or any other server on the machine; teardown is in a `finally` and uses namespace-scoped `kill-server` before any force-kill. I verified the process set was identical before and after the run, and that it left no registry files behind.

## Scope

This addresses the warm-server replication mechanism specifically. `cac2dcf` earlier fixed hook accumulation under the same issue number; both contribute to #459, and I have not tried to establish whether the reporter's ~100 processes were all from this path.

---

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -759,10 +759,11 @@ pub(crate) fn read_fresh_config_warnings(since_epoch: u64) -> Vec<String> {
 /// Release before acquiring so renaming a session onto a name it already holds
 /// cannot block on itself. A refused acquire is not fatal: the rename has
 /// already happened, so we simply run unguarded rather than abandon the session.
-/// Warm servers stay exempt (the pool intentionally runs several at once).
+/// Warm servers are guarded too: a namespace holds exactly one `__warm__`
+/// server, and releasing that name here is precisely what lets the replacement
+/// warm spawned after a claim acquire it (issue #459).
 fn rekey_session_guard(guard: &mut Option<crate::platform::SessionMutex>, new_base: &str) {
     *guard = None; // drop releases + closes the old name's mutex
-    if crate::session::is_warm_session(new_base) { return; }
     *guard = crate::platform::acquire_session_mutex(new_base);
     if guard.is_none() {
         warm_debug(&format!("session guard: '{}' already owned by a live server — running unguarded", new_base));
@@ -813,15 +814,15 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
     // a cold-spawn race (has-session false-negatived under load, or two
     // `new-session -s X` raced) — exit cleanly so the winner stays the single
     // source of truth. Two servers on one name desync the .port/.key files and
-    // wedge the session ("appears lost"). Warm (standby) servers are exempt (the
-    // warm pool intentionally runs several). Fail-open: any FFI hiccup yields a
-    // live guard, never a blocked legitimate start.
+    // wedge the session ("appears lost"). Warm (standby) servers are guarded on
+    // the same terms: `__warm__.port` is a single file, so a namespace can only
+    // ever publish one warm server, and an unguarded warm name let every failed
+    // or slow registration strand another live process (issue #459). Fail-open:
+    // any FFI hiccup yields a live guard, never a blocked legitimate start.
     // Re-keyed on every rename/claim, see rekey_session_guard (issue #505).
     let mut session_guard = {
         let base = app.port_file_base();
-        if crate::session::is_warm_session(&base) {
-            None
-        } else {
+        {
             match crate::platform::acquire_session_mutex(&base) {
                 Some(g) => Some(g),
                 None => {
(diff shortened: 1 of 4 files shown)


psmux:
Merged. @acoliver this was a pleasure to review. "There is no pool" is the kind of sentence that unlocks a whole bug class, and your three point takedown of the spawn lock (time boxed, not crash safe, wrong scope) matched exactly what I saw in the field evidence on #459.

I ran your repro script on both sides before merging and it discriminates perfectly:

On unfixed master:

```
FAIL  exactly one warm server survives  (found 2)
FAIL  the duplicate exited on its own
FAIL  namespace settles at 2 servers    (found 3 after the claim)
```

With your fix, all seven checks pass: the duplicate exits at startup on its own, the winner keeps its `.port` beacon, and after a claim the namespace settles at exactly two servers, the claimed session plus its replacement warm. Your point about the claim ordering already being what makes the handoff safe (`rekey_session_guard` before `spawn_warm_server` in `ClaimSession`) checked out in the code and in the test.

I also appreciated the test hygiene: probing mutex ownership from a spawned thread because a Windows mutex is recursive for its owner is a trap I have hit before, and your repro matching processes on both the repo image path and a unique namespace means it can never touch a real server. The `WAIT_ABANDONED` behavior coming along for free is a nice bonus over the 20 second staleness steal.

Full suite green at 2745 on your branch and 2760 on the final merged tree. Thank you for the careful work here, and for the field inventory on the issue that made the mechanism undeniable.

PR #532, opened 2026-08-03
fix: prune registry files orphaned by their .port

### LINKED ISSUE #530: ~/.psmux accumulates .sid/.pid/.spawnlock files forever (6,534 files, oldest from July 13)
## Summary

`~/.psmux` accumulates `.sid`, `.pid` and `.spawnlock` files forever. On a developer machine that has been running psmux under an automated multi-agent workload since mid-July, the directory holds **6,534 files**:

```
.sid          5988
.pid           310
.spawnlock     162
.key            35
.port           35
(no ext)         2
.log             2
```

Oldest entry: `jefe-compat-probe____warm__.sid`, dated 07/13/2026.

## The tell

`.key` and `.port` sit at 35 each, which tracks the 36 live `psmux.exe` processes almost exactly. Those two extensions **are** being cleaned up correctly on teardown.

`.sid`, `.pid` and `.spawnlock` are not cleaned up by anything, ever. Nothing prunes them on server exit, on `kill-server`, or on any startup sweep — so they are a pure monotonic accumulation, one set per session that has ever existed.

The 162 stranded `.spawnlock` files are a related but distinct symptom: `WarmSpawnLock` removes the file in `Drop`, and `Drop` does not run when a process is terminated rather than exiting. The 20s staleness steal keeps those stale locks from being *harmful*, but nothing ever deletes them.

## Why it is worth fixing

- 6,500 files in a directory that every psmux invocation enumerates is a startup cost that grows without bound. `cleanup_stale_port_files` and the reaper both scan this directory on client startup.
- It makes the directory useless for diagnosis. Working out which of 5,988 `.sid` files corresponded to a live server required cross-referencing against the process table.
- It is invisible until someone goes looking, because the files are tiny.

## Related: `instances/` has the same shape

`instances/` (added in #509, which I contributed) holds **168** namespace token files. My design note there argued that accumulation was "bounded by distinct namespace names, exactly as it already is for `.port`/`.key`/`.sid`/`.pid`".

That assumption is wrong in practice, and I want to correct it on the record. The workload on this machine mints a **fresh namespace on every startup** for conformance probing, so distinct namespace names are unbounded, and `instances/` grows with them. The premise that made unpruned tokens acceptable does not hold for callers that use namespaces as disposable isolation — which is exactly what namespaces are good for.

If a pruning pass lands for `.sid`/`.pid`/`.spawnlock`, `instances/` should ride along in it. Happy to write that patch if you want it.

## Suggested shape

A sweep on server startup that removes registry files for namespaces with no live server would handle all four extensions at once, and would piggyback on the directory scan `cleanup_stale_port_files` already performs rather than adding a new one.

---

acoliver:
Fixes #530.

## Overview

Files in `~/.psmux` are never pruned. They accumulate for the life of the machine, and the largest producer is `kill-server` itself.

## Root cause

Every sweep over the data directory is anchored on `.port` files:

```rust
let ext = path.extension().and_then(|e| e.to_str()).unwrap_or("");
if ext != "port" {
    continue;
}
```

The `.port` file is therefore the *only* entry point to a session's registry set. Once it is gone, its `.sid`, `.key`, `.pid` and `.spawnlock` siblings are unreachable by any code path in the program — no sweep will ever look at them again.

Teardown removed the `.port` first and the `.sid` never:

- `kill-server` (two sites in `main.rs`) removed `.port`, `.key` and `.pid`, leaving one stranded `.sid` per session, permanently.
- Roughly a dozen further teardown sites remove only the `.port`/`.key` pair.
- `.spawnlock` is released solely by `WarmSpawnLock::Drop`, which by definition cannot run when the holder is killed.

Only two call sites removed the complete set.

This is not only a disk problem. `resolve_session_by_id` walks **every** `.sid` file to map `$N` to a session name, so each orphan is re-read on every id lookup.

## Field evidence

From the data directory on the machine in the report:

| Extension | Count |
|---|---|
| `.sid` | 6,060 |
| `.pid` | 311 |
| `.spawnlock` | 161 |
| `.key` | 17 |
| `.port` | 17 |

6,634 files total, oldest three weeks old, against 17 live sessions. The `.sid` count tracks the number of sessions ever created; the shape is exactly what the missing `.sid` unlink predicts. Tooling that mints a fresh `-L` namespace per run reaches this state much faster.

## The fix

**Teardown removes the whole set.** Both `kill-server` sites now call the existing `remove_session_registry_files` helper instead of unlinking three of the four files by hand.

**A startup sweep collects what earlier versions stranded.** `prune_orphaned_registry_files` runs alongside the existing `cleanup_stale_port_files`, and is deliberately conservative:

- Only `.sid`, `.key`, `.pid` and `.spawnlock` are considered, and only when no sibling `.port` exists. A set with a `.port` belongs to the port-driven sweep and is never touched here. Files that are not registry satellites — `next_session_id`, logs, the `instances/` and `servers/` directories — are not candidates at all.
- A **60 second grace period** protects a starting server. `ensure_session_registry_files` writes `.sid`, `.key` and `.pid` *before* it publishes the `.port` beacon, so a healthy server transiently looks exactly like an orphan. A running server also rewrites its whole set on the 5s self-heal tick, so its files can never age past this bound.
- An orphan whose anchoring pid still belongs to a live psmux process is **kept**, so a server that is merely slow to publish is never disturbed. `.spawnlock` carries its holder's pid in its own body; the other extensions anchor on the sibling `.pid`.

## Behaviour changes

Recovering a session by finding a bare `.sid` on disk was never possible — nothing reads a `.sid` without a `.port` — so nothing that used to work stops working. On the reporter's machine the sweep reclaims 6,060 `.sid` and 161 `.spawnlock` files on the next invocation and keeps all 17 live sessions.

## Validation

New E2E `tests/repro_issue530_registry_pruning.ps1`, against the **unfixed** binary:

```
--- Scenario B: orphans are swept, the LIVE session is not ---
  FAIL  orphaned .sid files were pruned
        still present: ghost-a__probe.sid, ghost-b__probe.sid
  FAIL  orphan set with a DEAD pid anchor was pruned
  FAIL  stale .spawnlock from a dead holder was reclaimed
--- Scenario C: kill-server must not strand the .sid ---
  FAIL  kill-server removed the ENTIRE registry set (incl. .sid)
        leftovers: psmux530e2e3516041119__work.sid
```

With the fix, 8/8 pass. The script redirects `USERPROFILE` to a temp directory so it can never touch a developer's real data directory, and tears down its servers by matching both the repo-built image and a per-run namespace.

`tests-rs/test_issue530_registry_pruning.rs` adds 9 unit tests. They drive an inner function taking an injectable grace period and liveness oracle, so both the startup window and the live-anchor branch are exercised deterministically without spawning servers or waiting a minute.

Full suite: **2698 passed, 0 failed** (2689 before these tests).

## Scope

Off Windows the process-table check is unavailable, so pid-anchored orphans are kept rather than guessed at. Unanchored orphans — the overwhelming majority, and the entire `.sid` backlog — are pruned on every platform.

The dozen teardown sites that remove only the `.port`/`.key` pair are left as they are; the sweep now collects after them. Converting them to the whole-set helper would be a reasonable follow-up.

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -216,6 +216,11 @@ fn run_main() -> io::Result<()> {
 
     // Clean up any stale port files at startup
     cleanup_stale_port_files();
+    // Then drop registry files whose `.port` entry is already gone (issue
+    // #530). The sweep above is the only thing that can reach them, and it
+    // finds entries BY their `.port` file — so a satellite that outlives its
+    // port is invisible to it and accumulates forever.
+    crate::session::prune_orphaned_registry_files();
     // Then reap any LIVE but orphaned server processes (issue #448): duplicates
     // or crashed-client headless servers that cleanup_stale_port_files cannot
     // see because they have no registry file. Bounds the process count so
@@ -533,19 +538,17 @@ fn run_main() -> io::Result<()> {
                             }
                         }
                     }
-                    // Remove port/key/pid files regardless
-                    let _ = std::fs::remove_file(&path);
-                    let _ = std::fs::remove_file(path.with_extension("key"));
-                    let _ = std::fs::remove_file(path.with_extension("pid"));
+                    // Remove the whole registry set regardless. Deleting only
+                    // port/key/pid used to strand the `.sid`, which no sweep can
+                    // reach once its `.port` is gone (#530).
+                    crate::session::remove_session_registry_files(&path);
                 })
             }).collect();
             // Wait for all threads to complete
             for h in handles { let _ = h.join(); }
-            // Clean up stale port/key/pid files
+            // Clean up stale registry sets (whole set, including `.sid` — #530)
             for path in &stale_ports {
-                let _ = std::fs::remove_file(path);
-                let _ = std::fs::remove_file(path.with_extension("key"));
-                let _ = std::fs::remove_file(path.with_extension("pid"));
+                crate::session::remove_session_registry_files(path);
             }
             // Force-kill any wedged server that ignored the graceful kill. The
             // candidates were read from this data dir's (and, with -L, this
--- a/src/session.rs
+++ b/src/session.rs
@@ -272,6 +272,27 @@ const ORPHAN_REGISTRY_EXTS: &[&str] = &["sid", "key", "pid", "spawnlock"];
 /// clearing the backlog on the next CLI invocation.
 const ORPHAN_REGISTRY_GRACE: Duration = Duration::from_secs(60);
 
+/// Most files a single sweep may remove.
+///
+/// psmux CLI commands are invoked constantly by scripts and prompts, and the
+/// backlog this fix targets can run to thousands of files. Deleting all of it
+/// inside one arbitrary invocation — `psmux -V`, say — makes a trivial command
+/// do a surprising amount of destructive work and stalls it on I/O. The budget
+/// keeps any one invocation cheap and bounded; the backlog drains across
+/// successive sweeps instead, which is just as effective and far less abrupt.
+const ORPHAN_REGISTRY_SWEEP_BUDGET: usize = 256;
+
+/// Minimum interval between sweeps, tracked by [`ORPHAN_REGISTRY_SWEEP_STAMP`].
+///
+/// Without this, every psmux invocation pays a full directory walk. Orphans are
+/// produced only by session teardown, so there is nothing to gain from looking
+/// more often than this.
+const ORPHAN_REGISTRY_SWEEP_INTERVAL: Duration = Duration::from_secs(300);
+
+/// Marker file recording when the last sweep ran. Leading dot, so it has no
+/// extension and can never be mistaken for a registry satellite.
+const ORPHAN_REGISTRY_SWEEP_STAMP: &str = ".registry_sweep";
+
 /// Delete per-session registry files whose `.port` entry is already gone
 /// (issue #530).
 ///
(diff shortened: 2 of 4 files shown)


acoliver:
## Revision: the sweep is now rate-limited and budgeted

I hit a real hazard with the first version of this patch and want to be upfront about it, because it changes the design.

The sweep was wired into `run_main()`, which every psmux invocation goes through — including ones that do no session work at all. I ran `psmux -V` on a build of this branch against a data directory holding the backlog from #530, purely to print a version string, and it deleted **6,552 files** in that one call. The cleanup was correct; every one of those files was a genuine orphan. But having `psmux -V` do thousands of unlink syscalls is the wrong shape for a tool that scripts and shell prompts invoke constantly, and it makes the first run after upgrading surprisingly destructive.

Two bounds, both in `session.rs`:

**A per-sweep budget** (`ORPHAN_REGISTRY_SWEEP_BUDGET`, 256). The loop stops once it has removed that many files and leaves the rest for the next sweep. A backlog still drains completely, just across several invocations instead of one. No single command can now do an unbounded amount of destructive work, and sweep latency is bounded regardless of how large the directory has grown.

**A rate limit** (`ORPHAN_REGISTRY_SWEEP_INTERVAL`, 5 minutes) tracked by a `.registry_sweep` stamp file in the data dir. Orphans are only produced by session teardown, so there is nothing to gain from walking the directory on every command. The stamp is written *before* the sweep runs, so a process killed midway waits a full interval rather than immediately retrying the same work. It has a leading dot and therefore no extension, so it can never be mistaken for a registry satellite.

Steady-state cost drops to one `stat` on the stamp for the overwhelming majority of invocations.

### Tests

Two added, 11 total for this issue:

- `a_single_sweep_removes_no_more_than_its_budget` — plants 40 orphans, sweeps with a budget of 10, asserts exactly 10 go and 30 survive, then drives successive sweeps and asserts the backlog drains to zero.
- `a_sweep_is_rate_limited_by_its_stamp` — a sweep is due with no stamp, suppressed immediately afterwards, due again once the interval has elapsed, and the stamp survives the sweep it schedules.

`prune_orphaned_registry_files_in_with` takes the budget as a parameter alongside the existing injected grace period and liveness oracle, so both bounds are driven deterministically without waiting on wall-clock time.

The E2E needed a corresponding change: it backdates the stamp before the invocation it expects to sweep, since an earlier command in the same scenario has already stamped it. That is the script standing in for the interval elapsing, and it is commented as such.

Full suite: **2700 passed, 0 failed**. E2E: 8/8.

psmux:
Merged in cb098c0 (and db5c87c for the first head). Full write up on #530: https://github.com/psmux/psmux/issues/530#issuecomment-5168757523

Issue #533, opened 2026-08-03
[vt100] VS16 emoji-presentation sequences stored as one narrow cell — every column after them drifts, leaving on-screen residue


kt81:
## Summary

A text-presentation base character followed by `U+FE0F` (VS16) requests emoji presentation, which real terminals, tmux, and `unicode-width`'s **string-level** `width()` all count as two columns. `Screen::text` measures width **one character at a time** with `UnicodeWidthChar::width()`, so the base character settles the cell at one column and the VS16 that follows is folded in afterwards as a zero-width combining mark. Emoji presentation is a property of the *sequence*, so a char-at-a-time measure can never see it.

## Reproduction

`crates/vt100-psmux` at master (`941a862`):

```rust
let mut p = vt100_psmux::Parser::new(1, 20, 0);
p.process("\u{2733}\u{FE0F}AB".as_bytes());
// 'A' should land on col 2, as it does in a real terminal and in tmux.
// It lands on col 1.
```

## Impact

The grid ends up **one column narrower** than the terminal the child process believes it is driving. When the child repaints using a position it computed in its own (correct) width model, the write lands one cell off, and **the cell it skips keeps the previous frame's content**.

## Where this was found — real usage

[banto](https://github.com/kt81/banto) is a resident TUI on Windows that manages agent-CLI sessions, hosting each pane in a PTY and interpreting the output with this crate. The integration is a thin wrapper:

- [`crates/banto-core/src/screen.rs`](https://github.com/kt81/banto/blob/main/crates/banto-core/src/screen.rs) — `Parser::new(rows, cols, 2000)` + `process(&[u8])`
- [`crates/banto-tui/src/render.rs`](https://github.com/kt81/banto/blob/main/crates/banto-tui/src/render.rs) — grid to ratatui cells
- [`Cargo.toml`](https://github.com/kt81/banto/blob/main/Cargo.toml) — `vt100 = { version = "0.16.9", package = "vt100-psmux" }`

The measured symptom: in a pane running Claude Code, its status line draws a ten-cell usage bar. It rendered across **eleven** cells, and the extra leading cell still held a **red block from before the context was compacted** — red is only used at ≥80% usage, so that cell had not been written since. Only the row containing a VS16 emoji (`✳️`) was damaged; rows containing plain emoji (`📛`, `📁` — wide on their own) were untouched.

[...]

Issue #534, opened 2026-08-03
[vt100] Panic: shrinking a row through a wide char's continuation, then erase-in-line


kt81:
## Summary

Shrinking a row past the continuation cell of a wide character leaves the row's last cell flagged wide with no continuation after it. An erase-in-line over that orphaned cell then panics.

## Reproduction

`crates/vt100-psmux` at master (`941a862`):

```rust
let mut parser = vt100_psmux::Parser::new(2, 4, 0);
parser.process("\u{4F60}".as_bytes());   // a wide CJK char
parser.screen_mut().set_size(2, 1);      // truncate away its continuation cell
parser.process(b"\x1b[K");               // panics at row.rs:66
```

## Analysis

`Row::clear_wide` is already bounds-checked here, so that half is covered. Two places remain:

1. **`Row::resize`** does not clear a wide cell whose continuation was just truncated when shrinking. `Row::truncate` already does the equivalent.
2. **`Row::erase`** derives the logical last column from the pre-erase wide flag as `cols() - if wide { 2 } else { 1 }`. For an orphaned wide cell in a one-column row that is `1 - 2`, which underflows `u16`.

## Where this was found — real usage

[banto](https://github.com/kt81/banto) hosts each pane in a PTY and interprets the output with this crate ([`crates/banto-core/src/screen.rs`](https://github.com/kt81/banto/blob/main/crates/banto-core/src/screen.rs)). Splitting and resizing panes is routine there, and panes render CJK, so this was a reachable path rather than a theoretical one.

## Known fix

[doy/vt100-rust#28](https://github.com/doy/vt100-rust/issues/28) / [PR #30](https://github.com/doy/vt100-rust/pull/30) by **Junyi Hou**, open. The truncation clear in `Row::resize` and a `saturating_sub` in `Row::erase`; behaviour is unchanged for any row with two or more columns.

Ported onto the same fork as the VS16 fix, at [`kt81/psmux` branch `banto/vt100-patches`](https://github.com/kt81/psmux/tree/banto/vt100-patches), if it is useful.

psmux:
Confirmed and fixed in 991f107. Your snippet reproduced verbatim on the first run, panicking at `row.rs:66` exactly as you said. Two reports in one day, both precise, both with the upstream fix already identified. Thank you.

## Reproducing it first

Before reading the parser I wrote fifteen tests around this path. Nine failed on master. Both of the sites you named were real:

[...]

PR #538, opened 2026-08-04
feat(pty): expose ConPTY passthrough mode

kt81:
Follows on from your note in #533, that `CreateProcessW` rejects a passthrough HPCON on Windows 11 build 26200 and portable-pty falls back to plain ConPTY. We are on that build, and we have a scrollback finding of our own that we attributed to passthrough — so we wanted to know which side we were actually on, and found we could not ask.

Passthrough is decided twice: at creation, gated on the build and `PSMUX_NO_PASSTHROUGH`, and again inside `spawn_command`, where an `ERROR_INVALID_PARAMETER` rebuilds the console without the flag and retries. Both are right. Neither is visible from outside: `used_passthrough` is `pub`, but `win::psuedocon` is private and `ConPtyMasterPty`'s `inner` is private, so nothing reachable from `native_pty_system().openpty()` can read it. A successful fallback and a successful passthrough are the same observation. The `log::warn!` is the only signal, and a consumer that never wired up a logger — ours had not — sees nothing.

This adds a defaulted `MasterPty::conpty_passthrough_mode(&self) -> Option<bool>`. `Option<bool>` rather than a transport enum because the state you already keep is a bool and `None` covers the backends with no answer; an enum would add public surface for states that do not exist yet. It reads from the master rather than from `openpty`'s return, because the second fallback happens during spawn — a value read before spawning would be wrong in exactly the case worth asking about.

## On testing

The environment-disabled path is covered on Windows. The test deliberately does not set `PSMUX_NO_PASSTHROUGH` in its own process: it re-runs its own binary with `--exact` for the child case and passes the variable through `Command::env`, so nothing else in this binary or any other can observe it set. That mattered here because the variable's reader is ConPTY creation itself, which the `ENV_MUTEX` pattern in `tests-rs/test_config_exhaustive.rs` could not have serialised against.

The post-spawn fallback is **not** covered, and it is the interesting one. It fires only when the OS rejects the flag, and there is no seam in the tree to force that, so a test would either depend on a particular build or assert nothing. I would rather say so than ship something that looks like coverage. If you have a seam in mind, I am glad to add it.

`cargo fmt --check` does not pass on this tree, but not because of this change — `build.rs` and parts of `win/` are already unformatted at `db303ae`. I matched rustfmt's output by hand for the added lines rather than sweep unrelated files into the diff. Per `AGENTS.md` I did not run the full `cargo test --all-targets`; targeted `cargo test -p portable-pty-psmux` and `cargo check` are green.

### DIFF
--- a/crates/portable-pty-psmux/src/lib.rs
+++ b/crates/portable-pty-psmux/src/lib.rs
@@ -101,6 +101,16 @@ pub trait MasterPty: Downcast + Send {
     /// It is invalid to take the writer more than once.
     fn take_writer(&self) -> Result<Box<dyn std::io::Write + Send>, Error>;
 
+    /// Whether this is a ConPTY created with passthrough mode enabled.
+    ///
+    /// `None` means that the pty implementation is not ConPTY, or does not
+    /// expose this implementation-specific detail.  The result should be
+    /// queried after spawning a child: ConPTY can fall back to a newly-created
+    /// non-passthrough console when process creation rejects passthrough mode.
+    fn conpty_passthrough_mode(&self) -> Option<bool> {
+        None
+    }
+
     /// If applicable to the type of the tty, return the local process id
     /// of the process group or session leader
     #[cfg(unix)]
@@ -124,6 +134,18 @@ pub trait MasterPty: Downcast + Send {
     }
 }
 impl_downcast!(MasterPty);
+
+#[cfg(all(test, not(windows)))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn non_conpty_master_reports_no_passthrough_mode() {
+        let pair = native_pty_system().openpty(PtySize::default()).unwrap();
+
+        assert_eq!(pair.master.conpty_passthrough_mode(), None);
+    }
+}
 
 /// Represents a child process spawned into the pty.
 /// This handle can be used to wait for or terminate that child process.
--- a/crates/portable-pty-psmux/src/win/conpty.rs
+++ b/crates/portable-pty-psmux/src/win/conpty.rs
@@ -141,6 +141,10 @@ impl MasterPty for ConPtyMasterPty {
                 .ok_or_else(|| anyhow::anyhow!("writer already taken"))?,
         ))
     }
+
+    fn conpty_passthrough_mode(&self) -> Option<bool> {
+        Some(self.inner.lock().unwrap().con.used_passthrough)
+    }
 }
 
 impl SlavePty for ConPtySlavePty {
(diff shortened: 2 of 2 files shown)


psmux:
Merged. Thanks @kt81, this is a well reasoned little API and the writeup made the review easy.

I verified the core claim before merging: `used_passthrough` is genuinely unreachable from outside the crate today. Even though the `win` module is public, `ConPtyMasterPty` keeps its `inner` private, so a consumer holding the master from `openpty()` has no path to the flag, and the downcast route dead ends at the same private field. A successful fallback and a successful passthrough really are the same observation from outside, which is exactly the situation worth fixing on build 26200 where `CreateProcessW` rejects the passthrough HPCON and the silent retry kicks in.

Design points I agree with: `Option<bool>` over an enum for state that is already a bool, the defaulted trait method so other backends are untouched, and reading from the master after spawn since the second fallback happens inside `spawn_command`. That last one is the subtle bit and the doc comment covers it.

On testing: your environment disabled test passes here (`win::conpty::tests::reports_when_passthrough_is_disabled_by_environment`, 11 crate tests green), and the child process pattern for keeping `PSMUX_NO_PASSTHROUGH` out of the parent environment is the right call given the reader is ConPTY creation itself. I agree the post spawn fallback path is not testable without a seam and I would rather have your honest note than fake coverage. If I add a seam for it later I will loop you in.

No concern about the rustfmt state, you are right that `build.rs` and parts of `win/` were already unformatted and keeping the diff clean was the better choice. Full suite green at 2760 on the merged tree.

kt81:
Thank you — and thank you for verifying the reachability claim yourself instead of taking it from the description. That is the second time in two days I have watched you re-derive something before acting on it, and it is a good thing to be on the other side of.

One correction I owe you, which I found while writing up our own notes an hour ago. In #534 I said we had a scrollback finding we attributed to ConPTY reinterpreting rather than passing through. The measurement behind that was a controlled A/B on this machine: identical child, prompt and geometry, vanilla `portable-pty` versus `portable-pty-psmux`, and the difference was stark — no scrollback and a stream full of `ESC[1;25r` DECSTBM on one side, 191 rows of scrollback and zero DECSTBM on the other. What I had not noticed until today is that the A/B swapped the whole crate, and the fork adds three things, not one: passthrough, `RESIZE_QUIRK` and `WIN32_INPUT_MODE`. So the difference is real and the attribution to passthrough specifically is not established. `PSMUX_NO_PASSTHROUGH=1` isolates it in one run, which we had not thought to do.

So: we were more confident than our evidence supported, and your #533 note is what eventually made that visible. We will re-run it properly — with the isolation, and now with `conpty_passthrough_mode` to say which side we actually landed on — and if it says anything you can use about 26200, it will be its own issue.

PR #539, opened 2026-08-04
test(vt100): cover OSC 7 with Windows drive-letter and UNC paths

Gerrrt:
Adds parser-level coverage for OSC 7 (`ESC ] 7 ; <uri> ST`) with the path shapes a Windows shell actually emits. There is currently no OSC 7 test in the `vt100-psmux` crate, and `#{pane_path}` reads directly from what this parses, so it is worth pinning down.

`tests/osc7_windows_path.rs` covers:

- `file:///C:/...` drive-letter paths, with both the `ST` (`ESC \`) and `BEL` terminators
- percent-decoding (`%20` → space, and similar), since `[uri]`-built URIs from PowerShell are percent-encoded
- host stripping for a `file://server/share` UNC authority
- the basename that `#{b:pane_path}` takes
- an empty OSC 7 payload not clobbering a previously recorded path

No behavior change. Test only.

### DIFF
--- a/crates/vt100-psmux/Cargo.toml
+++ b/crates/vt100-psmux/Cargo.toml
@@ -65,6 +65,10 @@ path = "tests/issue502_altscreen_attrs.rs"
 name = "issue502_decsc_altscreen"
 path = "tests/issue502_decsc_altscreen.rs"
 
+[[test]]
+name = "osc7_windows_path"
+path = "tests/osc7_windows_path.rs"
+
 [dependencies.itoa]
 version = "1.0.15"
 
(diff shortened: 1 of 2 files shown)


psmux:
Merged. Thank you Garrett, this is exactly the kind of coverage the parser needed.

I ran the suite here before merging: all 6 tests green against the current parser, so they pin real behavior rather than aspirational behavior. The shapes you chose are the right ones too. Drive letter paths with both terminators, percent decoding, the UNC authority strip, the basename that `#{b:pane_path}` takes, and the empty payload guard are precisely the cases a Windows shell emits through `[uri]::AbsoluteUri`. The comment tying `#{pane_path}` to the cheap alternative for `#{pane_current_path}` (which walks a full toolhelp snapshot) is a genuinely useful note for future readers as well.

Appreciate you taking the time to write test only coverage with this much care. More of this is always welcome.

PR #540, opened 2026-08-06
fix(capture): honor -t pane targets and -N trailing-space preservation

krhougs:
## Context

We ship psmux as the tmux stand-in behind [tmex](https://github.com/krhougs/tmex)'s Windows gateway (tmex is an open-source terminal-multiplexer web frontend; Vibe X is its commercial distribution). The gateway drives psmux exclusively through control mode and rebuilds pane screens with `capture-pane -p -e -J -N -t %N` whenever a client reattaches or switches windows — exactly how iTerm2-style tmux integrations restore state.

## Problems

Two fidelity gaps surfaced on real user machines through that path:

1. **`-t %N` was accepted but ignored** on the one-shot capture paths: `CtrlReq::CapturePane/CapturePaneStyled/CapturePaneRange` carried no target, so every capture read the *active* pane of the *active* window. Restoring a background window painted it with another pane's content; the styled path also temporarily focus-switched to work around this, disturbing the active window.
2. **`-N` was parsed but never honored**: styled captures unconditionally trimmed trailing cells. Full-screen alt-screen TUIs (opencode, vim with themes) paint their background via colored trailing blanks — after trimming, a restored screen lost its entire background color and only text islands kept theirs.

## Fix

- Thread an optional pane target through the three capture requests; the server resolves `%N` across windows via the pane tree and falls back to the active pane when absent (old behavior preserved). Id-targeted captures no longer take the temporary-focus path, so background captures stop disturbing the foreground.
- Thread `preserve_trailing` through styled/plain/range rendering: with `-N`, rows render at full width including trailing styled blanks; without it, the tmux default trim is unchanged.

## Tests

`tests-rs/test_capture_pane_fidelity.rs`: 9 closed tests on a dummy PTY (no server, no spawn — per AGENTS.md) covering trailing-background preservation, fully-styled blank rows, cross-window `-t` targeting, absent-target fallback, and default-trim compatibility. The first commit allows the crate to compile and run these closed suites on non-Windows hosts too (dev convenience; no runtime behavior change on Windows).

Verified on Windows (all 9 pass) and macOS.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1637,7 +1637,9 @@ pub fn run_remote(terminal: &mut Terminal<crate::platform::PsmuxBackend>, input:
     let mut paste_stage2_last_len: usize = 0;
     // Suppression window: after right-click copy, discard text key events
     // for a short period to prevent VS Code ConPTY duplicate injection.
-    #[cfg(windows)]
+    // Declared on every platform because the right-click handler assigns it
+    // unconditionally; only the Windows code paths ever read it.
+    #[allow(unused_variables)]
     let mut paste_suppress_until: Option<Instant> = None;
 
     // Track whether a modified Enter Press was already handled this keypress
--- a/src/copy_mode.rs
+++ b/src/copy_mode.rs
@@ -605,14 +605,39 @@ fn capture_row_text(screen: &vt100::Screen, row: u16, cols: std::ops::Range<u16>
     out
 }
 
-pub fn capture_active_pane_text(app: &mut AppState) -> io::Result<Option<String>> {
-    let win = &mut app.windows[app.active_idx];
-    let p = match active_pane_mut(&mut win.root, &win.active_path) { Some(p) => p, None => return Ok(None) };
+/// Resolve the (window index, tree path) a capture should read.
+///
+/// An explicit `-t %N` pane id wins and is searched across every window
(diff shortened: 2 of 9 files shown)


psmux:
Merged. Thank you krhougs, both halves of this were verified here from every angle before landing.

What I ran:

1. Your 9 closed tests: all green on the branch.
2. Live E2E on this machine: with the branch build, a raw socket `capture-pane -p -N` returns all 24 rows at the full 80 column width while master trims to the longest text (66 columns). Cross window capture by pane id returns the targeted pane's content without the temporary focus dance.
3. tmux 3.4 parity (WSL): plain `capture-pane -p` trims to 67 columns, `capture-pane -p -N` returns the full 80. Your behavior matches exactly.

One gap I found while testing, now fixed on master as a follow up commit: the one shot CLI client rewrites `capture-pane` through a flag whitelist and it silently dropped `-N`, so the flag only worked through control mode or a raw socket. Your gateway drives psmux through control mode so you never hit it, but CLI users would have. `-N` is now forwarded standalone and in POSIX clusters like `-pN`.

The `-t %N` resolution inside the capture itself (instead of the temporary focus workaround) is the right architecture and it also stops background captures from disturbing the foreground window for attached clients. Nice work, and the dummy PTY test scaffolding is a pattern I expect we will reuse.

PR #541, opened 2026-08-06
conhost crash resilience: ConPTY sideloading, input-mode diagnostics, and positive-evidence server reaping

krhougs:
## Context

psmux runs as the persistent terminal backend for [tmex](https://github.com/krhougs/tmex)'s Windows gateway (Vibe X is tmex's commercial distribution); sessions are expected to survive for weeks on end-user machines, unattended.

## Problems

A single conhost crash cascaded into total session loss through two independent weaknesses:

1. **In-box ConPTY crash bugs.** conhost.exe only updates with the OS; specific builds carry crash bugs — 10.0.26100.8737 access-violates on teardown of some alt-screen TUI apps, taking the pane's console *and shell* with it. Windows Terminal, VS Code and WezTerm all escape such bugs by shipping their own conpty.dll + OpenConsole.exe; psmux had no way to.
2. **Post-crash misreaping.** After such a crash the startup reaper could kill a *live* server: a stale PID anchor alone was treated as proof of death, and an inconclusive port probe (timeout, transient error) was treated the same as an actively refused connection. Users lost every session that had in fact survived.

## Fix

- **ConPTY sideloading** (opt-in, zero behavior change by default): load order is `PSMUX_CONPTY_DLL` explicit override → `conpty.dll` next to the psmux executable, only when its matching `OpenConsole.exe` sits beside it → the system kernel32 implementation. Absolute paths only, so the default-DLL-search-order hijack concern does not apply.
- **`PSMUX_NO_WIN32_INPUT` diagnostic switch** (same shape as `PSMUX_NO_PASSTHROUGH`): create the pseudo console without `PSEUDOCONSOLE_WIN32_INPUT_MODE`, for isolating conhost faults tied to that mode — this is how we bisected the crash above.
- **Reaping tightened to positive evidence only**: a stale PID anchor alone never reaps; only an actively refused probe may prove a server dead. Inconclusive probes leave the server alone.

## Tests

Extended `tests-rs/test_startup_stale_port_tax.rs` pins the reaping rules; new `tests-rs/test_conpty_env_flags.rs` locks the env-switch branches (no DLL loading in tests). Verified on Windows.

### DIFF
--- a/crates/portable-pty-psmux/Cargo.toml
+++ b/crates/portable-pty-psmux/Cargo.toml
@@ -97,6 +97,9 @@ version = "0.2"
 [dependencies.shell-words]
 version = "1.1"
 
+[dev-dependencies.parking_lot]
+version = "0.12"
+
 [dev-dependencies.futures]
 version = "0.3"
 
--- a/crates/portable-pty-psmux/src/win/psuedocon.rs
+++ b/crates/portable-pty-psmux/src/win/psuedocon.rs
@@ -178,6 +178,14 @@ fn supports_passthrough_mode() -> bool {
 /// always starts on a fresh screen, so inheriting the host cursor row buys
 /// nothing here.
 fn base_flags() -> DWORD {
+    // `PSMUX_NO_WIN32_INPUT=1` drops PSEUDOCONSOLE_WIN32_INPUT_MODE — a
+    // diagnostic escape hatch for conhost builds that crash around it.
+    if std::env::var("PSMUX_NO_WIN32_INPUT")
+        .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+        .unwrap_or(false)
+    {
+        return PSEUDOCONSOLE_RESIZE_QUIRK;
+    }
     PSEUDOCONSOLE_RESIZE_QUIRK | PSEUDOCONSOLE_WIN32_INPUT_MODE
 }
 
--- a/crates/portable-pty-psmux/src/win/psuedocon.rs
+++ b/crates/portable-pty-psmux/src/win/psuedocon.rs
@@ -395,3 +395,7 @@ mod tests {
         assert_ne!(base_flags() & PSEUDOCONSOLE_WIN32_INPUT_MODE, 0);
     }
 }
+
+#[cfg(test)]
+#[path = "../../../../tests-rs/test_conpty_env_flags.rs"]
+mod tests_conpty_env_flags;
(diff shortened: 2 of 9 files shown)


psmux (changes requested):
Thank you krhougs. There is a lot of genuinely good work in here and I want most of it. I cannot merge it as it stands though, because the reaping rule resurrects a measured regression on a whole class of hosts, including the machine psmux is primarily developed on. Requesting changes for that one point.

What I verified and like:

1. All 16 of your reaping tests and all 5 conpty env flag tests pass here.
2. `PSMUX_NO_WIN32_INPUT=1` sessions come up and round trip typing fine.
3. The startup PID re assert, the `process_exists` escalation for unopenable processes, the anchor repair path, and the sideload precedence design are all sound and I want them on master.

The blocker, with numbers from this machine:

This host never actively refuses a dead loopback port. A probe of three unbound ports each timed out at the full deadline with no RST (the stealth firewall behavior issue 7 batch D documented). Under your rule that only an active refusal may reap, that means no stale registry can ever be reaped here. I planted three fake stale registries (dead PID anchor, unbound port) and timed `psmux ls`:

master: first run 34ms, second run 15ms, all three reaped on the first pass by the dead anchor.
this branch: first run 2450ms, second run 1869ms, all three kept, forever, on every future invocation.

That is exactly the 2.4 second stale port tax the 448 campaign eliminated, and it comes back permanently because the kept files are re probed on every command.

Suggested middle ground that keeps your safety fixes intact: treat anchor dead plus every probe attempt timing out with zero response bytes as stale (the old batch D rule), while keeping timeouts inconclusive whenever the anchor is live, missing, or unopenable. Your field case (a busy live server reaped on refused false, timed out true) is already closed by your other fixes: the startup re assert means a live server's anchor names a live psmux process, so it takes the anchor alive fast path and is never probed at all. With the anchor dead AND nothing answering in three attempts, the two independent death signals together are as definitive as a refusal.

One smaller item: the PR description says `PSMUX_CONPTY_DLL` is absolute paths only, but `load_conpty` never checks `is_absolute()`, so a relative value like `conpty.dll` would resolve through the default DLL search order, which is the exact hijack scenario the old comment warned about. A one line guard (ignore non absolute values with a warning) closes the gap and makes the code match the description.

Happy to merge quickly once the reaping rule handles silent drop hosts. The conpty sideloading and diagnostics half could also land on its own if you prefer to split the PR.

PR #542, opened 2026-08-06
control mode: bounded buffering for slow clients, atomic command blocks, bounded output lines

krhougs:
## Context

[tmex](https://github.com/krhougs/tmex)'s Windows gateway (Vibe X is its commercial distribution) holds a long-lived control-mode connection to psmux and forwards pane output to remote web clients. The consumer side can stall for seconds at a time — network backpressure, a phone client on a weak link — which makes the control channel's robustness under a slow reader a production-critical property.

## Problems

1. **Slow readers were disconnected.** A control client that fell behind hit the connection's 5s write timeout and was dropped. For an integration this is catastrophic: the whole gateway loses every subscription and must resync from scratch — a stall became an outage.
2. **A command could stall the whole channel**, and error/success block framing wasn't atomic under concurrency (interleaved `%begin`/`%error` footers), so one late handler corrupted the line protocol for everything after it.
3. **Unbounded `%output` lines.** A single burst from a chatty pane produced one multi-hundred-KB notification line; real tmux emits one notification per PTY read. Line-based consumers allocate and parse per line, so these mega-lines caused latency spikes and buffer blowups.
4. **One-shot `capture-pane` could wait forever on the server response** — a wedged server hung the CLI and any script driving it, the one-shot flavor of the same stall class.
5. **The client's debug logger panicked on multi-byte output**: each OUT line was truncated to 200 raw bytes; any line whose 200th byte fell inside a multi-byte character — routine for TUI box-drawing — panicked the reader thread and killed the control client, observed as a command timeout followed by a reconnect loop.

## Fix

- Replace disconnect-on-slow with **bounded buffering + shedding**: below a soft backlog bound everything queues; past it, `%output`/`%extended-output` frames are shed (any client resynchronizes with `capture-pane`) while structural notifications (window/pane lifecycle) still queue; past a hard bound everything sheds. Queued bytes are accounted per client and drained symmetrically by the writer.
- Make each command handler complete with an explicit success/error response (`ControlCommandResponse`), keeping `%begin…%end/%error` blocks atomic and the channel responsive while a handler waits.
- **Split pane output into bounded notification lines** (4 KiB max, split on UTF-8 boundaries) matching tmux's one-notification-per-read shape.
- Bound the one-shot capture-pane wait; on expiry it reports `ERROR: capture-pane timed out` instead of hanging.
- Truncate the client debug log **on a char boundary** instead of a raw byte offset.

## Tests

`test_control_mode_liveness.rs` (handler timeout/error and block atomicity), `test_control_slow_client_backlog.rs` (9 tests: shed thresholds, structural-vs-output classes, cost accounting, non-blocking burst), `test_control_output_chunking.rs` (8 tests: bounds, UTF-8 boundary, reassembly, wire-level line bounds), `test_capture_pane_wait_bound.rs` (real connection handler over loopback: the 5s bound produces the timeout error), and a subprocess test driving the real client reader against a scripted control server with a mid-glyph 200th byte (`test_client_debug_log_truncation.rs`). All closed tests; no psmux server spawned. Verified on Windows.

### DIFF
--- a/src/control.rs
+++ b/src/control.rs
@@ -660,3 +660,11 @@ mod tests {
         assert_eq!(line, "%subscription-changed test_sub $1 @2 3 %4 : ");
     }
 }
+
+#[cfg(test)]
+#[path = "../tests-rs/test_control_slow_client_backlog.rs"]
+mod tests_control_slow_client_backlog;
+
+#[cfg(test)]
+#[path = "../tests-rs/test_control_output_chunking.rs"]
+mod tests_control_output_chunking;
--- a/src/control.rs
+++ b/src/control.rs
@@ -219,6 +219,38 @@ pub fn try_send_notification(client: &ControlClient, notif: ControlNotification)
     let _ = client.output_tx.send(output);
 }
 
+/// Maximum raw payload carried by a single %output / %extended-output line.
+/// tmux emits one notification per PTY read (bounded by its event-buffer read
+/// size), so clients never face a single multi-hundred-KB line. Splitting on
+/// UTF-8 boundaries keeps each chunk independently escapable.
+pub const OUTPUT_CHUNK_MAX: usize = 4096;
+
+/// Split pane output into chunks of at most `OUTPUT_CHUNK_MAX` bytes without
+/// cutting through a UTF-8 sequence.
+pub fn chunk_output(data: &str) -> impl Iterator<Item = &str> {
+    let mut rest = data;
+    std::iter::from_fn(move || {
+        if rest.is_empty() {
+            return None;
+        }
+        if rest.len() <= OUTPUT_CHUNK_MAX {
+            let chunk = rest;
+            rest = "";
+            return Some(chunk);
+        }
+        let mut split = OUTPUT_CHUNK_MAX;
+        while split > 0 && !rest.is_char_boundary(split) {
+            split -= 1;
+        }
+        if split == 0 {
+            split = rest.len();
+        }
+        let (chunk, tail) = rest.split_at(split);
+        rest = tail;
+        Some(chunk)
+    })
+}
+
 /// Emit the initial state burst to a freshly-attached control mode client.
 /// This mirrors what real tmux sends right after `-CC attach` so that
 /// integrations like iTerm2's tmux mode can bootstrap their UI without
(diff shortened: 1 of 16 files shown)


psmux (changes requested):
Thank you krhougs. This is the most substantial of the batch and the core of it holds up very well under testing here. Two things need a pass from you before it can land, so requesting changes, but both are cheap.

What I verified on this machine (on top of current master, details below on why that matters):

1. All 79 closed control tests green: backlog shedding bounds, chunking on UTF8 char boundaries, atomic block formatting, the send keys `-H` oneshot completion fix, and the warm spawner generation logic.
2. The debug log truncation child process test passes in 1.5s and the 5 second capture wait bound test passes at exactly the bound.
3. Your `test_control_mode_liveness.ps1` on a debug build: 10 of 11 pass, including the injected 8 second warm ConPTY delay with `display-message` and `capture-pane` staying responsive, error blocks not poisoning the socket, and half closed input draining every completed block.

The two items:

1. The last assertion of your own E2E suite fails here: a non reading control client with a 1KB receive buffer drained 4.5MB and its socket simply stayed open. That matches the code, which deliberately removed the write timeout (the comment says a slow reader must be buffered, not disconnected, and a dead peer surfaces via TCP reset). But the test still expects the client to be disconnected after the bounded socket write timeout, so either the test is asserting a behavior you removed on purpose (then fix the test) or you intended a bounded disconnect for hard stuck readers past the hard shed threshold (then the code is missing it). As written, the writer thread for such a client can sit in `write_all` indefinitely while everything sheds, which is survivable but never reclaims the thread until TCP teardown. Please make the code and the test agree.

2. The branch depends on PR 544 without declaring it: `test_client_debug_log_truncation` steers the child through `PSMUX_DATA_DIR`, which did not exist on the master this branch is based on. Standalone, that test does not fail cleanly, it hangs (I gave it ten minutes). PR 544 is merged now, so a rebase onto current master fixes this for free; you will pick up a small conflict in the test module registrations at the bottom of connection.rs.

On the substance: replacing disconnect on slow with bounded accounting plus shedding of resynchronizable frames, keeping structural notifications, making blocks atomic through a single writer, chunking output at 4096 on char boundaries, and moving warm pane creation off the server thread with a generation guarded worker are all the right calls, and the liveness suite proves the big one (command responsiveness during ConPTY creation) beautifully. The char boundary truncation fix alone kills a real crash class. Once the two items above are addressed I will run the full battery again and merge promptly.

PR #543, opened 2026-08-06
pty: per-pane write queue and transient write-error resilience

krhougs:
## Context

Through [tmex](https://github.com/krhougs/tmex)'s Windows gateway (Vibe X is its commercial distribution), remote clients paste large blocks and stream keystrokes into panes whose foreground apps are sometimes wedged or slow to drain their console input — a state end users produce routinely.

## Problems

1. **A blocked PTY write froze the server.** Pane writes happened on the request path; when a pane's console buffer was full (wedged TUI, suspended app), `write()` blocked and took the whole server loop — every other pane and client — with it.
2. **A transient write error permanently killed the writer.** One failed write tore down the writer thread, so a pane whose console hiccuped once never accepted input again even after recovering.

## Fix

- **A dedicated writer thread per pane** with a queue in front: the server enqueues and returns immediately; order is preserved; the thread exits cleanly when the queue closes.
- The writer **survives transient write errors**: a failed write is dropped (input is lossy-by-nature at this boundary — the alternative is blocking or death), the inner writer is kept, and subsequent writes proceed.

## Tests

`test_pane_writer_queue.rs` (delivery order, non-blocking enqueue while the inner writer is gated, clean thread exit) and `test_pane_writer_transient_error.rs` (a dummy writer injecting one failure: the thread must survive, keep the inner writer, and deliver subsequent writes). Closed tests, no server. Verified on Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1335,6 +1335,7 @@ dependencies = [
  "libc",
  "log",
  "nix 0.31.3",
+ "parking_lot",
  "serde",
  "serde_derive",
  "serial2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,9 @@ path = "tests-rs/test_pr469_osc_command_safety.rs"
 name = "test_issue494_copymode_freeze"
 path = "tests-rs/test_issue494_copymode_freeze.rs"
 
+[dev-dependencies]
+parking_lot = "0.12"
+
 # Windows only
 [target.'cfg(not(windows))'.dependencies]
 [target.'cfg(windows)'.dependencies]
--- a/src/pane.rs
+++ b/src/pane.rs
@@ -61,9 +61,14 @@ impl std::io::Write for QueuedPaneWriter {
     }
 }
 
-/// Wrap a raw PTY writer in a queue drained by a dedicated thread. The thread
-/// exits when the queue side is dropped or the underlying pipe write fails
-/// (child gone), after which queued writes report `BrokenPipe`.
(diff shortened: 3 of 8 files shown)


psmux:
Merged. Thank you krhougs. The queue architecture is right and it matches what tmux gets for free from libevent bufferevents, so this closes a real structural gap on Windows.

What I verified here before landing:

1. All 7 of your closed tests green: ordering, immediate completion under a gated inner writer, clean teardown on queue drop, and the transient failure keeping the inner writer alive.
2. Typing round trips end to end through the new queue on a live session, and split, zoom, and popups all behave.
3. I tried hard to reproduce the original server freeze on this host (300KB and then 3MB floods into a pane whose child sat in `Start-Sleep` and never read input) and could not wedge master; conhost kept draining here. I take your word that wedged or suspended children on real user machines block the pipe, and the fix is correct regardless since the enqueue path can no longer block the server loop by construction.

Two notes, neither blocking:

1. The PR text says a failed write is dropped and subsequent writes proceed, but the code sets `broken = true` permanently and your own test asserts later writes are consumed without delivery. So after one write error the pane keeps its window and its PTY handle (a strict improvement over the old thread death that EOFed the child's stdin and closed the window) but it never accepts input again. If the error really can be transient on your machines, consider retrying on the next queued buffer instead of latching; a permanently broken pipe would just fail again harmlessly. Happy to take that as a follow up if you see recovery in the field.
2. The queue is unbounded, which is the same contract as tmux's evbuffer, so I am fine with it; worth remembering if a runaway paste into a wedged pane ever shows up in a memory profile.

Good, focused change. Appreciated.

PR #544, opened 2026-08-06
tmux parity batch: warm-claim sizing, refresh-client flag rejection, PSMUX_DATA_DIR override

krhougs:
## Context

A batch of small tmux-parity and hygiene fixes found while running psmux under [tmex](https://github.com/krhougs/tmex)'s Windows gateway (Vibe X is its commercial distribution), which drives psmux the way iTerm2 drives tmux: control mode plus one-shot CLI commands.

## Problems and fixes

1. **`new-session -x/-y` sizes were discarded when a warm server claimed the session** — the pre-spawned warm server kept its default size, so the first window came up 120x30 regardless of what the caller asked. The claim path now applies the requested size.
2. **Control-only `refresh-client` flags (`-C`/`-B`/`-A`/`-f`) were silently accepted on the CLI path** — tmux errors on these outside control mode. They are now rejected with the tmux error text, and the server's error actually reaches the CLI user on stderr with a non-zero exit (previously it was swallowed).
3. **`PSMUX_DATA_DIR`**: every data-directory consumer resolved the root on its own, mixing `USERPROFILE`/`HOME` and hardcoded separators. All consumers now route through `paths::psmux_dir()`, and an absolute `PSMUX_DATA_DIR` takes precedence — analogous to tmux's `TMUX_TMPDIR`, and what lets an embedding application keep server state, logs and session files inside its own root instead of the user profile. Relative/empty values are rejected.

## Tests

`test_refresh_client_flags.rs` drives the real connection handler over loopback sockets (each control-only flag is rejected with tmux wording; flag-free and `-S` still forward). `test_data_dir_override.rs` locks every derived path under the override root, separator trimming, rejection of relative/empty values, and the home fallback (env access serialized with a lock + drop-guard restore). Verified on Windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,9 @@ path = "tests-rs/test_pr469_osc_command_safety.rs"
 name = "test_issue494_copymode_freeze"
 path = "tests-rs/test_issue494_copymode_freeze.rs"
 
+[dev-dependencies]
+parking_lot = "0.12"
+
 # Windows only
 [target.'cfg(not(windows))'.dependencies]
 [target.'cfg(windows)'.dependencies]
--- a/src/debug_log.rs
+++ b/src/debug_log.rs
@@ -29,10 +29,7 @@ use std::sync::atomic::{AtomicU32, Ordering};
 
 /// Resolve the psmux data directory (`~/.psmux/`).
 fn psmux_dir() -> String {
-    let home = std::env::var("USERPROFILE")
-        .or_else(|_| std::env::var("HOME"))
-        .unwrap_or_default();
-    format!("{}/.psmux", home)
+    crate::paths::psmux_dir()
 }
 
 /// Open a log file in the psmux data directory, creating the directory if needed.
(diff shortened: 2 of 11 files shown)


psmux:
Merged. Thank you krhougs, all three fixes verified here end to end before landing.

What I ran:

1. Warm claim sizing, A against B on this machine: with a warm server present, master gave a `new-session -d -x 200 -y 50` session a 120x29 window; your branch gives exactly 200x50. Your 13 Rust tests are green too.
2. `refresh-client -C` from the one shot CLI: master silently exits 0 doing nothing; your branch prints `psmux: refresh-client -C: not a control client` to stderr and exits 1, which matches the wording tmux uses in cmd refresh client for a non control client.
3. `PSMUX_DATA_DIR`: pointed it at a scratch directory and the entire registry (port, key, sid, pid, servers, instances, session id counter) landed under the override with the session fully reachable. The absolute only rejection also behaves as documented. This unlocks properly isolated test environments for us as well, which is a bigger deal than the feature itself.

One housekeeping note: the branch added `parking_lot` to `[dev-dependencies]` in Cargo.toml but did not commit the matching Cargo.lock update, so the first build after merge regenerated it. I committed the lock refresh on master right after the merge. Worth a `git add Cargo.lock` next time so CI and fresh clones do not see a dirty lockfile.

Routing every data directory consumer through `paths::psmux_dir()` also killed several hand rolled `USERPROFILE`/`HOME` resolutions that had drifted apart. Quality batch, appreciated.

Issue #360, opened 2026-06-07
Mouse wheel does not enter copy/scroll mode despite `scroll-enter-copy-mode on`


Kstheme:
## Mouse wheel does not enter copy/scroll mode despite `scroll-enter-copy-mode on`

  ### Environment

  - OS: Windows 10/11
  - Terminal: Windows Terminal / PowerShell
  - psmux version: `<please fill in: psmux --version>`
  - Shell inside psmux: PowerShell

  ### Description

  Mouse wheel scrolling does not work in a normal psmux pane. It only works after manually entering copy/scroll
  mode with `Prefix + [`.

  I have mouse support enabled, and I also tried enabling tmux-style automatic copy-mode entry:

  ```tmux
  set -g mouse on
  set -g scroll-enter-copy-mode on

  However, scrolling with the mouse wheel in a normal pane does not enter copy mode or scroll the pane history.

  ### Expected behavior

  When mouse is enabled, and especially when scroll-enter-copy-mode is enabled, using the mouse wheel in a pane
  should automatically enter copy/scroll mode and scroll through the pane history, similar to tmux behavior.

  ### Actual behavior

  Mouse wheel scrolling has no effect in the normal pane view.

  If I manually enter copy/scroll mode with:

  Prefix + [

  then the mouse wheel works and scrolls through history correctly.

  ### Steps to reproduce

  1. Create or update ~/.psmux.conf with:

     set -g mouse on
     set -g scroll-enter-copy-mode on
     set -g history-limit 10000

  2. Start a new psmux session.
  3. Run a command that produces more output than fits on screen, for example:

     1..200

  4. Try to scroll up with the mouse wheel in the normal pane.
  5. Enter copy mode with Prefix + [.
  6. Try scrolling again with the mouse wheel.

  ### Observed result

  - Mouse wheel does not scroll in normal pane mode.
  - Mouse wheel does scroll after manually entering copy/scroll mode.

  ### Notes

  The documented options list includes mouse, but I could not find scroll-enter-copy-mode in the supported
  options. If this option is not currently supported, it would be useful to either support it or document that
  mouse wheel scrolling only works after entering copy/scroll mode manually.

```tmux
PS C:\Users\34220>psmux --version
tmux 3.3.5
```

psmux:
Hey @Kstheme, thank you for the detailed report and the exact reproduction steps. You are right, and I reproduced it.

[...]

Issue #277, opened 2026-05-01
Scroll Issue


mejares-jamesmichael:
Hello, I am encountering a mouse scroll issue while inside a psmux session.  I typically use nvim and opencode both in a split pane session.
In the previous versions, I can still scroll at opencode instance running inside of psmux. But I just recently updated my psmux version and now that is not the case anymore.

Thanks!

psmux:
Hey @mejares-jamesmichael, thanks for reporting this and for the context about opencode + nvim in split panes.

You were right that this was a regression. I tracked it down to a change in how psmux injects mouse events into ConPTY. The previous approach used Win32 `MOUSE_EVENT` records via `WriteConsoleInputW`, which worked universally. A later commit switched to writing SGR mouse escape sequences directly to the ConPTY input pipe, which works great for VT aware apps (nvim, crossterm based TUIs) but breaks for Go/Bubble Tea apps like opencode.

Here is why: opencode enables `ENABLE_VIRTUAL_TERMINAL_INPUT` for keyboard handling, and when that flag is on, ConPTY passes SGR mouse sequences through as `KEY_EVENT` text instead of converting them into `MOUSE_EVENT` records. Since Bubble Tea reads mouse input via `ReadConsoleInput` (expecting `MOUSE_EVENT` records), the scroll events were silently lost.

**The fix** (commit 9c46e09):

1. For wheel/scroll events on native ConPTY children, psmux now injects **both** the SGR sequence to the PTY pipe (so VT apps like nvim still work) **and** a Win32 `MOUSE_EVENT` record via `WriteConsoleInputW` (so console API apps like opencode receive the scroll). Only applied to wheel events to avoid duplicating click/drag for apps where ConPTY already handles the conversion.

2. Scroll coordinates now use the center of the pane instead of position (0,0), since some TUI frameworks ignore events at the top left corner if it falls outside their scrollable viewport.

This is on `master` now. You can build from source or grab the next release. Would love to hear if it fixes things on your end.

mejares-jamesmichael:
Hey man @psmux , I  just built from source with the new change you've just implemented and I can see that the issue is now resolve!

It works flawlessly now. Thanks for the very quick fix, really appreciate it!

psmux:
Awesome @mejares-jamesmichael . If you like psmux, please support by sharing good news about it. thank you!

Issue #51, opened 2026-02-26
Mouse scrolling


meluskyc:
Any plans to support mouse scrolling. Thanks!

Thank you for this great project. 🙂

tscodeplus:
Can't agree with you more. Having mouse scrolling will be very handy!
With the latest fixes, Claude in psmux seesion works well now. Mouse scrolling will make it very strong and easy to use. @computersrmyfriends @marlocarlo

psmux:
Hey @meluskyc and @tscodeplus, happy to share that **mouse scrolling is now fully implemented** as of commit e937a92!

Here's what works now:

- **Scroll up** at a shell prompt automatically enters **copy mode** (same as tmux) and scrolls through the scrollback buffer
- **Scroll down** in copy mode scrolls back toward live output, and auto-exits copy mode when you hit the bottom
- **Scroll inside TUI apps** (htop, vim, Claude CLI, etc.) gets forwarded directly to the application
- **Mouse text selection** with click and drag highlights text in blue and copies it to clipboard on release
- **Right-click** pastes from clipboard (PowerShell style), or gets forwarded to TUI apps when they're running
- **Copy mode mouse selection** lets you click to position the cursor, drag to select, and auto-yanks on release (matching tmux behavior)

All the standard tmux mouse behaviors are covered. Just make sure `set -g mouse on` is in your config (it's on by default).

Thanks for the kind words about the project, that really means a lot! If you're enjoying psmux, please share it with anyone who might find it useful. Stars and word of mouth go a long way in helping the project reach more people!

Keinns:
latest 0.4.3 ,mouse scroll not working ,tested everything i can . env :windows 10 , powershell7  with .psmux.conf (set -g mouse on)， and tried change the windows terminal settings.json with "scrollbarState": "hidden".  also not working.
copy mode works fine thouth (after enter copy mode , mouse scroll up /down works fine )

psmux:
@wukeying  You're trying the older version without mouse scroll support. Please try 0.4.5

tscodeplus:
Works with v0.4.5
---- Replied Message ----
***@***.***>Date02/28/2026 16:51 ***@***.***> ***@***.***>,
***@***.***>SubjectRe: [marlocarlo/psmux] Mouse scrolling (Issue #51)
wukeying left a comment (marlocarlo/psmux#51)
[...]

PR #548, opened 2026-08-08
fix(mouse): wheel over inline TUIs (pi) cycled prompt history instead of scrolling

PhotoAtomic:
## Symptom

Inside psmux under Windows Terminal, with an inline TUI like [pi](https://github.com/earendil-works/pi) focused on its input box, the mouse wheel over the conversation transcript **cycles the input box's prompt history** instead of scrolling — regardless of where the mouse pointer is. Copy mode is never entered; the only workaround is pressing PgUp first. Outside psmux the same app scrolls fine, and it also scrolls fine on psmux **3.3.4**.

## Version matrix / bisect

| psmux | wheel over pi |
|---|---|
| 3.3.3 | ✅ scrolls (copy mode) |
| 3.3.4 | ✅ scrolls (copy mode) |
| 3.3.5 | ❌ cycles prompt history |
| 3.3.6 / 3.3.7 | ❌ cycles prompt history |

`git bisect` names b64408ed (`fix(mouse): restore heuristic fallback in pane_wants_mouse for TUI apps (#285)`) as the first bad commit; 7d6300b7 later compounded it with a different mechanism producing the same symptom.

## Root cause (two stacked issues on the live wheel path)

The attached client sends `pane-scroll` → `handle_pane_scroll`. There:

1. **The wheel gate trusted `mouse_protocol_mode`** (`pane_wants_scroll_forward`, from #360). But PSReadLine spuriously enables mouse tracking on ConPTY, so *any pane whose shell is pwsh* reports an active mouse protocol even when the foreground app never asked for the wheel — and the wheel got injected into the app. The 3.3.4 wheel handlers documented exactly this false positive; b64408ed removed that rationale along with the code.

2. **The "general alternate-scroll" fallback** (7d6300b7) translated each wheel notch into 3× arrow keys for *any* confirmed non-shell foreground. pi's focused input box reads Up/Down as history navigation — hence the history cycling. In tmux/xterm, alternate-scroll (DECSET 1007) only ever applies to **alternate-screen** panes; blanket arrows on the main screen are not tmux parity.

## Fix

Restores the known-good 3.3.4 wheel semantics on all wheel paths:

- New `pane_in_alt_screen()` helper is the single wheel-forwarding gate; `handle_pane_scroll`, `remote_scroll_wheel`, and the `input.rs` wheel arms all use it. Real wheel consumers (nvim, htop, less, opencode) run on the alternate screen, which modern ConPTY reports reliably. Click/drag/hover gates are untouched.
- Drop the general arrow-key fallback; keep the exact-allowlisted `more.com` Enter-advance special case.
- Remove the now-unused `pane_wants_scroll_forward`.

## Verification

New E2E test `tests/test_pi_wheel_copymode.ps1` drives a real attached client in classic conhost with real `WriteConsoleInput` wheel injection against a pane simulating pi's exact profile (non-shell foreground + spurious mouse protocol + filled screen with bottom input box), asserting copy mode turns on **and** nothing leaks into the app's stdin. A/B results:

| binary | copy mode | input leaked to app |
|---|---|---|
| official 3.3.7 | ❌ never | `ESC[A` ×3 per notch |
| official 3.3.4 | ✅ | none |
| this branch | ✅ | none |

`tests/test_issue360_mouse_wheel_copymode.ps1` (filled-shell wheel-up enters copy mode) still passes. The reporter also confirmed the fix interactively with the real pi app.

🤖 Co authored with [Claude Code](https://claude.com/claude-code)

## Related issues

No pre-existing psmux issue describes this exact bug. Adjacent pi-side reports about transcript scrolling/focus exist (earendil-works/pi#5745 — overlay focus API, earendil-works/pi#1891 — anchored-input feature request), but they are distinct from this regression: here pi behaves correctly outside psmux and on psmux ≤ 3.3.4, so the fault is psmux's wheel routing.

Prior psmux issues in the same area, already discussed above: #360 (wheel vs filled shell), #285 (heuristic fallback that started the regression window), #277/#295 (scroll for inline Bubble Tea TUIs).

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -2756,14 +2756,11 @@ pub fn handle_mouse(app: &mut AppState, me: MouseEvent, window_area: Rect) -> io
                 win.active_path = path.clone();
                 active_area = Some(*area);
             }
-            // Forward scroll to child if pane wants mouse events (real TUI app
-            // like nvim/htop).  If not (shell prompt), auto-enter copy mode.
-            //
-            // Uses pane_wants_mouse() which includes heuristic fallback for
-            // older Windows 10 builds where ConPTY strips DECSET 1049h.
-            // (fixes #285)
+            // Wheel gate: alternate_screen ONLY — known-good pre-3.3.5
+            // semantics.  pane_wants_mouse() must not be used for the wheel
+            // (see pane_in_alt_screen for the rationale).
             let child_in_alt = active_pane(&win.root, &win.active_path)
-                .map_or(false, |p| crate::window_ops::pane_wants_mouse(p));
+                .map_or(false, |p| crate::window_ops::pane_in_alt_screen(p));
             if child_in_alt {
                 if let Some(area) = active_area {
                     if let Some(active) = active_pane_mut(&mut win.root, &win.active_path) {
@@ -2801,10 +2798,10 @@ pub fn handle_mouse(app: &mut AppState, me: MouseEvent, window_area: Rect) -> io
                 win.active_path = path.clone();
                 active_area = Some(*area);
             }
-            // Forward scroll-down to child only if pane wants mouse events.
-            // Uses pane_wants_mouse() with heuristic fallback. (fixes #285)
+            // Wheel gate: alternate_screen ONLY — known-good pre-3.3.5
+            // semantics (see pane_in_alt_screen for the rationale).
             let child_in_alt = active_pane(&win.root, &win.active_path)
-                .map_or(false, |p| crate::window_ops::pane_wants_mouse(p));
+                .map_or(false, |p| crate::window_ops::pane_in_alt_screen(p));
             if child_in_alt {
                 if let Some(area) = active_area {
                     if let Some(active) = active_pane_mut(&mut win.root, &win.active_path) {
--- a/src/window_ops.rs
+++ b/src/window_ops.rs
@@ -248,27 +248,22 @@ pub(crate) fn pane_wants_mouse(pane: &Pane) -> bool {
     is_fullscreen_tui(pane)
 }
 
-/// Stricter than `pane_wants_mouse`, used ONLY for the scroll-wheel decision.
+/// Wheel-forwarding gate: is the pane's child on the alternate screen?
 ///
-/// The wheel must auto-enter copy mode for an ordinary shell pane (tmux parity,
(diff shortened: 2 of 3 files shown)


z-ph:
我也遇到了同样的问题,这个pr很关键

mejares-jamesmichael:
Is this similar with the OpenCode TUI scroll issue before?

PhotoAtomic:
> Is this similar with the OpenCode TUI scroll issue before?

Not sure, i very rarely use open code in psmux, what i've noticed anyway, Is that with the involvend versions of psmux, open code appears to behave well. But my tests was limited only in opening a long conversation and observe if the mouse wheel scrolls It up. A 2 seconds test, i would't call It reliable.

psmux:
Merged. This is one of the best bug reports and fixes this repo has received. The version matrix, the bisect naming both offending commits, and the two stacked root causes were all exactly right.

I reproduced it independently before reading a line of your diff: on current 3.3.8 semantics your own harness leaked `ESC[A` three times per wheel notch straight into the app's stdin (precisely what cycles a focused input box's prompt history) and copy mode never engaged. On your branch the same harness enters copy mode with zero bytes leaked. A cleaner A/B is hard to imagine.

One thing surfaced during integration: the alternate screen only gate regressed the issue #570 suite, where a genuine main screen mouse consumer (a child that really enables mouse tracking itself, the Copilot CLI shape) stopped receiving wheel events entirely, 20 failing checks. Your analysis of WHY the protocol signal lies (PSReadLine enables tracking spuriously and it lingers) pointed straight at the fix, so I kept your gate and added the missing discriminator in a follow up commit: psmux now records who was foreground when the DECSET transition happened. Enabled while the shell owns the prompt means spurious, wheel stays on your copy mode semantics; enabled by a confirmed non shell foreground means a real consumer, wheel is forwarded with true coordinates. That is tmux `mouse_any_flag` parity restored for honest apps while pi stays fixed.

Full sweep on the final tree: your `test_pi_wheel_copymode.ps1` 2/2, `#570` coordinates suite 0 failures, `#360` filled shell 2/2, `#285` nvim 10/10, `#51` 7/7, `#573` 5/5. Your removal of the blanket alternate scroll arrows fallback survives unchanged; that was never tmux behavior on the main screen, exactly as you argued, and the `more.com` allowlist stays as the one documented exception.

Thank you for the persistence and rigor on this one, and please pass my thanks to the pi users who confirmed the fix interactively. This ships in the next release.

Issue #473, opened 2026-07-14
OSC 4/10/11 terminal color queries are not answered inside psmux, causing incorrect theme detection


hcoona:
## Summary

On Windows/ConPTY, a pane application can issue terminal color queries without receiving usable replies from stock psmux. This prevents applications that require OSC 10, OSC 11, and OSC 4 palette responses from determining the terminal palette.

GitHub Copilot CLI is one affected application: under a light Windows Terminal scheme, it selects its dark fallback inside psmux while selecting its light variant when run directly. This appears to be a terminal color-query problem rather than an SGR rendering or `bold-is-bright` problem.

I built and tested a local prototype that restores the expected result. I am filing this issue primarily to share the observations, failed approaches, and tradeoffs. The prototype is one possible implementation, not a request that psmux adopt the same design.

## Environment

| Component | Version |
|---|---|
| psmux | 3.3.6 (`tmux 3.3.6`) |
| GitHub Copilot CLI | 1.0.71-0 |
| PowerShell | 7.6.3 |
| Windows Terminal | 1.24.11321.0 |
| Windows | Windows 11 Enterprise, build 26200 |
| Windows Terminal scheme | Solarized Light |

Copilot was configured with the `github` theme. Windows Terminal reported foreground `#657B83`, background `#FDF6E3`, and the standard Solarized Light 16-color palette.

## Minimal protocol probe

The following Node.js probe is independent of Copilot and theme rendering. It sends the light/dark and color queries, waits 2.5 seconds, and summarizes the responses observed on stdin.

<details>
<summary><code>terminal-color-query-probe.js</code></summary>

```js
const chunks = [];

if (process.stdin.isTTY) {
  process.stdin.setRawMode(true);
}
process.stdin.resume();
process.stdin.on("data", (chunk) => chunks.push(chunk));

const queries = [
  "\x1b[?996n",
  "\x1b]10;?\x1b\\",
  "\x1b]11;?\x1b\\",
  ...Array.from({ length: 16 }, (_, i) => `\x1b]4;${i};?\x1b\\`),
];
process.stdout.write(queries.join(""));

setTimeout(() => {
  const text = Buffer.concat(chunks).toString("utf8");
  const oscEnd = "(?:\\x07|\\x1b\\\\)";
  const palette = [
    ...text.matchAll(
      new RegExp(`\\x1b\\]4;(\\d+);rgb:[^\\x07\\x1b]+${oscEnd}`, "g"),
    ),
[...]

PR #557, opened 2026-08-09
workaround: disable color-query answering that races yazi's startup probe (fixes #556)

### LINKED ISSUE #556: yazi opens a "Shell:" popup with `rgb:0c0c/0c0c/0c0c` at startup inside psmux (3.3.7)
Follow-up on #59 (yazi probe family, fixed for the timeout case in 0.4.7) — this is a regression introduced by the #473 color-query answering.

**Bug**: yazi's startup probe (`yazi-emulator/src/probe.rs`) sends `\x1b]11;?\x07` (OSC 11 bg query) alongside DA1/kitty/etc. and reads responses in a short window that closes on DA1 or 3s. psmux answers OSC 4/10/11 / CSI ?996n from the server loop, so the response (`\x1b]11;rgb:0c0c/0c0c/0c0c\x1b\\`, Campbell default) lands after the window closes. yazi's event layer then re-parses it as an interactive `shell` action — a "Shell:" popup appears with `rgb:0c0c/0c0c/0c0c` as its input value, and it steals keystrokes.

**Repro**: `yazi` in a psmux pane on Windows — the popup shows immediately at startup and persists. Raw WezTerm / real tmux are clean.

**Evidence**:

1. Raw-mode capture test: the response bytes arrive intact at the pane input (`1B 5D 31 31 3B 72 67 62 3A ... 1B 5C`) — malformed bytes ruled out; this is a timing race.
2. Timing probe (headless, 3.3.7): sending `DA1 + OSC 11 + ?996n` returns only `\x1b[?1;0c` (the DA1 answer) in **33 ms** — ConPTY answers DA1 itself on the input path. yazi's probe window therefore closes ~33 ms after the bundle is sent, before psmux's color response can survive the pane-output → reader → server-loop → input round trip. **Even answering synchronously at detection cannot win this race**, because ConPTY's DA1 auto-answer is inherently faster than any psmux-mediated response.
3. Removing the answering block fixes the popup completely (verified). Unanswered queries make apps fall back to their defaults — the same as real tmux when a query is unanswered.

**Impact of the workaround**: apps that rely on OSC 10/11/4 + `?996n` answers (e.g. GitHub Copilot CLI, per #473) fall back to their defaults again. This is a tradeoff, not a clean fix.

**Suggested proper fix**: control the DA1 response ordering so color answers land inside the app's probe window — e.g. intercept/relay DA1 the way CPR is handled, or answer the whole probe bundle in one ordered pass. May interact with `PSEUDOCONSOLE_PASSTHROUGH_MODE` (#502 notes passthrough is why raw sequences reach psmux's parser).

Workaround branch kept for reference: `mixsoulgithub/psmux` \`fix/color-query-race-yazi-popup\` (PR #557 closed — the removal regresses #473, see impact note below).

mixsoulgithub:
Workaround for #556 (full diagnosis and timing evidence there). The #473 color-query answering delivers OSC 4/10/11 / CSI ?996n responses after the app's probe window has closed — ConPTY answers DA1 itself in ~33 ms, closing yazi's window first — so yazi re-parses the response bytes as an interactive `shell` action and shows a "Shell:" popup with `rgb:0c0c/0c0c/0c0c` in it.

This removes the answering block; the scanning/answering helpers stay as dead code for a follow-up.

**Known regression**: color-querying apps (GitHub Copilot CLI, per #473) fall back to their defaults again. A proper fix — answering within the probe window, e.g. ordering the DA1 relay like CPR — is described in #556; this PR is the interim workaround.

Verified: yazi starts clean in psmux (no popup), `;`/`:` shell input still works. Repro and byte-level evidence in #556.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1345,29 +1345,14 @@ pub fn run_server(session_name: String, socket_name: Option<String>, initial_com
                     }
                 }
             }
-            // Issue #473: answer terminal color queries (OSC 4/10/11, CSI ?996n)
-            // detected in pane output, so pane applications can discover the
-            // terminal palette.  Colors come from the attached client's report
-            // of its host terminal, else the Campbell defaults.
-            if crate::types::COLOR_QUERY_PENDING.swap(false, std::sync::atomic::Ordering::AcqRel) {
-                let colors = app.host_colors.clone()
-                    .unwrap_or_else(crate::types::HostColors::campbell);
-                for win in &mut app.windows {
-                    helpers::drain_color_queries(&mut win.root, &colors);
-                    for fp in win.floating.iter_mut() {
-                        let bits = fp.pane.color_query_pending.swap(0, std::sync::atomic::Ordering::AcqRel);
-                        if bits != 0 {
-                            helpers::answer_color_queries(bits, &mut *fp.pane.writer, fp.pane.child_pid, &colors);
-                        }
-                    }
-                }
-                if let Mode::PopupMode { popup_pane: Some(ref mut pane), .. } = app.mode {
-                    let bits = pane.color_query_pending.swap(0, std::sync::atomic::Ordering::AcqRel);
-                    if bits != 0 {
-                        helpers::answer_color_queries(bits, &mut *pane.writer, pane.child_pid, &colors);
-                    }
-                }
-            }
+            // Issue #473 color-query answering removed: answering OSC 4/10/11
+            // / CSI ?996n queries from pane apps races yazi's startup probe —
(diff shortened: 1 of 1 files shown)


mixsoulgithub:
Closing on our side: this workaround trades yazi's bug for the #473 regression (color-querying apps like Copilot CLI fall back to defaults). It is only correct for machines whose apps don't query colors — not a change upstream should merge. The full diagnosis + timing evidence and the proper-fix directions live in #556. The branch (fix/color-query-race-yazi-popup) stays on the fork for reference if a stopgap is ever wanted.

Issue #82, opened 2026-03-05
tmux parity: Zoom mode borders persist; split does not unzoom


spooki44:
## Summary
Two tmux parity differences when a pane is zoomed:
- Borders remain visible and interactable while zoomed. This allows dragging a divider to bring hidden panes back into view. tmux hides borders entirely while zoomed.
- Splitting while zoomed creates a "zoom-only" layout in psmux. tmux first unzooms, then performs the split in the original layout.

## Environment
- OS: Windows 11
- Terminal: Windows Terminal
- Shell: PowerShell 7
- psmux: 0.4.8 (default config)
- reference tmux: 3.6a (default config)

## Expected
When a pane is zoomed (tmux behavior):
- No pane borders are visible while zoomed.
- Pane borders are not interactable while zoomed, so hidden panes cannot be brought into view via border drag.
- If a split is initiated while zoomed, tmux first unzooms and then performs the split in the original layout.

## Observed
In psmux 0.4.8:
- While zoomed, pane borders remain visible.
- The border is interactable. Dragging the divider can bring the hidden pane back into view while still zoomed, producing a zoomed layout with 2 panes visible.
- If a split is initiated while zoomed, psmux performs the split inside the zoomed view. This produces a "zoom-only" layout showing only the newly split panes. Toggling zoom off restores the original pre-zoom layout.

## Reproduction
Repro 1: split while zoomed
1) Create a left/right split:
   - `<prefix> %`
2) Zoom the focused pane:
   - `<prefix> z`
3) Split top/bottom while still zoomed:
   - `<prefix> "`
4) Toggle zoom again:
   - `<prefix> z`

Repro 2: border visible and interactable while zoomed
1) Create a left/right split:
   - `<prefix> %`
2) Zoom the focused pane:
   - `<prefix> z`
3) Drag the left divider to the right with the mouse or use `<prefix> <left/right>` (#81)

### Result
- Repro 1:
  - Expected (tmux): step 3 unzooms first, then performs the split in the original layout. After step 3, three panes are visible.
  - Observed (psmux): step 3 splits inside the zoomed view. While zoomed, only the two new panes are visible. After step 4 (unzoom), three panes are visible.
- Repro 2:
  - Expected (tmux): no divider visible or draggable while zoomed.
  - Observed (psmux): divider remains visible and can be dragged to bring the hidden pane back into view while still zoomed. Unzoom restores the pre-zoom layout.

psmux:
[...]

PR #572, opened 2026-08-14
fix(split): honor -Z after splitting

MattKotsenas:
## Summary
- preserve `split-window -Z` through the CLI and both server command paths
- zoom the active pane after a successful tiled split, including detached and already-zoomed windows
- leave floating-pane splits as overlays instead of applying tiled-window zoom

## Testing
- `cargo test --all-targets -- --test-threads=4`
- isolated client/server checks for `split-window -Z` and `split-window -d -Z`

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -10,7 +10,7 @@ use crate::pane::{create_window, split_active, kill_active_pane};
 use crate::copy_mode::{enter_copy_mode, scroll_copy_up, switch_with_copy_save, paste_latest,
     capture_active_pane, save_latest_buffer};
 use crate::session::{send_control_to_port, list_all_sessions_tree};
-use crate::window_ops::toggle_zoom;
+use crate::window_ops::{toggle_zoom, unzoom_if_zoomed};
 
 /// Parse a popup dimension spec: "80" (absolute) or "95%" (percentage of term_dim).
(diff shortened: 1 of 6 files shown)


psmux:
Merged. Thank you Matt, this one is textbook quality.

Before touching your branch I reproduced the gap on 3.3.8: `split-window -v -Z` created the pane and left `window_zoomed_flag` at 0. I then checked your semantics against the tmux source itself (`cmd-split-window.c`, `window.c`): tmux does `window_push_zoom(always=1)` before the split and `window_pop_zoom` after, which zooms whatever pane is active at that point. Your implementation lands on exactly that behavior, including the two subtle corners:

* `split-window -d -Z` keeps the old pane active and zooms it, matching `pop_zoom` acting on the active pane
* floating pane splits stay overlays and skip tiled zoom

I wrote a dedicated 10 check verification suite (`tests/test_pr572_split_zoom.ps1`, now in master) covering CLI, raw TCP, `-P` print mode, detached, an already zoomed window, plain splits still permanently unzooming (the old `#82` parity), and an attached TUI window. All 10 pass. Your `split_succeeded` guard also does the right thing when a split fails for lack of space: no zoom happens, which my suite proves by exhausting a window.

Appreciated the clean CtrlReq plumbing through both server paths and the sharpened flag parity test. Great contribution.

PR #575, opened 2026-08-16
Add opt-in drag-to-copy inside mouse-aware applications

Adamkadaban:
## Summary

Add a new session option:

```tmux
set -g mouse-selection-force on
```

This allows psmux drag-to-copy to remain active inside applications that enable terminal mouse tracking, while preserving application scrolling, hovering, and ordinary clicks.

The option defaults to `off`, so existing behavior is unchanged.

## Motivation

psmux currently yields the complete mouse gesture to any pane whose application enables DEC mouse tracking. This is appropriate for applications that implement their own selection, but prevents users from combining:

- application-controlled wheel scrolling
- psmux client-side drag selection
- copy-on-release behavior

GitHub Copilot CLI is one example: it enables DECSET 1002/1003 for mouse tracking. Users must currently choose between Copilot scrolling and psmux drag-to-copy.

## Behavior

With `mouse-selection-force on`:

- Mouse wheel and hover events continue reaching the application.
- A left-button press is temporarily buffered.
- If the pointer moves, psmux owns the gesture and copies the selected text.
- If the pointer does not move, the press and release are replayed to the application as an ordinary click.
- Right and middle mouse behavior is unchanged.

## Compatibility

- Default: `off`
- Existing per-pane mouse-yield behavior remains unchanged unless explicitly enabled.
- `mouse-selection off` still disables psmux selection globally.
- `pwsh-mouse-selection` remains independent.

## Testing

- `cargo check`
- Targeted client-selection and option-plumbing unit tests
- Boolean-option and configuration parity tests
- ARM64 Windows release build
- Isolated namespace option round-trip validation
- Manually verified with Copilot CLI inside psmux:
  - wheel scrolling works
  - ordinary clicks work
  - drag selection copies on release

---

Developed and tested with assistance from GitHub Copilot CLI.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -130,6 +130,7 @@ Details worth knowing:
 | `status-interval` | Int | `15` | Status refresh (seconds) |
 | `mouse` | Bool | `on` | Mouse support |
 | `mouse-selection` | Bool | `on` | psmux's client-side drag selection. Set `off` to let in-pane TUI apps (opencode, nvim, etc.) handle their own mouse selection without psmux drawing on top |
+| `mouse-selection-force` | Bool | `off` | Keep psmux drag selection active in apps that request mouse tracking. Plain clicks are replayed to the app; drags are copied by psmux |
 | `scroll-enter-copy-mode` | Bool | `on` | Enter copy mode on mouse scroll (set `off` to disable) |
 | `pwsh-mouse-selection` | Bool | `off` | tmux-like release-copy selection with word/line multi-click and pane-clipped extraction |
 | `paste-detection` | Bool | `on` | Detect Ctrl+V paste from console host and send as bracketed paste (set `off` to let Ctrl+V reach child apps like neovim) |
--- a/src/client.rs
+++ b/src/client.rs
@@ -498,6 +498,14 @@ fn pane_wants_mouse_json(layout: &LayoutJson, pane_id: usize) -> bool {
     }
 }
 
+fn client_selection_owns_drag(
+    mouse_selection: bool,
+    mouse_selection_force: bool,
+    pane_handles_mouse: bool,
+) -> bool {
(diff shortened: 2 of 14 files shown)


psmux:
Merged. Thank you Adam, the deferred click replay design is elegant and it holds up under real gesture testing.

I could not verify this with a scripted sim: ConPTY eats DECSET written through WriteConsole, so a node child that prints `?1002h` never registers as mouse aware inside psmux. So my verification suite (`tests/test_pr575_mouse_selection_force.ps1`, now in master) drives a real attached client in classic conhost with `WriteConsoleInput` press, held move, and release records against real nvim with `set mouse=a`, where nvim maps every mouse event it receives to an append only log. That gives byte level proof of what the application actually saw. All 9 checks pass:

* default off: nvim receives the full press, drag, release gesture and psmux copies nothing (clipboard sentinel intact)
* force on: the drag is consumed, the selected text lands in the psmux buffer, and nvim's log stays empty
* force on: a motionless click is replayed to nvim as press then release
* the wheel keeps reaching nvim either way
* toggling the option off live restores yield behavior immediately

Option plumbing checks out end to end as a real boolean (catalog, `show-options`, config file, TCP set, dump state field), and your unit tests all pass.

One heads up: PR #548 merged alongside this one and reworked the wheel gate, since PSReadLine spuriously enables mouse tracking on ConPTY and poisoned that signal. I added a follow up that attributes each protocol enablement to whoever was foreground at the time, so an app that genuinely asks for the mouse itself (your Copilot CLI case) keeps receiving wheel events while spurious shell owned tracking does not. Your drag and click arbitration is untouched by all of it, and the full mouse suite is green on the final tree. If you hit any wheel oddity with Copilot CLI on the next release, please open an issue and tag me.

Really appreciate the compatibility table in the description and defaulting to off. Welcome aboard.

Issue #559, opened 2026-08-11
split-window with a pane-id -t target silently creates nothing and -P returns a non-existent pane id


ChrisonSimtian:
`split-window` with a pane-id (`%N`) as its `-t` target creates nothing, exits 0, and — with `-P -F '#{pane_id}'` — returns a **pane id that does not exist**. The same command targeted at `session:window` works correctly.

This looks like the next member of the #545 / #554 family: a command left off the target-validator's command list. #554 added `select-pane`; `split-window` appears to be still missing, and it fails worse than a silent no-op because it hands back a fabricated id that callers then use.

### Repro

```powershell
psmux -L repro new-session -d -s r

# initial pane is %1, window has 1 pane

# A) pane-id target
psmux -L repro split-window -t %1 -P -F '#{pane_id}'
#   -> prints %17   exit 0   ... and the window STILL has 1 pane

# B) window target
psmux -L repro split-window -t r:1 -P -F '#{pane_id}'
#   -> prints %2    exit 0   ... window now has 2 panes

psmux -L repro list-panes -t r:1 -F '#{pane_id}'
#   -> %1
#   -> %2          <- %17 from (A) is nowhere; it was never created
```

Observed on `tmux 3.3.7 psmux 3.3.7 (05cc5d4 2026-07-20)`, Windows 11 26100.

### Two separate problems

1. **The split is silently dropped.** Exit code 0, no diagnostic.
2. **`-P -F '#{pane_id}'` returns a fabricated id.** `%17` on a server whose only live panes are `%1` and `%2`. The pane counter is evidently incremented even though no pane is created, so the returned id is not merely stale — it never existed. A caller has no way to distinguish this from success short of re-listing panes and checking membership.

The second point is what makes this more damaging than #545's silent no-op: the documented "capture the new pane id" idiom actively lies.

### Real-world impact

`psmux-resurrect` never restores split panes. `scripts/restore.ps1:236` does exactly the documented thing:

```powershell
$paneIds += Get-PaneId (& $PSMUX split-window -t $initialPaneId -c $pDir -P -F '#{pane_id}' 2>&1)
```

`$initialPaneId` is a pane id, so every split is dropped and the plugin cannot detect it — it receives a plausible id back and proceeds. Restoring a saved environment gives you the right sessions, windows, window names, per-pane directories and relaunched processes, but every multi-pane window comes back with a single pane.

[...]

PR #577, opened 2026-08-16
fix: make psmux faithful for programmatic control - pipe-pane file sinks, verbatim buffers, foreground pane_current_command

### LINKED ISSUE #576: pipe-pane: the documented `cat > pane.log` sink can never capture anything (0-byte file, rc 0) — plus two more programmatic-control fidelity gaps
## Summary

The canonical tmux logging idiom — and the exact example shown in `docs/scripting.md` —

```
psmux pipe-pane -o "cat > pane.log"
```

creates the log file but never writes a single byte to it, and exits 0 as if it had worked. This is not a timing or quoting problem: **the idiom cannot work on Windows by construction**, and two silent-failure layers hide that from the user.

While porting a tool that drives psmux programmatically, I found two more fidelity gaps of the same "accepts the tmux command, silently does something different" class; they are included below because the fix set only makes sense together.

## Root cause 1: the sink shell makes `cat` a no-op

`pipe-pane` runs sink commands under PowerShell (`resolve_run_shell()`, `src/commands.rs`: `pwsh`/`powershell -NoProfile -Command`, falling back to `cmd /c`). In PowerShell, `cat` is the `Get-Content` alias — it does **not** read stdin. The sink process exits immediately; the `>` redirection has already created an empty file; the pane's reader thread has nobody to tee to. Result: a 0-byte file, forever.

Since the tee path (`src/pane.rs`, `PIPE_WRITERS`) feeds pane bytes to the sink's **stdin**, any sink that doesn't actually read stdin fails the same way. `cat` is just the most important instance, because it is *the* tmux logging idiom and it's the example in this repo's own docs — every Windows user following the docs hits this.

## Root cause 2: sink startup failures are swallowed

`CtrlReq::PipePane` spawns the sink with `c.spawn().ok()` (`src/server/mod.rs`). On failure:

- the error is discarded — CLI exits 0, stderr empty;
- a `PipePaneState { process: None, .. }` entry **is still recorded**, so the pane looks piped: a later `pipe-pane -o` toggles the phantom pipe *off* instead of arming a new one.

So both "the sink can't work" (cause 1) and "the sink never started" (cause 2) are indistinguishable from success at the CLI.

## Measured matrix (Windows 11 Pro 26100, x64)

| sink command | v3.3.7 (winget) | master 499bb4d |
|---|---|---|
| `cat > log` (docs example) | file created, 0 bytes | file created, 0 bytes |
| `$input \| Out-File -Encoding utf8 log` | file not created (#563, fixed on master) | works (57 bytes) |
| `& "C:\Program Files\Git\usr\bin\cat.exe" > log` | — | works (114 bytes) |

(The v3.3.7 column is worse because #563 — unreleased fix — additionally mangles any sink command containing spaces. On the released version, effectively every practical sink form fails silently.)

## Related fidelity gaps found the same way

- **Buffers are not byte-exact on the wire.** `load-buffer` a file containing `printf '%s' 'QUOTED ARG'`, then `paste-buffer`: the pane receives `printf %s QUOTED ARG` — quotes stripped, TAB collapsed. `show-buffer` shows the corruption happens at ingestion (the file contents are put on the control line as bare words and re-tokenized server-side), so any programmatic paste of shell commands executes something other than what was loaded. CLI `set-buffer` has the same defect.
- **`#{pane_current_command}` reports the root shell, not the foreground process.** With `cat` running in the pane, `display-message -p '#{pane_current_command}'` returns the shell name. tmux reports the foreground process; tools that gate on "is a shell in the foreground" (nested-session detection, completion heuristics) misjudge every non-shell foreground.

## Suggested fix

I have a PR ready with three commits: (1) in-process direct file sinks for the `cat > file` idiom plus loud startup failures, (2) a byte-exact buffer wire (`set-buffer -H <hex>`, same encoding as the existing `send -H`) so paste is verbatim, (3) foreground-process `#{pane_current_command}` via the existing deepest-descendant walker. Details, full test matrix, and real-consumer validation in the PR.

## Environment

- psmux 3.3.7 (winget `marlocarlo.psmux`) and master `499bb4d` built from source
- Windows 11 Pro 10.0.26100, x64; Windows PowerShell 5.1 and PowerShell 7.5 both present

quolu:
Fixes #576.

## Why

I drive psmux programmatically (an MCP server that gives AI agents persistent Windows terminals — previously tmux-over-WSL, now native psmux). Porting it surfaced three fidelity gaps where psmux accepts the tmux command but silently does something different. All three break any tool that scripts psmux the way tmux is scripted; all three are fixed here, each with regression tests, and the whole set is validated end-to-end by the real consumer (details at the bottom).

## Commit 1 — `fix(pipe-pane)`: service `cat > file` sinks in-process; stop swallowing sink startup failures

`pipe-pane -o "cat > pane.log"` (the docs' own example) always produced a 0-byte file with rc 0: sinks run under PowerShell, where `cat` is the `Get-Content` alias — it never reads stdin. On top of that `spawn().ok()` swallowed startup failures *and* recorded the dead pipe, so a later `-o` toggled a phantom pipe off.

- The plain-redirect idiom (`cat > path` / `cat >> path`, case-insensitive, path optionally quoted) is now serviced in-process: the reader thread tees raw ConPTY bytes straight into the file via the existing `PIPE_WRITERS` mechanism. Byte-faithful, no shell, no child process.
- Strictly literal local paths only: shell/tmux expansions (`$var`, backticks, `%var%`, `#I`, leading `~`) fall through to the shell sink; UNC in every spelling, mapped network drives (`GetDriveTypeW`), the `\\.\` namespace, and DOS device names (`cat > CON` opens the console DEVICE) are refused loudly — the open runs on the server's single event loop and must never stall or hit a device. `\\?\C:\...` stays allowed.
- Failures are loud: unopenable path → `ERROR: ...`, non-zero CLI exit (reply shape of #559/#566) + status bar mirror; failed shell spawn → status bar, not recorded; acceptance is answered before spawning (a cold AV scan of the shell image stalls CreateProcess for seconds, and holding the reply hostage to it turned success into a client-side timeout); `pipe-pane` mid-chain keeps executing the rest of the chain.

## Commit 2 — `fix(paste)`: buffer contents survive the wire byte-exactly

`load-buffer` + `paste-buffer` delivered `printf %s QUOTED ARG` for a file containing `printf '%s' 'QUOTED ARG'` — quotes stripped, TAB collapsed. The paste path itself was innocent: the *buffer ingestion* line put file contents on the control wire as bare words, and the server re-tokenized and re-joined them. Buffer contents now travel as `set-buffer -H <hex>` (the same byte-exact wire the repo already uses for `send -H`); the plain-word form remains for config/bind-key/control-mode compatibility, and a malformed/non-UTF-8 `-H` payload is rejected with an ERROR instead of being lossily decoded. The CLI `set-buffer` path had the same defect and is fixed the same way.

## Commit 3 — `fix(format)`: `#{pane_current_command}` reports the foreground process, not the root shell

With `cat` running in a pane, `display-message -p '#{pane_current_command}'` returned the shell name. The existing Ctrl+C router already had a deepest-descendant process walker; the format now shares it (`deepest_descendant()`), with the render path's 250ms snapshot reuse and a quiet fallback to the pane's own name. `pwsh → cmd → PING.EXE` now reports `PING`; idle panes still report the shell. Known limit (pre-existing, documented in the commit): MSYS2/Git-Bash children re-parent to a dead intermediate PID, so processes under an *interactive* Git Bash cannot be attributed by PPID walk.

## Tests

- Rust: `cargo test --bin psmux` full — **2853 passed / 0 failed** (new units for the sink parser/path gate and the hex wire).
- New PowerShell suites under the repo's `New-PsmuxTestEnv` isolation + dedicated `-L` namespaces: `test_pipe_pane_cat_file_sink.ps1` (14 checks), `test_paste_buffer_verbatim.ps1` (3), `test_pane_current_command_foreground.ps1` (5).
- Non-regression, all green (`$env:PSMUX_EXE`, pwsh 7): `test_issue440` 8/8, `test_issue482` 2/2, `test_issue563` 2/2, `test_issue564` 2/2, `test_named_buffers` 12/12, `test_issue264_paste_buffer_proof` 5/5, `test_issue264_paste_buffer_repro` 8/8; chained `pipe-pane ; rename-window X` verified over a raw control connection.

## Validated by a real consumer

With these three fixes (and nothing else), the aiterm-mcp test suite — which drives every one of these surfaces hard (pipe-pane logging as its completion-detection substrate, buffer-based paste as its only text-input path, `#{pane_current_command}` for shell-state gates) — went from 8 hard failures to fully green on native Windows, including a live end-to-end run: launch Claude Code in a psmux pane, dispatch a prompt via bracketed paste, capture the response through the pipe-pane file sink, and recover the transcript.

Happy to split this into three PRs if you prefer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,7 @@ windows-sys = { version = "0.61", features = [
     "Win32_Foundation",
     "Win32_System_Memory",
     "Win32_System_DataExchange",
+    "Win32_Storage_FileSystem",
 ] }
 
 [[bin]]
--- a/docs/scripting.md
+++ b/docs/scripting.md
@@ -609,7 +609,8 @@ psmux next-layout
 psmux previous-layout
 psmux select-layout tiled         # Apply a specific layout
 psmux clear-history
-psmux pipe-pane -o "cat > pane.log"
+psmux pipe-pane -o "cat > pane.log"    # Direct file sink (see below)
+psmux pipe-pane                        # Stop piping
 
 # Hooks (event callbacks) - see Hooks section below for full reference
 psmux set-hook -g after-new-window "display-message created"
@@ -790,6 +791,57 @@ bind-key e run-shell -b "my-helper.ps1 -Pane '#{pane_id}' -Path '#{pane_current_
 > swallowing spawn errors, such a bind failed completely silently. If you are
 > targeting an older psmux, pass the values from the caller instead of relying
 > on expansion.
+
+## Piping Pane Output (pipe-pane)
+
+Stream a pane's raw output (ANSI escapes included) to a file or a command,
+like tmux's `pipe-pane` (one deviation: tmux expands format variables such
+as `#I` in the sink command; psmux does not):
+
+```powershell
+# Log the pane to a file (direct file sink - see below)
+psmux pipe-pane -o "cat > pane.log"
+psmux pipe-pane -o "cat >> pane.log"     # append instead of truncate
+psmux pipe-pane -t %2 -o "cat > out.log" # pipe a specific pane
+
+# Stop piping (or use -o again to toggle off)
+psmux pipe-pane
+```
+
+**Direct file sink.** The canonical tmux logging idiom `cat > <path>` /
+`cat >> <path>` (path may be quoted) is serviced by the psmux server itself:
+the pane's raw ConPTY bytes are written straight to the file, byte-for-byte,
+with no shell in between. This exists because arbitrary sink commands run
+under PowerShell, where `cat` is the `Get-Content` alias — it never reads
+stdin, so the idiom cannot work as a shell command on Windows.
+
+Prefer an **absolute path**: a relative path resolves against the psmux
+server's working directory (the session's start directory), not your
+shell's. The path must be a literal **local file** path — anything a shell
+would expand (`$var`, `` `...` ``, `%var%`, `#I`, leading `~`) falls
+through to the shell sink, and UNC paths, mapped network drives, and DOS
+device names (`CON`, `NUL`, ...) are refused loudly: an unreachable host
+would stall the whole server, and a device is not a log file. To log to a
+network location, use a sink command that reads stdin — it runs as a child
+process and stalls only itself.
+
+**Arbitrary command sinks.** Any other command runs under the sink shell
+(`pwsh`/`powershell -NoProfile -Command`, falling back to `cmd /c`) and
+**must actually read its stdin** — that is where the pane bytes arrive. In
+PowerShell, read `$input` or `[Console]::In`:
+
+```powershell
+# PowerShell sink: line-oriented processing of pane output
+psmux pipe-pane -o '$input | Out-File -Encoding utf8 pane.log'
+```
+
+A sink that never reads stdin (like the `cat` alias) exits immediately and
+captures nothing.
+
+**Errors are loud.** If a direct file sink cannot be opened (e.g. the log
+file's directory does not exist), `pipe-pane` prints `ERROR: ...` and exits
+non-zero instead of silently recording a dead pipe. A shell sink that fails
+to spawn is reported on the status bar and is likewise never recorded.
 
 ## Interactive Choosers
 
--- a/src/format.rs
+++ b/src/format.rs
@@ -1348,8 +1348,10 @@ fn expand_var_inner(var: &str, app: &AppState, win_idx: usize) -> String {
                     }
                 }
                 if let Some(pid) = p.child_pid {
-                    // Fallback: upstream process-tree heuristic, then shell binary name.
-                    crate::platform::process_info::get_foreground_process_name(pid)
+                    // Fallback: the deepest foreground descendant (what tmux
+                    // reports — `cat` running under a shell, not the shell),
+                    // then the pane's own process, then a generic label.
+                    crate::platform::process_info::get_deepest_foreground_process_name(pid)
                         .or_else(|| crate::platform::process_info::get_process_name(pid))
                         .unwrap_or_else(|| "shell".into())
                 } else if !p.title.is_empty() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -2695,7 +2695,10 @@ fn run_main() -> io::Result<()> {
                 }
                 let mut cmd = "set-buffer".to_string();
                 if let Some(b) = buffer_name { cmd.push_str(&format!(" -b {}", b)); }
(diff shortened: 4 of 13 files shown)


psmux:
Merged in 0fc0720. This is one of the strongest contributions this project has received: the issue report alone was a full investigation, and the PR held up under an adversarial review that reproduced every claim independently before reading a line of the diff. All three fixes verified A/B on a live build: the `cat > file` idiom now captures real bytes with loud failures for unopenable, UNC, and device paths and no phantom pipe state; buffers came through byte identical up to the 4 MB stress files, including quotes, TABs, and space runs that master mangled; and `#{pane_current_command}` now answers `PING` and `sleep` for the deep chains where master froze at the first wrapper, while idle panes still report their shell. The tmux source confirms all three directions: the sink contract belongs to `/bin/sh -c` with pane output on the sink's stdin, buffer payloads are never re lexed after parse time, and `pane_current_command` is the tty's foreground process group with exactly the argv then shell fallback you kept.

I added one commit on top (0509351) with three review findings, all verified before and after:

1. Chained `pipe-pane -o "<sink>" \; <command>` answered correctly on the wire but exited 1 with a bogus "no response from server (timed out)". The freshly spawned sink child inherited a duplicate of the client's connection socket (children are spawned with handle inheritance on), so the server's close never became an EOF for a client that reads to EOF as end of reply. Master only dodged this because its broken sinks died instantly. The connection sockets, including every `try_clone`, are now scrubbed of HANDLE_FLAG_INHERIT; the failing shape went from a 3.1 second timeout to exit 0 in 65ms and a raw TCP probe shows EOF immediately after the reply.
2. `refuse_file_sink_path` let `nul:` through: the DOS style trailing colon is still the device to CreateFile. The stem check now trims it, with unit cases for `nul:`, `NUL:`, `COM1:`.
3. `test_pane_current_command_foreground.ps1` relied on `send-keys C-c` to stop a nested `cmd /c ping` and a bash `sleep` in a detached session, which ConPTY cannot deliver, so tests 3 and 4 failed deterministically in CI mode. They now end the foreground chains by pid; the suite runs 5 of 5.

Two nits I chose not to block on, for the record: a new CLI talking to an old server sends `set-buffer -H <hex>` that the old server stores as literal hex text (transient during upgrade windows only), and `hex_encode` allocates per byte via `format!` (harmless at buffer sizes that matter). Both are fair game for a follow up if you feel like it.

Thank you for the quality of this work. The Rust suite (2859 tests) and all three of your new PowerShell suites run green on the merged build; the full end to end sweep is running now and I will follow up here if it surfaces anything.

PR #584, opened 2026-08-19
SBAI-7534: expose base64 send-paste CLI transport

BizaNator:
Adds a one-shot CLI entrypoint for psmux's existing base64 send-paste control frame. This lets remote callers transport multiline UTF-8 without injecting newline-delimited control commands or nesting payloads inside PowerShell EncodedCommand.\n\nValidation:\n- cargo check --tests --target x86_64-pc-windows-msvc\n- valid base64 frame forwarded exactly once\n- missing and malformed payloads rejected\n\nParent: SBAI-7120

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -416,6 +416,42 @@ fn probe_session_alive(session_name: &str) -> bool {
     }
 }
 
+fn build_send_paste_control(cmd_args: &[&str]) -> io::Result<String> {
+    let mut payload: Option<&str> = None;
+    let mut i = 1;
+    while i < cmd_args.len() {
+        match cmd_args[i] {
+            "-t" => {
+                i += 1;
+                if i >= cmd_args.len() {
+                    return Err(io::Error::new(
+                        io::ErrorKind::InvalidInput,
+                        "send-paste -t requires a target",
+                    ));
+                }
+            }
+            value if payload.is_none() => payload = Some(value),
+            _ => {
+                return Err(io::Error::new(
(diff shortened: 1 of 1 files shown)


BizaNator:
Manager review complete at exact head daf238c: one-file diff is scoped to the existing send-paste control path; validation rejects missing, malformed, or multiple payloads; no plaintext logging or parallel transport. Fresh focused tests pass 3/3 for pmux, psmux, and tmux. The documented Windows-target cargo check --tests also passed on this same head. Ready for an upstream maintainer to merge; the Biloxi manager account lacks MergePullRequest permission in this external repository.

psmux:
Merged. Thanks for the tight scope on this one.

I verified the baseline first: on 3.3.8 the CLI silently exited 0 for a malformed payload, an empty payload, and multiple payloads (last one quietly winning). Your validation closes all three holes at the protocol boundary while keeping the frame format untouched.

My verification suite (`tests/test_pr584_send_paste_cli.ps1`, now in master) runs 10 checks against a live session: single line and multiline UTF-8 payloads land byte for byte in the pane, malformed and missing and duplicate payloads now exit non zero with a clear message, a dangling `-t` is rejected, payload before `-t` still works, and the raw TCP `send-paste` server path is unchanged. All 10 pass, plus your three unit tests.

The doc comment explaining why base64 is the right boundary (no second newline delimited command can be injected, and no PowerShell `EncodedCommand` nesting that hits the cmd.exe 8191 character wall) is exactly the kind of context that keeps this maintainable. Good work.

Issue #582, opened 2026-08-19
Unable to replace pwsh in `new-window` command


mrkingmidas:
OS: Windows 10 21H1 19043.1200 RU

Example for new session:
```
psmux -f %USERPROFILE%\.config\psmux\.psmux.conf new-session -d -s SN -- cmd.exe /k echo hello
```

Start cmd without `pwsh`. But if create new window with

```
psmux new-window -t SN -- cmd.exe /k echo hello2
```

hidden starts `pwsh`, that starts `cmd.exe`.
Why and how can start without `pwsh` (many resources use) ?

<img width="476" height="171" alt="Image" src="https://github.com/user-attachments/assets/9309026d-f85e-4556-ac0a-49650535b0b3" />

psmux:
Thanks for the clear report, and you were right on both counts: the two commands really did behave differently, and the hidden `pwsh` had no business being there.

I reproduced it exactly as you described. `new-session -d -s SN -- cmd.exe /k ...` spawned cmd.exe directly, while `new-window -t SN -- cmd.exe /k ...` interposed `pwsh.exe -NoLogo -Command "cmd.exe /k ..."` above it. It turned out to be broader than filed: `split-window` and the quoted string form of `new-session` took the same wrapper path. Only the initial session boot honored the `--` argv directly.

This is now fixed in master, commit 9bfdfc3, following tmux's own rule from `spawn.c`: a multi argument command after `--` is exec'd directly (tmux uses `execvp`), while a single string command goes through the shell. Concretely on the fixed build:

* `psmux new-window -t SN -- cmd.exe /k echo hello2` spawns cmd.exe as a direct child of the server. No pwsh anywhere in the chain.
* Same for `split-window ... -- cmd.exe ...` and `respawn-pane ... -- prog args`.
* `#{pane_pid}` now reports cmd.exe itself for these panes, matching tmux, instead of the wrapper pid.
* The quoted single string form, `new-window "cmd.exe /k echo hello3"`, still routes through the shell. That is tmux parity too: tmux runs single string commands via `$SHELL -c`.

So the answer to your question of how to start without pwsh: use the `--` form with the program and its arguments as separate words, exactly as you were already doing. On the next release it will do what you expected.

[...]

Issue #583, opened 2026-08-19
set-option parses flags after the option name: a dash-leading value is eaten — `-u` deletes the key at rc 0, and there is no `--` end-of-options (3.3.8)


dracic:
## Summary

`set-option` scans for flags **past the option name**, so a dash-leading VALUE is consumed as a flag instead of stored. Worst case is silent data loss: a value of `-u` **deletes the key at rc 0**. And there is no `--` end-of-options escape — `--` itself is rejected as an unknown flag — so there is no way to spell the intent at all. This is the `set-option` sibling of #562 (send-keys), which named the general pattern in its title.

Environment: psmux 3.3.8 (66cf613), Windows 11. All rows measured as raw argv (no shell quoting layer). tmux 3.4 parity measured in WSL.

## Measured

| command | psmux 3.3.8 | tmux 3.4 |
|---|---|---|
| `set-option -t S @k keepme` (control) | rc 0, value `keepme` | rc 0, value `keepme` |
| `set-option -t S @k -u` | **rc 0, key DELETED, empty stderr** | rc 0, value `-u` |
| `set-option -t S @k -x` | **rc 1, `psmux: set-option: unknown flag -x`** | rc 0, value `-x` |
| `set-option -t S -- @k -u` | **rc 1, `psmux: set-option: unknown flag --`**, value unchanged | rc 0, value `-u` |

tmux (getopt) stops flag parsing at the first non-option argument, so everything after the option name is a value; psmux keeps scanning.

## Root cause (v3.3.8 source), two layers

- **Client `src/main.rs:3307-3311`:** the validation loop folds EVERY `-`-leading token anywhere in argv into a `flags` string — position-blind; only `-t` consumes its value. `:3322-3326` then rejects unknown letters against an allowlist: that produces the `-x` rc 1, and `--` itself becomes the flag char `-` → `unknown flag --` rc 1, so no end-of-options state exists.
- **Server `src/server/connection.rs:2357-2368`:** `combined_has_set` scans all args for flag tokens, equally position-blind, so a `-u` in the value position (allowed through by the client) routes the whole command to the unset path — that is the silent rc-0 key deletion.

## Same pattern elsewhere (family)

The position-blind scan also backs `show-options` (`connection.rs:2448`, `combined_has`) and `set-hook` (`connection.rs:2866-2867`, `args.iter().any(...)`). #562 fixed exactly this for `send-keys` with an `end_of_opts` state — the same fix shape applies here.

## Impact

[...]

Issue #565, opened 2026-08-11
detach-client's client-spec -t is deleted by the generic stripper and misrouted as a session name (3.3.7/master)


dracic:
## Summary

`psmux detach-client -t <client>` cannot force-detach a client, and the first thing a caller sees is a
wrong error message. The generic argument pre-pass in `src/main.rs` strips `-t` **and its value** from
`cmd_args` after the subcommand, so the `-t` arm inside the `detach-client` handler is unreachable and
`t_target` is always `None`. Two observable consequences:

- **without `-s`** the client spec has already been consumed by the global routing pre-pass as a
  *session* name, so the command exits 1 with `psmux: no session '/dev/pts/3'` — an error that names the
  wrong kind of object and points the caller at a session that was never mentioned;
- **with `-s`** the handler's `effective_all` sees `t_target.is_none()` and rewrites the request as
  `detach-client -a`, so a request to detach one client is sent as "detach every client of this
  session", at rc 0.

The server side is complete and correct — `CtrlReq::ForceDetachClient` / `ForceDetachClientByTty` handle
both the `%N` and the tty form — but the binary never forwards `-t`, so those arms are reachable only
over the wire.

Everything under **Reproduction** was measured on the installed **3.3.7**
(`psmux 3.3.7 (05cc5d4 2026-07-20)`). The master half is a **source read only**: 3.3.8 could not be
built on this box. The relevant `src/main.rs` code is byte-identical between `v3.3.7` and master (see
**Mechanism**).

## Reproduction

Stock Windows box, PowerShell 7.

```powershell
# ── A. `-t <client>` without `-s`: the client spec is consumed as a SESSION name
psmux new-session -d -s up7du_c
psmux detach-client -t /dev/pts/3          ; "rc=$LASTEXITCODE"   # -> rc 1, "no session '/dev/pts/3'"
psmux has-session   -t up7du_c             ; "rc=$LASTEXITCODE"   # -> rc 0, session untouched

# ── B. `-t <client>` with `-s`: the request is sent as detach-all.
#    `destroy-unattached` is used here only as a DETECTOR: the detach-all arm
#    carries the destroy block, the by-tty arm does not (see Mechanism).
psmux -t up7du_c set-option -g destroy-unattached on
psmux -t up7du_c show-options -gv destroy-unattached                # -> on
psmux -t up7du_c display-message -p '#{session_attached}'           # -> 0  (no clients at all)
[...]

PR #587, opened 2026-08-22
fix(commands): preserve deferred command targets

MattKotsenas:
## Summary
- stop generic `-t` extraction at `bind-key`, `set-hook`, and `confirm-before` deferred-command boundaries
- route and remove only the process CLI's outer target, preserving nested targets and child `-t` arguments after `--`
- share the same boundary calculation across process, one-shot TCP, and control-mode ingress

## Why
psmux currently treats every `-t` token as its own target, even when it belongs to a stored command or spawned program. This can rewrite a binding or hook to target the wrong window, and `new-window -t dev -- tool -t worker` can route on or remove the child's `-t worker` instead of passing it through.

This complements the direct-exec and `--` handling added for #582/#583; it does not introduce another command parser.

## Testing
- focused Rust tests for wrapper and process target boundaries
- `cargo check --all-targets`
- isolated native-recorder smoke proving child `-t child` survives `new-window ... --`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -153,6 +153,42 @@ pub fn normalize_attached_target_flag(args: Vec<String>) -> Vec<String> {
     out
 }
 
+pub fn deferred_command_start<S: AsRef<str>>(command: &str, args: &[S]) -> Option<usize> {
+    let value_options: &[&str] = match command {
+        "bind-key" | "bind" => &["-T"],
+        "set-hook" => &["-t"],
+        "confirm-before" | "confirm" => &["-p", "-t"],
+        _ => return None,
+    };
+
+    let mut i = 0;
+    while i < args.len() {
+        let token = args[i].as_ref();
+        if token == "--" {
+            i += 1;
+            break;
+        }
+        if !token.starts_with('-') || token == "-" {
+            break;
+        }
+        i += if value_options.contains(&token) { 2 } else { 1 };
+    }
+
+    match command {
+        "bind-key" | "bind" | "set-hook" => (i < args.len()).then_some(i + 1),
+        "confirm-before" | "confirm" => (i < args.len()).then_some(i),
+        _ => None,
+    }
+}
+
+pub fn outer_target_scan_end<S: AsRef<str>>(command: &str, args: &[S]) -> usize {
+    deferred_command_start(command, args)
+        .into_iter()
+        .chain(args.iter().position(|arg| arg.as_ref() == "--"))
+        .min()
+        .unwrap_or(args.len())
+}
+
 /// Same as [`normalize_flag_equals`] but operates on `Vec<&str>`, returning
 /// owned strings (needed where the caller already has borrowed slices).
 pub fn normalize_flag_equals_borrowed(args: &[&str]) -> Vec<String> {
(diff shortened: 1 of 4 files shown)


psmux:
Thank you Matt. Merged, after rebasing it onto current master by hand (the `#583` positional discipline and the `#582` direct exec arm had moved underneath you; the logic composed cleanly, nothing semantic had to change).

**Reproduced on master before reading the diff**, with a detached session and the plain CLI:

* `psmux -t S bind-key -T prefix F5 select-window -t 2` was stored as `bind-key -T prefix F5 select-window`. The `-t 2` was eaten as the process target.
* `psmux bind-key -T prefix F6 select-window -t 2` with no outer target failed with `no server running on session '2'`.
* `psmux -t S set-hook -g after-new-window select-window -t 0` was stored as `after-new-window -> select-window`.
* `psmux new-window -t S -- cmd /k echo CHILDARGS -t worker` spawned a child whose real command line (read back from `Win32_Process`) was `cmd.EXE /k echo CHILDARGS`, the child's `-t worker` was gone, and the no outer target form routed the CLIENT on a session called `worker`.

**On your branch** (`tests/test_pr587_deferred_targets.ps1`, CLI, raw TCP one shot, control mode, and the edge cases): the bindings and hooks come back verbatim from `list-keys` and `show-hooks` (`bind-key -T prefix F5 select-window -t 2`, `after-new-window -> select-window -t 0`, root table `bind-key -T root F8 select-window -t 4`), the child's command line is `cmd.EXE /k echo CHILDARGS -t worker`, no phantom `worker` or `2` session ever appears, the outer target still routes and is stripped (`new-window -t S -n named`, TCP `rename-window -t :0`), `detach-client -t` keeps its `#565` exemption, and a `set-hook -t S after-rename-window select-window -t 5` routes on the first `-t` and keeps the second. The Rust suite came back 2938 of 2939 on the branch, and the single miss is ours rather than yours: `test_issue559_monitor_silence` asks for an `Instant` 3600s in the past, which a machine booted less than an hour ago cannot represent, so it panics. I fixed that helper on master rather than leave a test that fails on any freshly booted CI runner.

**tmux parity**: matches. tmux hands the whole tail after the key or hook name to the bound command untouched, `getopt` stops at `--`, and `-t` after the marker is the program's. Your boundaries for `bind-key`, `set-hook` and `confirm-before` reproduce that. For the record, tmux has a few more deferred command carriers (`if-shell`, `run-shell`, `command-prompt`, `display-menu`, `display-popup`) that still get the generic scan here when their command is written unquoted; the quoted forms are one token and already fine, so I am leaving that as a follow up rather than blocking on it.

One thing I noticed while poking at this, which I have now fixed on master: `new-window -- cmd ...` named the window `--` instead of `cmd`. It is a leftover from #582, not from your PR. 9bfdfc3 started keeping the `--` marker at the head of the command string so `build_command` knows to exec the argv directly; `build_command` decodes the marker, but the naming path did not, so `resolve_shell_program("-- cmd /k echo hi")` returned `--`. Before 9bfdfc3 the server passed only the first non flag token through, so the window was named `cmd`, which is also what tmux does. The naming path now strips the marker.

Thanks for the clean argv arithmetic and for the unit tests that pin the three boundaries, they made the rebase easy to check.

Issue #199, opened 2026-04-10
Clicking on pane in copy mode shows weird text selection outline


ivilkhovyi:
When I press anywhere on a pane while in copy mode, it shows some random text selection outline. So I just do a single left mouse click, and then scroll.

https://github.com/user-attachments/assets/d8b1c838-2174-4dfd-a48c-cd648d288a36

This is the latext psmux version, no ssh, just opened psmux in pwsh7.

Thank you.

psmux:
Hey @ivilkhovyi, thanks so much for reporting this! Great catch with the recording too, that made it super clear.

I tracked this one down and found the root cause. Here is what was happening:

**The problem:**
When you click (mouse down then up) in copy mode, even the tiniest mouse movement between pressing and releasing the button fires a "drag" event internally. That drag event was setting a selection anchor at the click position. Since the anchor and cursor ended up on the same cell, no selection was visible right away.

But then when you scrolled, the anchor's displayed row was recalculated using scroll offset math (`anchor_row + current_scroll - anchor_scroll`), which made it drift away from the cursor position. That created a phantom selection spanning from the stale anchor to wherever the cursor was, producing the weird highlight you saw.

**The fix:**
On mouse release, if the anchor and cursor are on the same cell (meaning no real drag happened), the anchor is now cleared. This prevents any phantom selection from appearing when you scroll afterward.

Committed in 8875bb5. Will be in the next release build. Let me know if you run into anything else!

Issue #62, opened 2026-03-01
Mouse selection behavior diverges from tmux parity


spooki44:
## Summary
- Wheel-up now correctly enters copy mode in `0.4.4`.
- Drag-select from normal mode still uses terminal-native selection and does not enter copy mode.
- While selecting in copy mode, moving cursor to pane top edge does not auto-scroll up/extend selection.
- Clicking another pane while in copy mode causes copy-mode behavior to become global again across panes.

## Environment
- OS: Windows 11
- Terminal: Windows Terminal
- Shell: PowerShell 7
- psmux: `0.4.5`
- tmux (parity reference): `3.6a` (no user config, mouse enabled)
- Config state for verification: reproduced with default config for psmux

## Note (keyboard selection status)
- The prior keyboard default-selection issue appears fixed in `0.4.3`:
  - copy-mode selection defaults to linear (stream), not rectangular.
- Rectangular selection capability appears present, but the tmux-style toggle entrypoint is missing.
- Remaining keyboard parity gap:
  - `v` does not toggle rectangular selection, but cancels selection.
  - `rectangle-toggle` is not exposed in `list-keys` or `list-commands`.

## Expected
- Mouse selection for tmux parity:
  - With `mouse off`, terminal-native selection behavior is expected (not pane-local, terminal selection highlight).
  - In tmux 3.6a parity testing with `mouse on`, observed behavior is:
    - mouse click-and-drag from normal mode enters copy mode and starts pane-local selection
    - while dragging, moving to the top edge of a pane scrolls up and extends selection upward
    - selection highlight is copy-mode highlight (not terminal-native highlight)
    - on mouse button release, selection is copied, copy mode exits, and pane returns to normal mode

## Actual
- Mouse selection:
  - In normal mode, mouse selection behavior appears identical to `mouse off`:
    - selection is not pane-local
    - highlight remains terminal-native selection highlight
    - drag/release does not appear to switch through copy mode semantics
  - In copy mode, mouse selection behavior appears mostly aligned with tmux semantics except:
    - moving mouse cursor to top of pane does not cause screen to scroll
    - switching panes using mouse click causes global copy mode behavior again (possible regression of #43).
    - after yank operation does not automatically return to normal mode.

## Reproduction
1. Start a fresh attached `psmux` 0.4.5 session (default config) and create split panes.
2. Generate enough lines in a pane to allow scrolling (for example hold `Enter`).
3. Set `mouse on`, confirm state:
[...]

PR #590, opened 2026-08-22
Drag selections auto-scroll through scrollback at pane edges

neich:
I have a problem when selecting command output in a PowerShell pane: when the output is taller than the window, I can't select past the last visible line. The selection stops dead at the pane edge, and moving the mouse outside the pane scrolls nothing. This PR makes drag selections auto-scroll through the scrollback at the pane's first and last row, the way tmux does, so output taller than the window can be selected in one gesture.

## Environment

- psmux master (this branch)
- The auto-scroll itself is server-side, in the copy-mode mouse handling. Getting a plain prompt drag there adds one control verb (`copy-drag-begin`) and one field in the layout JSON (`view_offset`); the client also re-sends the last drag every 50ms while the pointer sits at an edge, so the view keeps scrolling without mouse movement.
- Nothing platform-specific — the behavior is the same wherever psmux runs, and old client/server mixes keep working (details below).

## What happens

When a command prints more lines than the pane is tall, I can't select all of it. Starting from the bottom and dragging up, the selection stops at the pane's top row; dragging out of the pane does nothing. The cause: no selection path could cross an edge. The client's overlay selection only knows the text currently on screen, and server-side copy-mode drags were only accepted while the pointer was inside the pane's rectangle — dragging on the first or last row never scrolled. The tmux version of this (#62) only survives in `input::handle_mouse`, which nothing calls anymore.

There's a second version of it with `set -g scroll-enter-copy-mode off` (#193), which is how I run: the wheel scrolls the pane directly, without entering copy mode. I scroll back, start selecting, drag to the bottom row to continue toward the live output — and nothing happens, because the dump never told the client the pane was scrolled back. Even the top edge was subtly wrong from that state: entering copy mode reset its offset to 0 while the view stayed scrolled, so the copy would grab different lines than the ones I selected.

## Expected

1. **Dragging on or past the pane's first or last row keeps scrolling** so the selection keeps growing — one line on the row itself, faster the further past the edge the pointer is, like tmux. Holding the pointer there is enough: the client re-sends the last drag every 50ms, so the scroll is continuous.
2. **A plain prompt drag continues into scrollback.** When it reaches the top row, the client hands the selection off to server-side copy mode with the new `copy-drag-begin` verb (the highlight switches to the copy-mode colors) and it keeps scrolling from there. With `scroll-enter-copy-mode off`, the same happens at the bottom row of a scrolled-back view, continuing toward the live output — and the view I was looking at is preserved, so the selection anchors on the lines actually on screen. That also fixes keyboard entry (`prefix-[`) over a scrolled view.
3. **Releasing copies the selection and returns to the live view**, like tmux's `copy-pipe-and-cancel` — even when the selection came up empty, so a drag can never strand me in copy mode. The release compares content positions rather than screen cells, so a selection whose edge scrolling parked the release back on the anchor's screen cell still copies everything it swept. Plain clicks and micro-jitter keep their #199 protection and never yank or cancel anything.

The scrolled-back signal is one new field on every layout leaf, filled by both dump paths:

```json
{"type":"leaf","id":7,...,"copy_mode":false,"scroll_offset":0,"view_offset":9,...}
```

It's `#[serde(default)]`: an old server that doesn't send it just means the bottom-edge handoff never triggers (the top edge works regardless), and an old client ignores it coming from a new server.

## Repro

1. Default config. In a pane, print something long: `1..100 | % { "line-$_" }`.
2. Drag-select from a lower row up past the pane's top row. Before this branch: the selection stops at the edge. After: it converts to the copy-mode selection, keeps scrolling for as long as the pointer stays there, and the release copies everything swept.
3. `set -g scroll-enter-copy-mode off`, wheel a few pages back, start a selection and drag to the bottom row. Before: nothing happens. After: it converts and scrolls back toward the live output, and the copied text matches the lines I selected.

The new `window_ops` tests cover the edge scrolling and its acceleration, the dwell repeat, the handoff keeping a direct-scrolled offset, the content-position release, and the empty-drag cancel; `tests/test_copy_drag_autoscroll.ps1` drives both flows end-to-end over the control protocol and asserts mode, scroll position, and buffer content (the `pane-mouse` / `pane-scroll` / `copy-drag-begin` CLI forwarding stays in as a testing aid). Seven existing test files gained a mechanical `view_offset` line.


psmux:
Thank you neich. This one is merged, and it is a genuinely nice piece of work. The write up was precise enough that I could reproduce the exact failure before reading a line of the diff, which is how I like to review.

**What I reproduced on master first** (attached client, real `MOUSE_EVENT` press, move, dwell and release injected into the console input buffer, plus the raw control verbs):

* Dragging a selection from row 25 up to row 0 and parking the pointer there for 1.5s: `pane_in_mode/scroll_position` sampled `0/0, 0/0` the whole time. Nothing scrolled, and the release yanked nothing past the screen.
* Inside copy mode, six `pane-mouse 32` drags on row 0 left `scroll_position` at 0, and the release left the pane in copy mode.
* `scroll-enter-copy-mode off`, wheel back 12 notches (view showed lines 137..165), drag to the bottom row: the yank came back with line 174, the live line, exactly the anchor drift you described. And the layout dump had no `view_offset` at all.

**On your branch** (same harness, `tests/test_pr590_drag_autoscroll.ps1`, 33 checks, all green):

* Dwell on row 0 with no further mouse events: `1/6` at 0.7s, `1/14` at 1.2s, so the 50ms client resend is doing its job. A 3s dwell reached 48 lines above the original screen versus 24 for 1.5s, so it is continuous, not a single hop.
* Release: `pane_in_mode=0`, `scroll_position=0`, and the buffer held 49 rows starting at a line that was 24 rows above the visible top.
* Server side: row 0 drags scroll one line each, drags at row minus 3 scroll two lines each (your acceleration), release cancels and yanks 7 rows.
* `view_offset=36` shows up in the dump after the direct wheel scroll, the bottom edge handoff anchored on line 140 with the view top at 137 and swept down to 189, and `prefix-[` over a wheel scrolled view kept `scroll_position=30` with no visible jump.
* Plain click and same cell jitter: mode unchanged, buffer untouched. A short in pane drag stays on the client overlay and lands in the clipboard as before.
* Your own `tests/test_copy_drag_autoscroll.ps1`: 12/12. The Rust suite in the branch worktree came back 2933 of 2936, and neither miss belongs to you: two `test_config_plugin_paths` cases collided over a shared temp directory because I had two `cargo test` jobs running at once (both green on their own rerun), and `test_issue559_monitor_silence` turned out to be a test of ours that asks for an `Instant` 3600s in the past, which a machine booted less than an hour ago cannot represent, so it panicked. I fixed that helper on master rather than leave a test that fails on any freshly booted CI runner.

**tmux parity**, checked against `window-copy.c`: the 50ms repeat, one line per tick on the first and last row, and `copy-pipe-and-cancel` on release all match. Two places where you went further than tmux, and I am fine with both, I just want them on the record rather than described as parity: tmux stops scrolling the moment the pointer leaves the pane rectangle (`cmd_mouse_at` returns nonzero and the timer is already deleted), and it never accelerates. On a console host that only ever reports pointer positions inside the window, keeping the scroll alive past the edge is the friendlier choice, so it stays. I trimmed the "(tmux parity)" wording in `docs/features.md` to say that.

Your harness also flushed out a real bug of ours that has nothing to do with this PR, and I have fixed it on master: after copy mode was cancelled from outside the client with `send-keys -X cancel`, the next plain click silently re entered copy mode. Root cause: `exit_copy_mode` clears the pane local `copy_state` (that line was added in 48b83a5 precisely so a stale copy mode is not restored), but the server's `send-keys -X` arms, written two days later in 4964aaf, hand rolled the exit from the pre fix body and never cleared it. `switch_with_copy_save` restores any pane whose `copy_state` is set, and the client sends `select-pane` before every mouse click, so one click brought it back. Four arms shared the defect (`cancel`, `copy-selection-and-cancel`, `copy-pipe-and-cancel`, `copy-end-of-line`); all four now call `exit_copy_mode`. Worth knowing for this PR: `copy-pipe-and-cancel` is tmux's default `MouseDragEnd1Pane` binding, so anyone driving your drag through that binding rather than your release path would have hit it.

Credit to you for the content position release compare and for keeping `view_offset` a `#[serde(default)]` field so mixed client and server builds stay compatible. Thanks again, this makes selecting long PowerShell output feel the way it should.

PR #591, opened 2026-08-23
ci: enforce Clippy for portable-pty

MattKotsenas:
## Summary

- make all `portable-pty-psmux` Cargo targets Clippy-clean in the Windows CI job
- add a dedicated CI job that treats future warnings in that crate as errors
- replace the uninitialized attribute-list buffer and apply behavior-neutral lint cleanups

## Why

The repository installs Clippy and defines `cargo lint`, but CI does not run either.

A workspace-wide `-D warnings` gate is not reviewable yet: before this cleanup, a comprehensive workspace Clippy scan found 792 unique warning sites. This PR establishes a strict ratchet at the smaller `portable-pty-psmux` boundary instead:

```powershell
cargo clippy -p portable-pty-psmux --all-targets --no-deps -- -D warnings
```

## Attribute-list storage

`Vec::set_len` requires newly exposed elements to already be initialized. The old implementation reserved capacity, set the vector length, and created a `&mut [u8]` over bytes that Win32 had not initialized yet. The replacement allocates `vec![0; bytes_required]` before passing the same pointer and size to `InitializeProcThreadAttributeList`.

The uninitialized pattern originated in WezTerm's 2019 ConPTY spawner (`b387464`) and appears intended to avoid zero-filling before Win32 initializes the structure; the introducing commit records no measurement. A microbenchmark compared zeroed storage with valid uninitialized `Vec<MaybeUninit<u8>>` storage around the same Win32 initialize/delete calls. For the 48-byte buffer, five repeated runs measured a 3.45-4.19 ns cost per construction, about 0.000004 ms per ConPTY child spawn.

## Intentional lint decisions

- `HPCON` keeps the Windows SDK spelling under an item-scoped `upper_case_acronyms` allow.
- The two scoped `const_is_empty` allows preserve illustrative stdin-write branches: `whoami` does not consume stdin, and injecting input into `narrow` would alter an arbitrary command.
- Registry enumeration continues to ignore individual unreadable values; `.flatten()` expresses the existing nested `if let Ok(...)` behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,21 @@ jobs:
       - name: Build
         run: cargo build --release --target ${{ matrix.target }}
 
+  clippy-portable-pty:
+    name: Clippy portable-pty
+    runs-on: windows-latest
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+
+      - name: Run Clippy
+        run: cargo clippy -p portable-pty-psmux --all-targets --no-deps -- -D warnings
+
   test:
     name: Test Windows x64
     runs-on: windows-latest
--- a/crates/portable-pty-psmux/examples/narrow.rs
+++ b/crates/portable-pty-psmux/examples/narrow.rs
@@ -66,6 +66,7 @@ fn main() {
         // want to send data to the child, you'd set `to_write` to
         // that data and do it like this:
         let to_write = "";
+        #[allow(clippy::const_is_empty)]
         if !to_write.is_empty() {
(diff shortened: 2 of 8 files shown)


psmux:
Thank you Matt. Merged.

I added two pins next to your change: `tests-rs/test_pr591_procthreadattr.rs` checks that the attribute list storage is the zeroed buffer we own (Win32 writes into it, `as_mut_ptr` aliases the Vec, larger attribute counts grow it), and `tests/test_pr591_conpty_spawn.ps1` runs every ConPTY spawn path we have (new-session, split -h and -v, new-window, respawn-pane -k, the `--` direct exec form, and a 200x50 session) and checks each produces a live pane. All green, and the Clippy gate now runs for `portable-pty-psmux` in CI, which is where I want it.

Thanks for the cmdbuilder cleanup too, that file reads much better now.

Issue #593, opened 2026-08-24
Status bar clicks: only the first status row is hit-tested, and tab hit-boxes are rebuilt from a hardcoded layout


ChrisonSimtian:
Two defects in status-bar click handling. Both show up as "clicking a window tab does nothing, or selects the wrong window" once the bar is customised. Read against v3.3.8; both still present on master.

## 1. Only the first status row is hit-tested

`src/input.rs:2468`:

```rust
let status_row = window_area.y + window_area.height; // status bar is 1 row below window area
if matches!(me.kind, MouseEventKind::Down(MouseButton::Left)) && me.row == status_row {
```

`status_row` is a single row, so with `set -g status 2` only the upper of the two rows is hit-tested. A window list placed in `status-format[1]` renders correctly but is completely dead to the mouse — the click falls through to pane hit-testing and matches nothing.

The multi-line status bar is documented in `docs/configuration.md` with the window list on `status-format[1]`, which is exactly the layout that cannot be clicked.

## 2. Tab hit-boxes are rebuilt from a hardcoded layout, not the rendered bar

`src/window_ops.rs:654` is the only writer of `app.tab_positions`, and `src/input.rs:2470` is the only reader:

```rust
pub fn update_tab_positions(app: &mut AppState) {
    let session_label_len = app.session_name.len() as u16 + 3; // '[' + name + ']' + ' '
    cursor_x += session_label_len;
    for (i, w) in app.windows.iter().enumerate() {
        let label = format!("{}: {} ", display_idx, w.name);
        ...
    }
}
```

It assumes the bar is literally `[session] 1: name 2: name `. It never consults `status-left`, `window-status-format`, `window-status-current-format`, `status-format[N]`, `window-status-separator`, or `status-justify`. Any customised bar puts the visible tabs somewhere other than these hit-boxes, so clicks select the wrong window or nothing.

This is #168, which is closed — but the hardcoding is still there at v3.3.8, and `tests/test_issue168_tab_positions_status_left.ps1` only covers `status-left`.

## Impact

A themed bar — the tokyonight/catppuccin plugin themes included, since they all set `window-status-format` — silently loses accurate tab clicking. There is no error; clicks just land on the wrong tab.

## Workaround

Compose the tab row to match `update_tab_positions` character for character, and keep it on row 0:

```
set -g status 2
[...]

PR #594, opened 2026-08-24
docs(status-format): the multi-line example produces an unclickable window list

ChrisonSimtian:
Refs #593.

The multi-line status bar example in `docs/configuration.md` gives a window list that cannot be clicked. Following it verbatim is how I hit #593.

```tmux
set -g status-format[0] "#[align=left]#S #[align=right]%H:%M"
set -g status-format[1] "#[align=left]#{W:#I:#W }"
```

Two independent reasons it cannot work:

- No `#[range=window|N]` markers. With a custom `status-format[0]`, hit-boxes come from that line's parsed ranges and nothing else (`client.rs:6000`), so the list is empty and no tab is a click target.
- The list is on line 1. Ranges are only collected inside the `use_status_format_0` branch, and only the first status row is hit-tested (`client_status_row = status_chunk.y`, `client.rs:5964`/`4173`).

It renders perfectly, which is what makes it confusing — there is no warning, the tabs just do nothing.

## What this changes

Docs only. The example moves the window list to line 0 and carries the markers, so it works as written. Plus a short note on the two traps in hand-writing them:

- `N` is zero-based — `base_index` is added back on resolve (`client.rs:6003`), so with the default `base-index 1` the marker needs `#{window_index}` − 1.
- Commas split `#{W:inactive,current}`, so a comma inside the loop must be escaped `#,` — including the one in `#{e|-:a,b}`. Styles are easier written `#[fg=green]#[bold]` than `#[fg=green,bold]`.

The final paragraph documents the line-0 restriction as current behaviour. If #593 is fixed so ranges are collected from every status line and all status rows are hit-tested, that paragraph should go.

Verified against 3.3.7 and the 3.3.8 release binary: the example's window list is unclickable, and the config in this diff is clickable.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -228,12 +228,20 @@ psmux supports a multi-line status bar using the `status-format[]` array. Set th
 # Enable a 2-line status bar
 set -g status 2
 
-# Configure each line (0-indexed)
-set -g status-format[0] "#[align=left]#S #[align=right]%H:%M"
-set -g status-format[1] "#[align=left]#{W:#I:#W }"
+# Line 0 carries the window list. #[range=window|N] is what makes each tab a
+# click target -- see "Clickable window tabs" below.
+set -g status-format[0] "#[align=left]#{W:#[range=window|#{e|-:#{window_index}#,1}]  #I #W  #[norange]}"
+set -g status-format[1] "#[align=left]#S #[align=right]%H:%M"
 ```
 
 The first line (`status-format[0]`) replaces the default status bar content. Additional lines stack below (or above, depending on `status-position`).
+
+**Clickable window tabs.** With a custom `status-format[0]`, tab click targets come from that line's `#[range=window|N]` markers and nothing else. A window list written without them renders correctly but cannot be clicked, with no warning. Two things to get right:
+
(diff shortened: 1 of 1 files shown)


psmux:
Hi @ChrisonSimtian, thank you for taking the time to write this up, and the analysis in it was spot on for 3.3.8. The underlying behavior just changed though, so I want to spare you a rebase against a moving target.

7985304 on master fixes #593 itself rather than documenting around it:

* ranges are now harvested from every `status-format` line, and every status row is hit tested, so a window list on `status-format[1]` is clickable and the old advice to keep it on line 0 no longer applies
* `#[range=window|N]` now takes the real window index with no `base-index` arithmetic, matching tmux's `winlink_find_by_index`, so the `#{e|-:#{window_index}#,1}` subtraction idiom this PR teaches would now aim one window too low

That same commit also rewrote the `docs/configuration.md` section this PR touches: the multi line example now carries `#[range=window|#{window_index}]` markers (verified clickable with real mouse injection before committing), and a new paragraph documents the marker contract, the `#,` comma escaping inside `#{W:...}`, and the zero arithmetic index semantics.

Since the fix landed with its own docs update covering this ground, I am closing this one as superseded. The parts of your text that survive, the comma escaping warning and the markers or nothing contract, are both in the new section, and your write up in #593 is what got the real fix shipped quickly. Genuinely appreciated, and please give master a try with your themed config.

Issue #570, opened 2026-08-14
Mouse scroll doesn't reach Neovim pane except at ~50% width split


MKoushikYadav:
## Mouse scroll doesn't reach Neovim pane except at ~50% width split

**Setup:** psmux, PowerShell 7, single window split into 2 side-by-side panes.
Neovim (full-screen within its pane) with a terminal buffer open.

**What I'm seeing:** Mouse wheel scroll doesn't reach the Neovim window
unless the pane happens to be close to exactly 50% of total terminal
width. At other ratios (tested ~30% and ~60%),( ~49%,51%), scroll up does nothing;
at 50%, scroll works normally.

**Repro:**
1. Open up neovim inside psmux. (Specifically lazyvim)
3. Split a window into two panes.
4. Resize the Neovim pane to ~30% width. Scroll up — no response.
5. Resize to ~50% width. Scroll up/down — works.
6. Resize to ~60% width. Scroll up — no response again.**What I ruled out on my end:**
- Stale Neovim size cache — `&columns`/`&lines` match `psmux display -p
  "#{pane_width}x#{pane_height}"` in both broken and working states.
- Custom keymaps — `:verbose map <ScrollWheelUp/Down>` shows no user mapping.
- `mouse=a` is set in Neovim throughout.
- Doesn't look app-specific — reproduces regardless of scroll history depth.

I don't have visibility into psmux's own mouse-coordinate handling, so I
can't confirm the actual cause — just wanted to report the pattern in case
it's useful.

**Environment:**
- psmux: 3.3.7
- Windows Terminal, PowerShell 7.6.4
- Neovim: 0.12.4

psmux:
Thank you for this one. Your instinct was right that something was off in psmux's mouse coordinate handling, and the pattern you described was the clue that cracked it.

**It reproduces, and it is a real bug.** Fixed in 7435244.

### What was happening

psmux was telling your application that every wheel notch happened at the **centre of the pane**, no matter where your pointer actually was.

The client worked out which pane the wheel was over, sent `pane-scroll PANE up|down` to the server, and threw the pointer position away. With nothing to report, the server synthesised the pane centre.

Clicks were never affected, because the click command already carried a column and a row. That is exactly why this looked like "scroll does not reach Neovim" rather than "the mouse is dead".

### The measurement

I ran a child process inside the pane that requests SGR mouse reporting and logs the raw bytes psmux forwards, then turned the wheel at the left edge, the middle and the right edge of the pane. Before the fix:

[...]

Issue #296, opened 2026-05-15
psmux -> claude code, ctrl-G to open nvim, then nvim hangs


zeroxia:
In Windows Terminal,

launch psmux

launch claude code

Press Ctrl-G to enter prompt inside nvim

nvim opens, but not responding, typing some keys, the entire claude code session exists.

psmux:
Fixed in 4d4d0ae.

**Root cause:** pane_wants_mouse() uses a 3-tier heuristic (explicit mouse protocol mode → alt-screen detection → fullscreen heuristic) to decide whether to forward mouse events. For hover/motion events (MouseEventKind::Moved), this was too permissive: when Claude Code (Bubble Tea/Go) spawns nvim via Ctrl+G, nvim enters alt-screen but does **not** enable mouse tracking (no set mouse=a / no DECSET 1000/1002/1003). psmux saw alt-screen=true and flooded nvim's PTY pipe with SGR motion sequences (ESC[<35;col;rowM) on every mouse move. Nvim interpreted these as keyboard input, causing it to appear hung.

**Fix:** Added pane_wants_hover() — a strict check that only returns true when the child process has explicitly enabled motion tracking via DECSET 1002 (ButtonMotion) or DECSET 1003 (AnyMotion). Used for hover and drag-motion forwarding. Click/scroll events still use the broader pane_wants_mouse() heuristic unchanged.

E2E test confirms nvim works correctly when spawned from a TUI wrapper (simulating the Claude Code → nvim path).

Issue #168, opened 2026-04-01
update_tab_positions() uses hardcoded format, causing tab click misalignment when status-left / window-status-format is customized


noctcross:
## Environment
- psmux version: built from latest commit (post-#164 fix)
- OS: Windows 11

## Description

`update_tab_positions()` in `src/window_ops.rs` (line 429-443) calculates tab click positions using hardcoded formats, ignoring `status-left`, `window-status-format`, `window-status-current-format`, and `window-status-separator` settings.

The client-side rendering (client.rs) does respect these settings, so when any of them are customized, the click regions no longer align with the visually rendered tabs.

## Steps to Reproduce

1. Customize status-left:
```
psmux set -g status-left ""
psmux set -g status-left-length 0
```
2. Click on a window tab in the status bar
3. The click target is offset by the width of `[session_name] ` (which is no longer displayed)

## Root Cause

In `src/window_ops.rs:429-443`:

```rust
pub fn update_tab_positions(app: &mut AppState) {
    let mut tab_pos: Vec<(usize, u16, u16)> = Vec::new();
    let mut cursor_x: u16 = 0;
    let session_label_len = app.session_name.len() as u16 + 3; // hardcoded "[name] "
    cursor_x += session_label_len;
    for (i, w) in app.windows.iter().enumerate() {
        let display_idx = i + app.window_base_index;
        let label = format!("{}: {} ", display_idx, w.name); // hardcoded format
        let start_x = cursor_x;
        cursor_x += label.len() as u16;
        tab_pos.push((i, start_x, cursor_x));
    }
    app.tab_positions = tab_pos;
}
```

Two hardcoded values:
1. **Line 433**: `session_name.len() + 3` — assumes `[session_name] ` prefix regardless of `status-left` setting
2. **Line 438**: `format!("{}: {} ", ...)` — ignores `window-status-format`, `window-status-current-format`, and `window-status-separator`

## Expected Behavior

`update_tab_positions()` should calculate click positions based on the actual rendered content, using the same `status-left` and `window-status-format` settings that the client uses for rendering.

ToxMox:
I actually have a PR https://github.com/psmux/psmux/pull/163 in the queue that fixes mouse alignment issues. Hopefully the fixes get merged in 😄

psmux:
Hey @noctcross, great catch on this one! Your analysis of the root cause was spot on.

[...]

PR #595, opened 2026-08-24
refactor(mouse): remove unreachable pre-server mouse path

MattKotsenas:
## Summary

- remove the unreachable pre-server mouse dispatcher and the mouse/clipboard helpers used only by it
- remove unused server tab-position state; the live client owns rendered tab hit ranges
- remove mouse-routing heuristics used only by that dispatcher and retire their direct tests
- preserve process-classification coverage and move hover coverage to the live attached-client route
- preserve the live client -> TCP -> `window_ops` mouse route

`input::handle_key` remains for direct Rust-test coverage; migrating that coverage to the live client/server handlers is outside this PR.

## Validation

- `cargo test --bin psmux mouse` - 73 passed
- `cargo test --bin psmux foreground_is_shell_classifies_live_processes` - 1 passed
- `cargo check --all-targets`
- `tests/test_issue295_scroll_regression.ps1` - 14 passed
- `tests/test_mouse_hover.ps1` - 7 passed, including exact attached-client SGR bytes and verified teardown


psmux:
Thank you Matt. Merged.

A refactor that deletes 589 lines of `input.rs` and claims zero behavioural change needed a live proof rather than a diff read, so I wrote `tests/test_pr595_mouse_route_parity.ps1`: 37 checks on real attached clients with injected `MOUSE_EVENT` wheel, click and motion records, covering every behaviour the trimmed or deleted tests used to protect, on the live route only (attached client, TCP, server, or the raw verbs with no client at all):

* `#381` and `#360`: a filled plain pwsh shell and a filled Git Bash shell, wheel enters copy mode, no raw SGR text appears, and Ctrl+C still reaches a running `ping`.
* `#285`, `#570`, `#296`: nvim with `mouse=a` on the alternate screen receives the wheel and the clicks (logged by nvim mappings), never copy mode, and bare motion over an app that never enabled motion tracking leaks nothing.
* `#168` and `#593`: status bar tab clicks select the right window with the default bar, a long `status-left` and a two row bar.
* the `copy-mode-vi` key table fires on the live `send-keys` route, and `pane-scroll` and `pane-mouse` behave the same over raw TCP.

36 of the 37 checks are green on your branch and identical on master before it, so the removed path was indeed unreachable. The one that is not green is my own harness rather than your change: the pwsh arm of the `#381` Ctrl+C check, where pwsh under ConPTY does not reliably pass an injected `^c` to its own `ping` child. It behaves the same on master, on your branch and on the current merge, and the Git Bash arm right below it exercises the same code path and passes, so I turned that arm into a skip with the reason recorded.

Good riddance to the pre server tab position state, it has confused more than one investigation of mine. Thanks for the careful sweep of the tests that referenced it.

1137MOM:
- @psmux **__**@psmux @claude

Issue #613, opened 2026-08-28
#598 gate: a node child's raw mode strips ENABLE_MOUSE_INPUT and the pane's wheel goes silent for good


457shop:
## Summary

The `#598` wheel gate authorizes forwarding on `ENABLE_MOUSE_INPUT` being set on the child console *right now*. That bit is a property of the **console**, not of the application, and any `node` child that calls `setRawMode` overwrites the whole console mode with `0x0208` and never restores it. From that moment the pane is permanently silent to the wheel — not degraded, not falling back to copy mode, nothing — and restarting the application in the pane does not recover it, because the console outlives the process. Opening a new window does.

On this machine every long-lived Claude Code pane ends up in that state after a session that spawns node children, and they break one at a time, which is what makes it look like a per-window curse.

## Environment

- psmux 3.3.8, built from `master` on 2026-08-27 (includes the `#598` gate from `1357bc1`)
- Windows 11 Enterprise 10.0.26200
- Outer terminal: ShellFish (iOS) over SSH; single client, `120x29`, six windows in one session
- Pane app: Claude Code CLI 2.1.247 / 2.1.248 / 2.1.250 (all three behave identically)
- `mouse on`, `scroll-enter-copy-mode on`, `history-limit 2000`

## Measurement

Console input mode of each pane's child, read with `AttachConsole` + `CreateFileW("CONIN$")` + `GetConsoleMode`, against the user's own report of which windows scrolled. Five panes, no exceptions:

| pane | window | console mode | `ENABLE_MOUSE_INPUT` | wheel scrolls? |
| --- | --- | --- | --- | --- |
| `%4` | win-A | `0x0208` | off | no |
| `%5` | win-B | `0x0298` | **on** | **yes** |
| `%7` | win-C | `0x0208` | off | no |
| `%8` | win-D | `0x0208` | off | no |
| `%9` | win-E | `0x0298` | **on** | **yes** |

`0x0298` = `0x0208 | ENABLE_MOUSE_INPUT | ENABLE_EXTENDED_FLAGS`.

Setting the two missing bits on a dead pane makes the wheel work **immediately**, with nothing else changed — no restart, no reattach, no new window. Verified on three panes.

Ruled out first, so they do not come up again: alternate screen (`alternate_on=1` on the scrolling panes too), Claude Code version (same pane went `2.1.248` → `2.1.250`, symptom unchanged), tmux options (one server, one client, all six windows share them), and `mouse_proto_owner` (see below).

## Mechanism

[...]

Issue #598, opened 2026-08-25
Using mouse scroll with psmux and codex cli messes up codex cli completely


larroy:
When I have codex cli running in a psmux pannel and scroll back using the mouse or touchpad, it completely messes up codex. I'm using iterm2 on mac.

When not using codex it shows [copy mode] on the top right and scrolls back the terminal. But inside codex it goes back to previous commands and the window history dissapears. What could be the cause for this?

When running htop for example, scrolling back also does strange things like "Search: " and a bunch of garbage. Terminal is messed up and I can't even kill the session.

psmux:
Hi @larroy , Thanks for sharing. I will definitely find the root cause, solve it and help you get through it.

Did you try using cargo install --git https://github.com/psmux/psmux psmux ? 

What version of Psmux are you using?

psmux:
Hi @larroy, I reproduced this, found the root cause, and it is fixed on master in [`1357bc1`](https://github.com/psmux/psmux/commit/1357bc1).

Thank you for the report, and especially for mentioning htop as well as codex. Those two details turned out to point at two different problems, and I would have missed one of them otherwise.

## What was actually happening

psmux decided "this pane wants mouse reports" from the **alternate screen alone**. Any full screen program therefore got the raw SGR mouse report written straight into its stdin on every wheel notch, whether or not it had ever asked for the mouse.

I caught it at the byte level with a model full screen app that logs every single byte it receives. One wheel notch over a pane whose application had enabled **no** mouse protocol delivered this:

```
RECV <ESC>[<64;31;15M  |  1B 5B 3C 36 34 3B 33 31 3B 31 35 4D
```

A program that does not parse mouse reports reads that as somebody typing. That is where the garbage in htop comes from, and why the terminal ends up in a state you cannot get out of.

## What tmux does, and what psmux does now

tmux only ever writes a mouse report to a pane whose application has enabled a mouse mode. From tmux next-3.8, `input-keys.c`:

```c
/* input_key_mouse() */
if (m->ignore || (s->mode & ALL_MOUSE_MODES) == 0)
        return;
```

and its default wheel binding in `key-bindings.c`:

```
bind -n WheelUpPane { if -F '#{||:#{alternate_on},#{pane_in_mode},#{mouse_any_flag}}' { send -M } { copy-mode -e } }
```

So `alternate_on` is only a reason **not** to fall through to copy mode. It never makes tmux write bytes. On the alternate screen with no mouse mode, `send -M` is a silent no op.

[...]

Issue #573, opened 2026-08-15
mouse support stops working upgrading to 3.3.7, 3.3.8


laciferin2024:
set -g mouse on , is no longer work now

psmux:
Thanks for flagging this. I went looking for it properly rather than guessing, and I could not make `set -g mouse on` fail on 3.3.7. Here is exactly what I checked, so you can tell me which part looks different on your machine.

### What I tested on 3.3.7

Every route you could turn the mouse on, checking both that the option reads back and that the wheel actually does something:

| route | result |
|---|---|
| fresh session default | `on` |
| `set-option -g mouse on` at runtime | `on` |
| `set -g mouse on` (alias) | `on` |
| `set -g mouse off` then back `on` | toggles correctly |
| `set -g mouse on` in the config file at startup | `on` |
| `set-option -g mouse on` in config | `on` |
| `set -g mouse 1` in config | `on` |
| `setw -g mouse on` in config | `on` |
| `set -gq mouse on` in config | `on` |

And the behaviour behind the option:

* wheel with mouse **on**: pane enters copy mode, so the wheel is being acted on
* wheel with mouse **off**: nothing happens, so the gate works in both directions
* click to focus a pane: focus moved to the clicked pane correctly
* wheel over a full screen TUI application: delivered
* wheel over an application that enables mouse the Win32 console way, with no DECSET at all: delivered

So the option plumbing itself is healthy on 3.3.7.

### What I think may actually have bitten you

3.3.7 did change mouse behaviour in two places. One of them is a genuine bug that I found and fixed today while working #570: psmux was reporting every wheel notch at the **centre of the pane** instead of at your pointer. Fixed in 7435244.

If your editor or terminal application has more than one window inside a single psmux pane, a file tree beside an editor for example, that bug meant the wheel scrolled whichever inner window happened to sit at the pane centre rather than the one you were pointing at. That very easily reads as "the mouse stopped working". Details are in #570.

3.3.7 also tightened when clicks and bare pointer motion get forwarded into a pane, so that container shells stop receiving stray escape sequences (#349). That gating keys off the application announcing a mouse mode. If your case is a click that no longer reaches something, this is the most likely suspect and I want to see it.

### What would help me

Could you tell me:

1. Which terminal you run psmux in (Windows Terminal, Alacritty, WezTerm, conhost, over SSH)
2. What "mouse" means in your workflow: wheel scrolling, clicking to switch pane, dragging to select text, or the mouse inside a specific application
[...]

PR #614, opened 2026-08-28
feat(#613): PSMUX_FORCE_WHEEL, an opt-in way past a gate a child can destroy

457shop:
Adds `PSMUX_FORCE_WHEEL`, an opt-in escape hatch for panes that can never satisfy the `#598` wheel gate. Off by default; the gate itself is unchanged.

Full measurement is in #613. The short version:

## Why the gate's signals do not hold

The gate accepts two answers to "did this application ask for the mouse": a DECSET attributed to a confirmed foreground (`mouse_proto_owner`), or `ENABLE_MOUSE_INPUT` on the child console right now.

The second one is not psmux's to keep. Console input mode belongs to the **console**, not to the process that set it, and a child entering raw mode overwrites the whole mode word instead of clearing single bits — libuv writes `ENABLE_WINDOW_INPUT | ENABLE_VIRTUAL_TERMINAL_INPUT` and does not restore the previous mode on exit. One `node` child anywhere in the pane's process tree therefore strips the gate's authorization permanently. Because the console outlives the process, restarting the pane's application does not earn it back; only a new window does.

Measured on five long-lived panes: every pane whose console mode read `0x0208` was silent to the wheel, every pane that read `0x0298` scrolled, no exceptions. Setting the two missing bits by hand made a dead pane scroll immediately with nothing else changed, on three of them. A scratch pane goes `0x01E4` -> `0x0208` the moment a node child enables raw mode and stays there after it exits.

The first answer is not always available either. Claude Code consumes SGR wheel reports fine — flipping the console bit makes it scroll instantly — but emits no DECSET, so `mouse_proto_owner` has nothing to attribute and `#{mouse_any_flag}` reads `no` on every version measured. For that pane the authorization is unearnable, not merely lost.

## What this does not change

`#598` is not relaxed and not reverted. It is what stops htop reading a raw report as keystrokes and filling its search prompt with the digits, and codex losing its transcript. `PSMUX_FORCE_WHEEL` is off by default, and because it is server wide, turning it on re-exposes that damage for panes that genuinely do not read the mouse. That is why it is an explicit opt-in rather than a widened heuristic — if you would rather scope it per pane or per window, I am happy to rework it.

## Why not `PSMUX_FORCE_MOUSE`

Same reason `conpty_needs_mouse_record_bypass` is deliberately not routed through it: #573's override governs whether psmux may write mouse DECSET **out** to the terminal, this one governs whether a report psmux already holds may be delivered **into** a pane. Two tests pin that independence in both directions, including that opting into the wheel cannot relax #457's build gate — whose whole purpose is keeping an SGR report away from a conhost that fast-fails on it and takes the pane process down.

## Diff

Six lines of logic; the rest is the doc comment and the tests.

- `src/ssh_input.rs` — `FORCE_WHEEL_ENV` + `wheel_gate_forced()`, placed beside the other pane-direction mouse policy
- `src/window_ops.rs` — one condition in the gate
- `docs/configuration.md` — one row in the Windows escape-hatch table
- `tests-rs/test_issue613_wheel_gate_override.rs` — 5 tests

## Verification

- `cargo build --release` clean, 0 warnings
- 5 new tests pass; `cargo test --bins mouse` is 100/100
- Mutation-checked rather than trusted green: reading the wrong env var kills the affirmative test and the independence test; treating any non-empty value as consent kills the "everything else keeps the gate" test

One gap to be transparent about: the wiring in `window_ops.rs` is a single condition verified by inspection, not by an end-to-end run. I did not execute the PowerShell suites, since they spin up a server on the machine this was reported from. If you would like `tests/test_issue598_wheel_mouse_gate.ps1` extended with a forced-override case, say the word and I will add it.

Refs #613. Not marked as closing it: this gives affected users a way out, but it does not make the gate's authorization signal durable, so the underlying report stands. Close it along with this if you consider the escape hatch a sufficient answer.


psmux:
Thanks @457shop, the reasoning in this PR is right and I have taken it, but as part of a larger change rather than on its own, so I am closing this in favour of c9ff113 on master.

What landed, in short (full measurements in my reply on #613):

* The reported mechanism reproduced exactly: 0 of 8 notches after one `node` raw mode child, and conhost republishes the strip upstream as `ESC[?1003;1006l`, so the same event that clears `ENABLE_MOUSE_INPUT` also cleared `mouse_proto_owner`. Both gate signals were the same console word.
* The durable fix is a pane owned latch (`Pane::wheel_auth`) modelled on tmux's per screen `ALL_MOUSE_MODES`: anchored to the pid that earned it, consulted last, dropped when that process leaves the pane or on a DECRST that does not carry libuv's `0x0208` fingerprint. The #598 audience is unchanged (14/14 on that suite).
* The "unearnable" Claude Code pane root caused to an attribution race (the DECSET parsed while the launcher shim was still the foreground leaf, then frozen), closed with tmux's `alternate_on` term.
* Your `PSMUX_FORCE_WHEEL` is in the tree verbatim in shape, `FORCE_WHEEL_ENV` + `wheel_gate_forced()` in `ssh_input.rs` with the same vocabulary as `PSMUX_FORCE_MOUSE` and the same independence from it, documented as the server wide last resort. Next to it there is the per pane spelling you offered to rework towards: `set-option -p -t %N @mouse-force on`, which wins over the env var in both directions.

Your measurement table and the libuv observation are what made the real fix possible, so thank you for the care that went into both the issue and this PR.

PR #617, opened 2026-08-29
fix(render): restore global window styles

MattKotsenas:
## Summary

- apply foreground and background defaults from global `window-style` and `window-active-style` across tiled and floating panes
- include both options in each live render-state payload and deserialize them in the client
- apply `set-option -a` changes immediately and allow `set-option -o` after `set-option -u`

## Why

psmux accepts and reports these tmux options, but the live client renderer does not apply them. This leaves configurations that assign different default colours to inactive and active panes with no visible effect.

Inactive panes use `window-style`; the active tiled or focused floating pane uses `window-active-style`. These styles fill omitted cells and replace reset foreground and background values. Explicit application colours still win. Clock mode keeps the pane's resolved background behind the clock. Unsetting either option immediately repaints attached clients.

Cross-session chooser previews use serialized cell colours because `window-dump` does not include `window-style` or `window-active-style`.

## Validation

- `cargo check --all-targets`
- `cargo test --bin psmux window_content_styles` - 9 passed
- `cargo test --bin psmux floating_render` - 6 passed
- `cargo test --bin psmux tests_live_client_styles` - 5 passed
- `cargo test --bin psmux config_file_window_style` - 1 passed
- `cargo test --bin psmux config_file_window_active_style` - 1 passed
- `cargo test --bin psmux config_only_if_unset_sets_window_style_after_unset` - 1 passed
- `tests/test_live_client_styles.ps1` - 12 passed, 0 failed

### DIFF
--- a/docs/preview.md
+++ b/docs/preview.md
@@ -87,7 +87,7 @@ psmux aims to keep the preview feature on par with tmux, with a few intentional 
 * `choose-session` and `choose-tree` both have a preview pane.
 * `p` toggles the preview while a chooser is open.
 * The preview is a live mirror of the target, not a frozen snapshot.
-* Pane borders, colours, and styles are preserved.
+* Serialized cell colours and styles, plus pane-border geometry, are preserved.
 * Wide characters are handled correctly.
 * The preview width is roughly half the popup width, with the picker list on the left.
 * The preview never modifies the target session in any way (it is read only).
@@ -104,7 +104,7 @@ psmux aims to keep the preview feature on par with tmux, with a few intentional 
 
 * The option name `choose-tree-preview` is psmux specific. tmux does not recognise it. Adding it to a shared configuration file is safe because tmux's set-option command will warn but not fail; if you want to be strict, guard the line with `if-shell` or split your config.
 * The option key in `show-options` output and in the JSON sent to the client uses kebab-case (`choose-tree-preview`) and snake_case (`choose_tree_preview`) respectively, matching the existing psmux convention.
-* The preview uses the target's serialized cell styles and the chooser's pane-border styles. The current `window-dump` response does not include target `window-style` or `window-active-style` options.
+* The preview uses the target's serialized cell styles and the chooser's pane-border styles. The `window-dump` response does not include target `window-style` or `window-active-style` options.
 
 ## Performance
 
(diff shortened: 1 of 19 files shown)


psmux:
Merged, thank you Matt. This is a lovely PR: small surface, dependency free, byte identical when the options are unset, and it came with the raw pseudoconsole harness already extended so the proof travels with the fix.

**What I did before merging**

* Reproduced the bug on current master first, with the repo's own ConPTY host grepping the raw VT stream: `window-active-style fg=colour201,bg=colour21` was stored (visible in `show-options -g`) and produced zero `38;5;201` / `48;5;21` in the pane body, while the control `window-status-current-style` painted both. On your build both needles appear in the pane body and, on a split, `window-style` paints the inactive pane too.
* Checked the semantics against tmux source: `tty.c` `tty_default_colours` only ever fills fg and bg that are still the terminal default, `style.c` `style_add` lets explicit application colours win, borders and the status line come from their own options, and `options.c` flags every pane on set so the change is immediate. Your `apply_window_content_style` replacing only `Color::Reset`, the `Paragraph` fill for never written cells, and the `state_dirty` on `-a` / `-u` all match. I also confirmed an explicit `SGR 41` background from the app survives untouched next to a `48;5;21` window background.
* Full `cargo test` on the merged tree: 9693 passed, 0 failed across 14 binaries, zero build warnings. `tests/test_live_client_styles.ps1`: 12 passed, 0 failed. The rename orphans nothing; the runner and CI discover suites by glob.
* The `flag 64` change from a single space to `" ".repeat(run_w)` is a real fix on its own: the old code advanced the column by `run_w` while emitting one column, so any hidden run wider than one cell shifted the rest of the row left.

**One measured divergence from tmux, not a blocker and not introduced by you**

tmux falls back to `window-style` for the active pane when `window-active-style` names no fg or bg (`tty.c`, the `cached_active_gc.fg != 8` test). With only `set -g window-style bg=colour52` and a single active pane, tmux paints it and psmux does not. psmux's legacy renderer in `src/rendering.rs` has had the same gap all along, so your change is consistent with the existing model; I opened #619 so both renderers gain the fallback together rather than making them disagree. It also covers `-o` after `-u` for the rest of the options and the two remaining single space `flag 64` sites in `client.rs` (popup runs) and `preview.rs`.

I fixed the `docs/configuration.md` wording for `window-style` on top of the merge (832c9be) so it now reads the way tmux's own options table describes it. Thanks again, this closes a gap people have quietly been living with.

PR #624, opened 2026-09-01
fix(render): preserve pane border backgrounds

MattKotsenas:
## Summary

This PR fixes pane-border backgrounds so colours configured through `pane-border-style` and `pane-active-border-style` render consistently around panes and in previews.

## User impact

With this change, settings such as:

```tmux
set -g pane-border-style 'fg=colour244,bg=colour235'
set -g pane-active-border-style 'fg=colour220,bg=colour52'
```

produce continuous inactive and active colour bands across straight separators, corners, intersections, pane labels, normal and floating panes, and `choose-tree` previews.

Themes that already set a pane-border background will start showing it. To keep an uncoloured border background, remove `bg=...` from the style or set `bg=default` explicitly.

## Behaviour after this change

- on border cells, `pane-border-style` and `pane-active-border-style` take precedence; `window-style` and `window-active-style` continue to style pane content
- each configured pane-border style sets its foreground and background outright; an omitted colour uses the terminal default
- unsetting the options returns inactive borders to dark grey and active borders to green
- `pane-border-hover-style` now applies background and attributes, but continues to leave omitted fields unchanged, so an `fg`-only hover style keeps the border background

## Validation

- passed: `cargo test --all-targets`
- built: `cargo build --release --target x86_64-pc-windows-msvc`
- passed: `tests/test_pane_border_background.ps1`
- passed: `tests/test_live_client_styles.ps1`: 12 passed, 0 failed


psmux:
Merged in 0a4e960. Thank you, this is an exemplary PR: the summary matched the diff exactly, the validation section was reproducible, and the new `tests/test_pane_border_background.ps1` is a genuine regression test (I confirmed it fails on the pre PR build with "Background sequence was not emitted", so it is guarding something real).

Before merging I re verified everything independently on a rebase onto current master (clean, zero conflicts):

* On the wire, pre PR emitted only `38;5;244` / `38;5;220` at the borders with the configured backgrounds absent; with your change all four of `38;5;244`, `48;5;235`, `38;5;220`, `48;5;52` are present, junctions sit inside the styled runs, and pane labels carry the border background too.
* tmux parity checked against the source: `window-border.c` and `style.c` start border cells from the default cell rather than window-style, and `style.c` leaves an omitted bg at the grid default, both of which your precedence and unset semantics match.
* Full suite: 3295 unit tests and 129 E2E assertions across the border, style, theme, zoom and window-style suites, zero failures. Your 11 new tests all pass and none of the existing ones broke.

Two small things I noticed while reviewing, neither blocking and neither introduced by you, are now tracked separately (see the follow up issue): attributes in border styles (`bold` etc) are dropped where tmux ORs them in, and `set -gu pane-border-style` lands on terminal default fg instead of the startup DarkGray because the option catalog's default string is literally "default". Your change actually brought the unset path closer to tmux than it was before.

One tiny ask if you touch this area again: `tests-rs/test_pane_content_box_drawing.rs` lost its longer #452 rationale comment in the rewrite; a one line "#452" reference would keep the history findable.

Thanks again, and the conpty host env var overrides are a nice touch for harness reuse. This ships with the next release.

Issue #626, opened 2026-09-02
pane border styles: attributes dropped vs tmux, and -u default disagrees with startup DarkGray (follow ups from PR #624 review)


psmux:
Two related pane border style gaps surfaced while reviewing and merging PR #624. Neither was introduced by that PR; both predate it and are catalog or style pipeline level, so they get their own issue.

**1. Border styles drop attributes where tmux ORs them in**

`pane-border-style 'fg=red,bold'` renders red but not bold. `complete_pane_border_colors` keeps only fg and bg, discarding `add_modifier` and `underline_color`. tmux applies attributes on border cells: style.c:459 does `gc->attr |= sy->gc.attr;`. The unit test `pane_border_colors_keep_colors_and_ignore_attributes` added in #624 currently enshrines the divergence, so fixing this means updating that test too.

**2. `set -gu pane-border-style` lands on terminal default fg instead of the startup DarkGray**

Repro on current master (0a4e960):

```
psmux new-session, split, observe inactive border emits ESC[90m (DarkGray), active ESC[32m
set -g pane-border-style 'fg=colour244,bg=colour235'
set -gu pane-border-style
```

After the unset, the active border correctly returns to ESC[32m, but the inactive border emits no colour at all (terminal default fg), not ESC[90m. A fresh session on the same build shows ESC[90m again, so unset and startup disagree.

Root cause: option_catalog.rs line ~101 stores the default for pane-border-style as the literal string "default"; `parse_tmux_style("default")` yields an empty style, and the renderer then pins fg to Reset. The catalog default should express what startup actually renders (the DarkGray and Green pair), or the renderer should map the empty style back to the built in defaults.

For what it is worth, tmux 3.4 defaults pane-border-style to `fg=themelightgrey` (options-table.c:1605) and unset returns to exactly that, so startup and unset never disagree there. That is the behaviour to match.

psmux:
Both gaps are fixed on master in 0c0bc4b (merged in 852d94d), with the #624 test updated in c7c3db0 and E2E coverage in 2c0f432.

**Gap 1: border styles dropped attributes.**
complete_pane_border_colors rebuilt the border style from Style::default() and copied only fg and bg across, so add_modifier, sub_modifier and the us= underline colour never reached the cell. tmux does not replace the attribute half of a border style, it merges it. From style.c:459:

```c
if (sy->gc.fg != 8) gc->fg = sy->gc.fg;
[...]

Issue #619, opened 2026-08-30
window-style: active pane fallback, generic -o after -u, remaining hidden-run sites (follow ups from #617)


psmux:
Follow ups measured while verifying #617 (merged in ce6a00b). None of them regress anything that PR shipped; they are gaps that pre date it or sit next to it.

**1. Active pane fallback to `window-style` (tmux parity)**

tmux `tty.c` `tty_default_colours` uses `window-active-style` for the active pane only when it actually names a fg or bg (`cached_active_gc.fg != 8`), otherwise it falls back to `window-style`. So in tmux `set -g window-style bg=colour52` alone tints every pane including the active one, and `set -g window-active-style bg=default` means "inherit `window-style`".

Measured on the merged build with the raw ConPTY harness: `window-active-style` unset, single active pane, `window-style bg=colour52`, status off: `48;5;52` absent from the stream. tmux would paint it.

Both psmux renderers pick one style or the other with no fallback: `src/client.rs` `WindowContentStyles::for_pane` (live client) and `src/rendering.rs` around line 385 (legacy local renderer). They should gain the fallback together so the two paths keep agreeing. In the live client model this is `for_pane(true)` returning `active` merged over `inactive` rather than `active` alone.

**2. `set-option -o` after `set-option -u` for every option**

#617 fixed `-o` after `-u` for `window-style` and `window-active-style` only. `src/config.rs` `parse_set_option` and the server `SetOptionUnset` arm leave the key in `user_set_options` for every other non `@` option, so `set -o` after `set -u` is silently ignored for the whole class. A generic `user_set_options.remove(key)` in the unset branch would fix it; needs its own tests.

**3. Two remaining single space `flag 64` (hidden) run sites**

#617 changed the pane content renderer to emit `" ".repeat(run_w)` for hidden runs because the old single space advanced the column accounting by `run_w` while painting one column, shifting the rest of the row. The same pattern is still present at `src/client.rs` (popup and menu run renderer, around line 6629) and `src/preview.rs` (chooser preview run renderer, around line 529).

**4. `dim=N` in `window-style` (minor, newer tmux)**

Current tmux consumes a `dim` value from both window styles (`tty.c` `tty_style_changed`, applied via `colour_dim`). psmux reads only fg and bg out of `parse_tmux_style` for these two options. Low priority.

psmux:
[...]

Issue #618, opened 2026-08-29
`set-option`/`set`/`setw` reject `-s` (server-option scope) entirely, breaking any tool targeting tmux 3.2+ semantics (e.g. thurbox's `default-terminal` bootstrap)


TGiles:
**Environment:** psmux 3.3.8 (66cf613, 2026-08-18), Windows 11 (build 26200), invoked via the `tmux` alias from Git Bash inside Windows Terminal

**Summary:** tmux 3.2 moved several options, including `default-terminal`, from session scope to **server scope**, settable via `set-option -s <option> <value>`. tmux can usually infer server scope from the option name without `-s`, per the man page: "If the option is not a user option, -w or -s may be unnecessary - tmux will infer the type from the option name"; `-s` is still correct, current tmux syntax that a version-safe or precision-minded caller reasonably passes explicitly.

psmux's `set-option`/`set`/`setw` parser has never implemented `-s`: at commit 66cf613 (this report's pinned build), `docs/tmux_args_reference.md` listed its accepted flags as `guaqowt:p:` (booleans `-g`/`-u`/`-a`/`-q`/`-o`/`-w` plus value flags `-t`/`-p`); no `-s`. Current master's docs string has since evolved (`guaqopwUt:` plus `--`), but `-s` remains absent from both, and the source-level enforcement hasn't moved: `src/main.rs` (function `run_main`) still rejects any character outside `agopqtuUw` for `set-option`/`set`/`setw`, confirmed against both the installed 66cf613 binary and current master source (HEAD `43302a8`, 2026-08-29).

Prior issues checked for duplicates: #535, #499, #547, #565, #583 (closed), #580 (open), and #553 (closed, the issue that put the current `set-option`/`show-options` flag validation in place and documents `-s` as deliberately unaccepted by `set-option`). None cover this gap; #583 is a flag-parsing-order bug, not a missing flag.

[...]

Issue #606, opened 2026-08-27
repeat-time Is Not Recognized


CalebMostyn:
Setting `repeat-time` does not work, despite being listed under `docs/configuration.md`. It does not appear to matter what value is used

conf:
```
set-option -g repeat-time 0 # also tried 1, 250, 1000, 3000, etc
```

Versions:
```
psmux --version
tmux 3.3.7
psmux 3.3.7 (05cc5d4 2026-07-20)

 [System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      26200  0

```

Logs:
```
===== autorename.log =====
[10:38:17.434] root=30448 no_direct_children
[10:38:17.435] pid=30448 fg_child=None (BFS found nothing)
[10:38:17.435] pid=30448 no_foreground_child
[10:38:18.461] root=30448 snapshot_entries=385
[10:38:18.461] root=30448 no_direct_children
[10:38:18.462] pid=30448 fg_child=None (BFS found nothing)
[10:38:18.462] pid=30448 no_foreground_child
[10:38:19.449] root=30448 snapshot_entries=384
[10:38:19.450] root=30448 no_direct_children
[10:38:19.450] pid=30448 fg_child=None (BFS found nothing)
[10:38:19.450] pid=30448 no_foreground_child
[10:38:20.483] root=30448 snapshot_entries=384
[10:38:20.483] root=30448 no_direct_children
[10:38:20.483] pid=30448 fg_child=None (BFS found nothing)
[10:38:20.483] pid=30448 no_foreground_child
[10:38:40.415] root=29860 snapshot_entries=385
[10:38:40.415] root=29860 no_direct_children
[10:38:40.415] pid=29860 fg_child=None (BFS found nothing)
[10:38:40.416] pid=29860 no_foreground_child
[10:38:41.448] root=29860 snapshot_entries=385
[10:38:41.449] root=29860 no_direct_children
[10:38:41.449] pid=29860 fg_child=None (BFS found nothing)
[10:38:41.450] pid=29860 no_foreground_child
[10:38:41.802] root=30448 snapshot_entries=385
[10:38:41.802] root=30448 no_direct_children
[10:38:41.802] pid=30448 fg_child=None (BFS found nothing)
[10:38:41.803] pid=30448 no_foreground_child
[10:38:42.808] root=30448 snapshot_entries=385
[10:38:42.808] root=30448 no_direct_children
[...]

PR #628, opened 2026-09-03
Add configurable active-pane border indicators

MattKotsenas:
## Summary

This PR makes the active-pane cues on pane borders configurable with `pane-border-indicators`. The existing `colour` behavior remains the default, so current configurations keep the same focus indication.

- `off` draws no split or arrows; the focused pane's border keeps `pane-active-border-style`.
- `colour` splits a two-pane divider between the active and inactive border colours.
- `arrows` places inward-pointing markers on dividers around the active tiled pane and on the focused floating pane.
- `both` combines the colour and arrow cues.

When a floating pane owns focus, the tiled layout no longer highlights an inactive pane.

Chooser previews now carry the target window's border line mode, border styles, indicator mode, and floating-focus state. `choose-tree` and `choose-session` therefore show the same active-pane cues as the target window while still omitting floating panes.

## Screenshots

Pane border indicator modes: off, colour, arrows, and both

## Compatibility and option behavior

`set-option -u pane-border-indicators` restores the default. The option is a global/default window setting. It rejects per-window overrides and unsupported values.

The preview protocol keeps its layout-only fallback. A client that receives no preview state uses single border lines, chooser border colours, and `colour` indicators.

This PR also shares `set-option` argument parsing across the CLI, TCP, and control-mode paths, while config parsing uses the same option catalog validation. Missing or invalid values are rejected before dispatch without changing existing `-s`, `-u`, `-o`/`-q`, append, target, or literal-value behavior.

## Implementation

The renderer models the four modes as a typed `PaneBorderIndicators` value. Tiled borders use shared geometry to place arrows beside the active pane and split the divider colour in two-pane windows; floating borders apply the same mode after drawing the pane title. Preview state extends `window-dump` with the target's border settings.

## Validation

- All 3,335 Rust tests pass per binary target.
- All four smoke integration suites pass.
- Live server-scope and set-option diagnostic regression suites pass.
- Live pane-border background, indicator, preview, and client-style scenarios pass.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -207,6 +207,7 @@ Details worth knowing:
 | `pane-border-style` | Str | | Inactive border style |
 | `pane-active-border-style` | Str | `fg=green` | Active border style |
 | `pane-border-hover-style` | Str | `fg=yellow` | Border style while the mouse hovers a draggable pane border |
+| `pane-border-indicators` | Str | `colour` | Active pane cues: `off`, `colour`, `arrows`, `both`. See [Pane Border Indicators](#pane-border-indicators) |
 | `pane-border-lines` | Str | `single` | Border glyph set: `single`, `double`, `heavy`, `simple`, `number`, `spaces`, `none`. See [Pane Border Lines](#pane-border-lines) |
 | `pane-border-format` | Str | | Pane border format string (e.g. `#{pane_index}: #{pane_title}`) |
 | `pane-border-status` | Str | | Pane border status position (`top`/`bottom`/`off`) |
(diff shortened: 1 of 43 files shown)


psmux:
Merged. This is an outstanding contribution, thank you. The shared set-option parser replacing four divergent hand rolled ones, the typed option catalog, and the depth of the test coverage (the ConPTY byte stream and preview renderer verification especially) made this a pleasure to review. The full review ran your branch through a security pass, all 10125 of its Rust test executions, both of your E2E scripts, and live regression checks against our recent set-option work (#583, #606, #618, #619, #626), and everything held.

Landed as merge 445e037 with three follow up commits on top, now on master at 1c41131:

- 38c5c03 retypes `main-pane-width` and `main-pane-height` from `Number(U16)` to strings. tmux types both as string options and documents the percentage spelling (options-table.c), so `set -g main-pane-width 50%` must not be rejected. The percentage now also round trips into the stored value instead of being silently dropped. While in there, integer options that parse but fall outside their range now report tmux's wording (`value is too small: -1`) instead of `expected a number`.
- 5424733 adds `F` to the accepted set-option flags and honours it on the CLI, TCP and control mode routes: the value is format expanded on the server thread before validation and storage, matching what config files already did. Your validate before dispatch structure made this a clean insertion.
- 4e4e358 makes `pane-border-indicators off` a genuine no cue mode: every separator keeps `pane-border-style` and no arrows are drawn. As reviewed, `off` previously fell through to the generic path and painted the whole adjacent divider with the active style, which read as a stronger highlight than the default. `colour`, `arrows` and `both` render byte for byte as in your PR; your `off` test is renamed and re pinned to the new behavior, and a new test pins that `arrows` keeps the active styled divider so the two cannot drift into each other. The docs table in configuration.md now spells out all four modes.

Two merge notes: the branch predated the #626 border attribute fix, so `complete_pane_border_colors` was resolved to the attribute preserving form and the catalog default for `pane-border-style` stayed `fg=brightblack`; and the two hardcoded DarkGray/Green pairs in the preview path now route through the shared `pane_border_default_style` helper so the defaults cannot drift between renderers.

After integration with the two fixes that landed on master while this was in review (#629, #630), the combined tree runs 10410 Rust test executions with zero failures, and both of your E2E scripts pass and have joined the full sweep automatically.

One open thought from the review, no action needed: `pane-border-indicators` refuses the bare `setw` form while other window options accept it and store globally. That inconsistency is on our side to settle and I will track it separately. Thanks again for a genuinely excellent PR.

PR #633, opened 2026-09-04
Centralize server-to-client render options

MattKotsenas:
## Summary

Server-to-client render options now have one typed owner. Status styles, window content styles, and pane-border indicators are expanded once, collected into `ClientRenderOptions`, and flattened into the existing render-state payload. The requested dump-state response and automatic render-state push therefore share one serialization path.

This is an internal refactor with no intended user-visible behavior change.

## Compatibility

The wire format remains unchanged. Existing keys, including the short `wsa_style`, `wsb_style`, and `wsl_style` names, retain their current names and values. Missing fields still use their existing client defaults.

## Implementation

The style-option expansions move into `expand_status_formats`, so one guarded helper covers every per-frame expansion before the typed options are serialized. An invalid typed option or malformed payload construction propagates through `run_server` instead of silently omitting the fields.

The change also removes the unused server-side pane renderer, its `AppState` status and layout helpers, and the `style.rs` re-exports. Client call sites import parsing directly from `style.rs`; shared terminal helpers and the live client's `BorderGeometry` intersection model remain in `rendering.rs`.

## Validation

- `cargo test --all-targets`: 3,432 passed per binary target; all integration and example targets passed
- `cargo build --release --target x86_64-pc-windows-msvc`: passed
- `tests/test_smoke_pr.ps1`: 5 passed, 0 failed
- `tests/test_session_mgmt.ps1`: 14 passed, 0 failed
- `tests/test_run_shell.ps1`: 11 passed, 0 failed
- `tests/test_issue209_e2e_verify.ps1`: 33 passed, 0 failed


psmux:
Merged. Thanks Matt, this is a nice piece of work and it held up under everything I threw at it.

I checked it against master rather than taking the no behavior change claim on faith. Two identically built binaries, separate socket namespaces, and a dump-state diff across seven scenarios: defaults, every style option set, all four `pane-border-indicators` modes, after `-u` unset, format bearing style values, after deliberately bad values, and after a split. The output is byte identical, key order included, so `wsa_style`, `wsb_style`, `wsl_style` and the rest really do land on the wire exactly as before.

I paid extra attention to the pane border indicator work that landed just before this, since your change moves where that value is read. Rendering all four modes through `_render-preview` on a two pane split gives four distinct outputs, and each one matches master byte for byte, so `off` is still a genuine no cue mode and the colour split survives for `colour` and `both`.

Test numbers: 10416 passed and 0 failed on your branch, the same 10416 and 0 on master, and the full workspace is green again after the merge. The PowerShell suites you touched all pass too, along with both pane border indicator suites.

One thing I did change, in 2a63d08 right after the merge. `expand_status_formats` returned a Result and both call sites in `run_server` used the question mark, so an option value that failed to parse would not just degrade the render, it would end the server and take every pane with it. I went looking for a way to trigger it and could not find one: `set-option` with `-g`, `-w`, `-u` and `-F`, and the config file route, all validate or refuse, and the server stayed up through every hostile value I fed it. So it was latent rather than live. But it meant any future write into `user_options` that bypassed the option catalog would turn a cosmetic option into total session loss. The follow up makes the expansion infallible, falling back to the catalog default with a debug log, which is the same fallback the client already applies to a missing field, and the two `append_client_render_options_json` call sites now skip the append on error the way the old no op guard did. The typed ownership you built is untouched, and there are regression tests pinning the fallback.

Two smaller notes from the review, neither needing action from you. `style::parse_status` and `style::expand_status` now have no callers at all once the `rendering.rs` wrappers are gone; they were already dead behind those wrappers, so that is a candidate for the same sweep someday, not something you introduced. And the escaping of control characters inside style values changed shape slightly, because serde_json writes the short two character escapes for backspace and form feed where the hand rolled escaper wrote the six character unicode form. The decoded bytes are identical and every consumer uses a real JSON parser, so it changes nothing; I mention it only so nobody writes a byte exact golden test on that payload later.

I also went in expecting `serde(flatten)` to cost something on the per frame deserialize, since it buffers keys the outer struct does not declare. Benchmarked against a real captured payload it came out inside the noise, so that worry was unfounded.

Removing the unused server side pane renderer is very welcome. I confirmed every one of those functions had zero callers left anywhere in the tree, so that is 439 lines of pure win. Thanks again, and I would be glad to see more of these.

PR #635, opened 2026-09-05
fix(kill-window): require a value for -t

MattKotsenas:
## Summary

This PR fixes `kill-window -t` and `killw -t` so a missing target fails with `-t expects an argument` before server connection, command storage, dispatch, or window mutation.

The same fix has to hold across the process CLI, local execution, simple TCP, control mode, config, bindings, hooks, `confirm-before`, and aliases. One typed command parser therefore owns `kill-window` option parsing.

## Details

- The parser uses `lexopt` and handles separate, attached, clustered, and equals-form targets. Repeated `-t` uses the last value, `--` ends option parsing, and unknown options are ignored.
- The CLI sends the composed target through the existing `TARGET` protocol line. Relative targets that name a window inherit the transport session; window targets with an explicitly qualified session replace it.
- Invalid local commands return `InvalidInput` and set the status message.
- Recursive deferred-command nesting is capped at 64 levels.
- Server request types and execution effects are unchanged. New parse errors use the existing one-shot and control-mode error formats.
- Tests cover each path, aliases, deferred commands, target composition, mutation safety, and early CLI rejection.

## Validation

- Full validation on an isolated `psmux-dev` VM
- `cargo test --all-targets`: the shared 3,456-test unit suite passed for each binary alias, along with all integration and example targets
- Release build succeeded
- `test_smoke_pr.ps1`, `test_session_mgmt.ps1`, `test_run_shell.ps1`, and `test_issue209_e2e_verify.ps1`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -940,6 +940,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
+name = "lexopt"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "803ec87c9cfb29b9d2633f20cba1f488db3fd53f2158b1024cbefb47ba05d413"
+
+[[package]]
 name = "libc"
 version = "0.2.189"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1369,6 +1375,7 @@ dependencies = [
  "chrono",
  "crossterm",
  "glob",
+ "lexopt",
  "parking_lot",
  "portable-pty-psmux",
  "ratatui",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,10 @@ name = "test_pr469_osc_command_safety"
 path = "tests-rs/test_pr469_osc_command_safety.rs"
 
 [[test]]
+name = "test_kill_window_cli_error"
+path = "tests-rs/test_kill_window_cli_error.rs"
+
+[[test]]
 name = "test_issue494_copymode_freeze"
 path = "tests-rs/test_issue494_copymode_freeze.rs"
 
@@ -117,6 +121,7 @@ regex = "1.13.1"
 glob = "0.3.4"
 anyhow = "1.0.104"
 base64 = "0.23.1"
+lexopt = "0.3.2"
 
 [profile.release]
 opt-level = 3
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -632,6 +632,89 @@ pub fn parse_command_to_action(cmd: &str) -> Option<Action> {
     }
 }
 
+/// Joins tokens into a command string while preserving ordinary token
+/// boundaries; `;` and `\;` remain command-chain separators.
+pub fn requote_command_tail<S: AsRef<str>>(args: &[S]) -> String {
+    args.iter()
+        .map(|token| {
+            let token = token.as_ref();
+            if token.is_empty() {
+                "''".to_string()
+            } else if token == ";" || token == "\\;" {
+                token.to_string()
+            } else if token.contains('\'') {
+                format!(
+                    "\"{}\"",
+                    token.replace('\\', "\\\\").replace('"', "\\\"")
+                )
+            } else if token.chars().any(|c| c.is_whitespace() || c == '"') {
+                format!("'{}'", token)
+            } else {
+                token.to_string()
+            }
+        })
+        .collect::<Vec<_>>()
+        .join(" ")
+}
+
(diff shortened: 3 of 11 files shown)


psmux:
Thank you for this. Your diagnosis is exactly right, and it turned out to be the thread that unravelled something considerably larger than `kill-window`. I have landed a fix crediting you in af0d353, but it is not your patch, and I want to explain why rather than just closing this.

## Your diagnosis is confirmed

I reproduced it against unmodified master before touching anything. In `src/main.rs` the `-t` arm does `if let Some(t) = cmd_args.get(i + 1)`, so a trailing `-t` with no value silently drops the flag, the wire command becomes a bare `kill-window`, and `server/connection.rs` then takes the no-target branch and kills the active window. The server replies empty, so the CLI's own error check at `main.rs:2735` passes and you get exit 0 with no message. Losing the wrong window silently is exactly the kind of bug I want reported.

Your patch is also clean work, and I checked it properly rather than skimming it. The `lexopt` 0.3.2 checksum matches the crates.io index and API exactly, sole owner blyxxyz, zero transitive dependencies, and the lock file diff adds nothing else. `requote_command_tail` is a straight relocation of the existing helper with identical logic, so there is no new quoting surface. The recursion cap checks depth before recursing and your test exercises 65 levels. Both new test files are pure additions and nothing existing was weakened.

## Why I went generic instead

After your report I probed master with a dangling flag across every command form I could construct. 71 of 71 accepted it at exit 0. `kill-window` is one of them, and it is not the worst.

```
kill-window -t      0    SIDE-EFFECT 5->4    <empty>
kill-session -t     0    SIDE-EFFECT 5->0    <empty>
unlink-window -t    0    SIDE-EFFECT 5->4    <empty>
respawn-pane -c     0    SIDE-EFFECT 9->0    <empty>
new-window -t       0    SIDE-EFFECT 5->6    <empty>
link-window -s      0    SIDE-EFFECT 8->9    <empty>
run-shell -d        1    5   -d: The term '-d' is not recognized  <- executed its own flag
```

`kill-session` is the one that changed my mind. `main.rs:2877` falls back to `PSMUX_TARGET_SESSION`, the server sets that to its own session at `server/mod.rs:1016`, and panes inherit it. So `psmux kill-session -t` run inside a pane destroys that whole session at exit 0. I confirmed it end to end: the pane echoed back `PTS=[ksprobe]`, the command ran, and the server was gone.

The trigger for both is the same everyday mistake that presumably found yours: `psmux kill-session -t $TARGET` in a script where `$TARGET` is unset and unquoted. The shell drops the word entirely and leaves a bare `-t`.

And tmux treats this as one problem, not seventy. `arguments.c:180` raises `-%c expects an argument` generically, driven by each command's `.args` template, before the command executes at all.

So `KillWindowCommand::parse` being keyed on `matches!(command, "kill-window" | "killw")` is the sticking point. Merging it would have fixed the case you hit and left `kill-session` still silently destroying sessions, which is the more expensive one.

## What landed instead

One table of 110 entries, 96 transcribed verbatim from tmux's own `cmd-*.c` `.args` strings, and one validator in `src/cli.rs` called from all eight ingresses: CLI dispatch, the TCP handler, control mode, local and prompt execution, the config file, and the three `bind-key` routes. It runs ahead of all 101 `cmd_args.get(i + 1)` sites rather than rewriting them, so the hole is closed by construction. No new dependency; the parser copies `args_parse_flags` semantics directly, so `kill-window -t -a` still targets `-a`, `resize-pane -x -5` still works, optional values are never made mandatory, and unknown letters stay boolean so nothing psmux accepts today is newly rejected.

All 71 forms now reject with exit 1 and no side effect, and 27 legitimate command lines (`send-keys -t S -- -foo`, `capture-pane -p -S -5`, the attached `-tS` and equals `-t=S` forms, the aliases) are unchanged. I verified both halves myself on top of the automated suites, because a false rejection here would be worse than the original bug.

Your review also turned up a genuine test defect I have fixed: `test_refresh_client_flags` was driving `-C`/`-B`/`-A`/`-f` bare to reach a later check, but all four take a value in tmux, so bare is a parse error there too.

## Two follow ups from reading your patch

Both are for the generic version, and both are things you got right that I want to carry forward:

- Unknown long options are swallowed by `Long(_) => { let _ = parser.optional_value(); }` where tmux errors with `unknown flag`. I have deliberately not implemented unknown-flag rejection yet, because psmux carries flags tmux has no template for and I would break them. It means `kill-server -t` is still accepted silently. That is a real remaining gap and I would take a PR for it.
- Adjacent, and separate from this bug: `unlink-window` ignores `-t` entirely on the CLI and always acts on the active window, and `run-shell` does not parse `-d`/`-c`/`-s`/`-t` at all. Also `bind-key` and `unbind-key` are fire-and-forget on the CLI, so any server side error is dropped and the CLI exits 0.

Closing this as superseded rather than rejected. The find is yours and the commit says so. Thanks for filing it with a working patch instead of just a report; that is what made the wider probe obvious to run.

PR #636, opened 2026-09-06
fix(build): refresh provenance after source edits and worktree changes

andrew-tawfeek:
## Problem

Cargo can rebuild changed source while reusing the previous build script's clean/dirty metadata. Watching `.git/HEAD` directly also misses the metadata layout of linked worktrees. A rebuilt executable can therefore identify stale source provenance.

## Fix

- Watch tracked files and Git-resolved HEAD/index/ref paths, including packed refs.
- Avoid watching missing optional metadata files, which would force needless rebuilds.
- Run Git queries without optional locks so `git status` does not refresh the watched index and invalidate the next otherwise unchanged build.

## Validation

Verified clean → source-only dirty → restored clean transitions, tracked documentation edits, linked-worktree HEAD changes, and a no-op build that stays fresh in an isolated Cargo/Git fixture. The combined fork also passed `cargo check --locked --all-targets` and an optimized Windows release build.

This changes only build provenance tracking; it is independent of the session reliability PR.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -22,11 +22,7 @@
 use std::process::Command;
 
 fn main() {
-    // Rebuild whenever HEAD moves (new commit, checkout, etc.) so the embedded
-    // hash never goes stale. These files may not exist in a tarball build; the
-    // rerun hints are harmless if absent.
-    println!("cargo:rerun-if-changed=.git/HEAD");
-    println!("cargo:rerun-if-changed=.git/refs");
+    watch_git_inputs();
 
     embed_windows_resources();
 
@@ -47,6 +43,29 @@ fn main() {
     println!("cargo:rustc-env=PSMUX_GIT_HASH_FULL={full}");
     println!("cargo:rustc-env=PSMUX_GIT_DIRTY={dirty}");
     println!("cargo:rustc-env=PSMUX_GIT_DATE={date}");
+}
+
+/// Cargo stops scanning package inputs once a build script emits an explicit
+/// rerun hint. Watch tracked files too: changing src/main.rs rebuilds the crate,
+/// but would otherwise reuse the previous build script's clean/dirty value.
+/// Resolve Git's metadata paths instead of assuming .git is a directory; in a
+/// linked worktree HEAD/index are private and refs/packed-refs are shared.
+fn watch_git_inputs() {
+    if let Some(files) = git(&["ls-files", "-z"]) {
+        for path in files.split('\0').filter(|path| !path.is_empty()) {
+            println!("cargo:rerun-if-changed={path}");
+        }
+    }
+    for name in ["HEAD", "index", "refs", "packed-refs"] {
+        if let Some(path) = git(&["rev-parse", "--git-path", name]) {
+            // A missing optional file would make Cargo rerun on every build.
+            // Packing refs also removes a watched loose ref; unpacking one
+            // changes the watched refs directory, so both transitions refresh.
+            if std::path::Path::new(&path).exists() {
+                println!("cargo:rerun-if-changed={path}");
+            }
+        }
+    }
 }
 
 /// Embed the icon and the VS_VERSIONINFO block (Windows hosts only).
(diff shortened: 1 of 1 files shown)


psmux:
This one holds up and I am happy to take it.

I confirmed the premise against the current file rather than taking it on trust. `build.rs:42` really does call `git status --porcelain`, and `git status` refreshes the index as a side effect, so once the index becomes a watched input it would invalidate every subsequent otherwise idle build, permanently. `--no-optional-locks` is the correct fix, and it is in the only position git accepts for a global option (`.arg("--no-optional-locks").args(args)` yields `git --no-optional-locks rev-parse ...`). Resolving `rev-parse --git-path` instead of hardcoding `.git/HEAD` is also the right handling for linked worktrees, which matters here because I run a lot of parallel worktrees. Skipping optional metadata files that do not exist avoids the needless rebuild you describe.

I read the whole thing line by line, because build.rs executes on every contributor and CI machine and that is the highest leverage place in a repo to hide something. It spawns nothing but git, makes no network calls, writes no files anywhere including OUT_DIR, embeds nothing new into the binary, captures no environment, and does not `include!` any generated content. The `rustc-env` values are unchanged; only the set of `rerun-if-changed` hints moves. Clean.

Two consequences I want on the record rather than discovered later:

`git ls-files` currently emits one `rerun-if-changed` line per tracked file, and this repo has 1267 tracked files. So Cargo stats 1267 paths and build.rs reruns five git subprocesses whenever any tracked file changes, documentation and tests included. That is the intended behaviour of your fix, not a defect, but it is a small ongoing build time tax and I would rather name it than have it surprise me later.

`--no-optional-locks` needs git 2.19 or newer. That is September 2018, so it is not a real constraint, just noting it.

CI is unaffected: ci.yml uses `actions/checkout@v7` at the default shallow depth, which still produces a full index and a working `rev-parse --git-path`.

Merging. Thanks for splitting this out from the reliability work instead of bundling it, that made it genuinely quick to review on its own merits.

psmux:
Merged as 4a508df, with your authorship preserved. Thank you, this is a good catch and the reasoning holds up.

I did not want to take it on the strength of the description alone, so I built an isolated fixture repository with two build scripts, the current one and yours, and ran the same sequence against each. The result is unambiguous:

```
== old ==
  clean tree, first build      : HASH=040225f DIRTY=false
  tracked source edited, build : HASH=040225f DIRTY=false      <-- stale
  source restored, build       : HASH=040225f DIRTY=false
  no-op rebuild reruns script? : 0

== new ==
  clean tree, first build      : HASH=b271b49 DIRTY=false
  tracked source edited, build : HASH=b271b49 DIRTY=true       <-- correct
  source restored, build       : HASH=b271b49 DIRTY=false
  no-op rebuild reruns script? : 0
```

So the failure you described is real and it is exactly as you characterised it. Emitting any rerun hint turns off Cargo's own package input scan, so watching `.git/HEAD` and `.git/refs` and nothing else meant a source only edit rebuilt the crate while quietly reusing the previous clean or dirty verdict. The binary then carried provenance that said it was reproducible from a commit it no longer matched, which is the worst kind of wrong for a field that exists to be trusted.

Same check on psmux itself after the merge:

```
psmux 3.3.8 (4a508df 2026-09-09)          clean tree
psmux 3.3.8 (4a508df 2026-09-09, dirty)   after touching src/help.rs
psmux 3.3.8 (4a508df 2026-09-09)          after restoring it
```

The other two parts are right as well. Using `git rev-parse --git-path` rather than assuming `.git` is a directory is the correct way to handle a linked worktree, where HEAD and index are private to the worktree while refs and packed refs are shared, and I use linked worktrees constantly so that path matters to me. The `.exists()` guard is the detail that makes it usable, since watching a path that is not there would otherwise force a rerun on every single build. And `--no-optional-locks` is the sharpest observation in the patch: once the index is a watched input, the `git status` call that computes the dirty flag would refresh it as a side effect and dirty its own input, so an idle tree would rebuild forever. I confirmed there is no such loop, a no-op rebuild reruns nothing.

Two small notes, neither of which blocked the merge. `git ls-files -z` emits one directive per tracked file, which is 1259 lines here, so the build script output is long, and Cargo now stats all of them on each build. That is the correct trade for accurate provenance and the cost is not measurable next to a compile. Also `-z` protects the parsing but the emitted lines are still newline delimited, so a tracked path containing a newline would confuse Cargo. There are none in this repository and Windows will not let you create one, so it is theoretical.

I have replied separately on #637.

psmux:
Landed on master as 4a508df with your authorship kept. Closing this since the cherry pick carries the change.

PR #637, opened 2026-09-06
fix(reliability): preserve live sessions and isolate command failures under load

andrew-tawfeek:
## Problem

Several independent failures can look like a psmux crash:

- A delayed discovery response can remove a live session's registration, followed by automatic orphan termination.
- A normal rejection such as `respawn-pane` without `-k` on a live pane can escape the server loop, terminate the server, and still leave the CLI reporting success.
- Failed AUTH or missing formatted output can appear as a blank successful `ls`.
- Rename/warm-claim races can overwrite another live owner's registration.
- A nonreading `pipe-pane` sink can block cancellation and the control loop; unbounded queues amplify latency and memory use.

The respawn failure was reproduced against the unmodified upstream runtime: client exit 0, server exit 1 with `pane still active`.

## Fix

- **Discovery and ownership:** keep observation read-only; preserve uncertain live registrations; identify processes by PID and creation time rather than executable basename. Reserve destination ownership before creation/rename/claim. Publish an atomic coherent manifest and clean up only the owning generation.
- **Command results:** require exact AUTH and complete bounded response frames. Incomplete listings retain healthy rows but return stderr/nonzero status. Keep command errors inside the event loop and attach them to their originating requests. Prepare respawn replacements before retiring the original pane.
- **Namespace routing:** resolve inherited and explicit namespaces consistently and encode ambiguous/path-unsafe storage names. An unresolved explicit shutdown target cannot kill the current session.
- **I/O and fairness:** bound allocated input, output staging, request admission/batches, handlers, and persistent/control queues. Give each sink an independent cancellable writer; move warm-pane refill and internal self-directed commands off the event loop. Propagate admission and permanent writer failures, including atomic paste rejection.
- **Lifecycle diagnostics:** append panic diagnostics without unregistering sessions; allocate session IDs under verified lock ownership with checked, atomic counter persistence.

The shared request/ack, registry, and writer APIs are included together so the change has one compilable transition. The build-provenance change is submitted independently as #636. The upgrade follow-up requires a coherent manifest before claiming a warm server, so legacy named sessions remain connectable while new sessions cold-start instead of adopting an old standby; four scratch-registry tests cover this path.

## Validation

- `cargo check --locked --all-targets` and compilation of the `psmux` unit-test harness passed.
- **200 targeted tests across 27 reviewed modules** passed, including fake AUTH peers, ownership rollback, renamed-executable identity, counter contention, panic-hook behavior, byte bounds/FIFO, partial writes, permanent failures, native Windows blocked-pipe cancellation, and control protocol regressions.
- A locked optimized Windows build passed **four discovery fault cases** and **55 lifecycle/load commands** in isolated data roots/namespaces. The latter includes failed respawn survival with original pane PIDs, foreign-live-owner collision, inherited/encoded routing, warm creation, output flood, nonreading sink cancellation, and capture verification. Slowest command in the final versioned-build probe: 204 ms.
- Existing user registrations/process identities were unchanged and all owned test processes were cleaned up. The full upstream suite was not run on the work machine because some tests can create default-namespace sessions or kill processes by image name; CI/disposable Windows remains the place for that suite.

Reproduction tools are `tests/audit_reliability_units.ps1`, `tests/audit_discovery_faults.ps1`, and `tests/audit_lifecycle_faults.ps1`. The runtime tools require an explicit executable SHA-256 and verify that the real baseline is unchanged.

## Tradeoffs and related work

- Overload returns a visible error or disconnects a slow consumer. A timed-out command has an unknown outcome; non-idempotent operations should not be blindly retried. Queue bounds do not cap total terminal-history/application memory.
- Legacy names already ambiguous under the old delimiter scheme are not silently migrated. A live process with missing/unreadable metadata is not automatically terminated. Existing old-code servers remain old code until restarted.
- A sink that exceeds its budget stops with an observable failure, so its transcript can be incomplete. Direct-file sink opening is asynchronous; delayed open failures reach pane diagnostics.
- This overlaps #541's discovery/reaping work: here even connection refusal is insufficient to reap a positively live server. It also overlaps #542's control buffering/ack work: this proposal disconnects slow consumers rather than silently shedding output. #635 overlaps command-ingress files. These are competing/overlapping changes, not a claim that the branches can be combined unchanged.

Long interactive/cross-version and overnight-load testing remain broader acceptance work; the bounded regressions are not a claim of exhaustive correctness.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,4 +1,4 @@
-use std::io::{self, Write, BufRead, BufReader};
+use std::io::{self, Write, BufReader};
 use std::time::{Duration, Instant};
 use std::env;
 
@@ -1860,19 +1860,17 @@ fn establish_connection_with_timeout(
     // answers would otherwise block the attach forever with no feedback. The
     // persistent read timeout is (re)set to 2s just below, after auth.
     let _ = reader.get_ref().set_read_timeout(Some(Duration::from_secs(3)));
-    let _ = writer.write_all(format!("AUTH {}\n", key).as_bytes());
-    let _ = writer.flush();
+    let key = crate::session::validate_auth_key(key)
+        .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "invalid session key"))?;
+    writer.write_all(format!("AUTH {}\n", key).as_bytes())?;
+    writer.flush()?;
     let mut auth_line = String::new();
-    let read_res = reader.read_line(&mut auth_line);
-    if std::env::var("PSMUX_AUTH_DEBUG").map(|v| v == "1").unwrap_or(false) {
-        eprintln!(
-            "[auth-debug pid={}] establish_connection addr={} key={} read_res={:?} got_line={:?}",
-            std::process::id(), addr, key, read_res.as_ref().map(|n| *n), auth_line
-        );
+    let n = std::io::BufRead::read_line(&mut std::io::Read::take(&mut reader, 4097), &mut auth_line)?;
+    if n == 0 || !auth_line.ends_with('\n') {
+        return Err(io::Error::new(io::ErrorKind::UnexpectedEof, "incomplete authentication acknowledgement"));
     }
-    read_res?;
-    if !auth_line.trim().starts_with("OK") {
-        return Err(io::Error::new(io::ErrorKind::PermissionDenied, "auth failed"));
+    if auth_line != "OK\n" && auth_line != "OK\r\n" {
+        return Err(io::Error::new(io::ErrorKind::PermissionDenied, "authentication rejected"));
     }
 
     let _ = writer.write_all(b"PERSISTENT\n");
@@ -1901,15 +1899,19 @@ fn establish_connection_with_timeout(
 
     // 2-second read timeout keeps the thread from blocking forever on process exit.
     let _ = reader.get_ref().set_read_timeout(Some(Duration::from_secs(2)));
-    let (frame_tx, frame_rx) = std::sync::mpsc::channel::<String>();
+    let (frame_tx, frame_rx) = std::sync::mpsc::sync_channel::<String>(4);
     std::thread::spawn(move || {
         let mut reader = reader;
         let mut buf = String::with_capacity(64 * 1024);
         loop {
             buf.clear();
             loop {
-                match reader.read_line(&mut buf) {
+                const MAX_CLIENT_FRAME_BYTES: usize = 16 * 1024 * 1024;
+                if buf.len() > MAX_CLIENT_FRAME_BYTES { return; }
+                let remaining = (MAX_CLIENT_FRAME_BYTES - buf.len() + 1) as u64;
+                match std::io::BufRead::read_line(&mut std::io::Read::take(&mut reader, remaining), &mut buf) {
                     Ok(0) => return,
+                    Ok(_) if buf.len() > MAX_CLIENT_FRAME_BYTES || !buf.ends_with('\n') => return,
                     Ok(_) => break,
                     Err(ref e) if e.kind() == std::io::ErrorKind::TimedOut
                         || e.kind() == std::io::ErrorKind::WouldBlock => {
@@ -1966,27 +1968,21 @@ pub(crate) fn no_such_session(name: &str) -> io::Error {
 
 pub fn run_remote(terminal: &mut Terminal<crate::platform::PsmuxBackend>, input: &crate::ssh_input::InputSource) -> io::Result<()> {
     let name = env::var("PSMUX_SESSION_NAME").unwrap_or_else(|_| "default".to_string());
-    let path = crate::paths::port_file(&name);
-    let port = std::fs::read_to_string(&path).ok().and_then(|s| s.trim().parse::<u16>().ok())
-        .ok_or_else(|| no_such_session(&name))?;
-    let addr = format!("127.0.0.1:{}", port);
-    // The .port file can be visible a beat before the .key has readable
-    // content (servers before the issue #496 fix wrote .port first, and a
-    // just-claimed warm server rewrites its files). Never AUTH with an empty
-    // key — that is guaranteed to fail — poll briefly for the credential.
-    let mut session_key = read_session_key(&name).unwrap_or_default();
-    if session_key.is_empty() {
-        for _ in 0..400 {
-            std::thread::sleep(Duration::from_millis(5));
-            session_key = read_session_key(&name).unwrap_or_default();
-            if !session_key.is_empty() { break; }
+    let deadline = std::time::Instant::now() + Duration::from_secs(2);
+    let (port, session_key) = loop {
+        match crate::session::read_session_endpoint(&name) {
+            Ok(endpoint) => break endpoint,
+            Err(error) if error.kind() == io::ErrorKind::NotFound && std::time::Instant::now() < deadline => {
+                std::thread::sleep(Duration::from_millis(5));
+            }
+            Err(error) => return Err(io::Error::new(error.kind(), format!("cannot read session '{}': {}", crate::paths::registry_session_name(&name), error))),
         }
-    }
-    let session_key = session_key;
+    };
+    let addr = format!("127.0.0.1:{}", port);
     if std::env::var("PSMUX_AUTH_DEBUG").map(|v| v == "1").unwrap_or(false) {
         eprintln!(
-            "[auth-debug pid={}] run_remote name={} port={} key={}",
-            std::process::id(), name, port, session_key
+            "[auth-debug pid={}] run_remote name={} port={}",
+            std::process::id(), name, port
         );
     }
     let last_path = crate::paths::psmux_dir_file("last_session");
@@ -2015,14 +2011,7 @@ pub fn run_remote(terminal: &mut Terminal<crate::platform::PsmuxBackend>, input:
         match establish_connection_with_timeout(&addr, &session_key, ATTACH_CONNECT_TIMEOUT) {
             Ok(conn) => conn,
             Err(e) => {
-                // Nothing answered on the registered port. Reap the entry so
-                // the next `ls`/`attach` is clean, but only once the pid anchor
-                // agrees the process is gone: a live server that was merely too
-                // slow must keep its registration (it rewrites these files on
-                // its own 5s tick anyway).
-                if crate::session::registry_pid_anchor_alive(&name) != Some(true) {
-                    crate::session::remove_session_registry(&name);
-                }
+                // A failed attach is observational: retain the registration.
                 if std::env::var("PSMUX_AUTH_DEBUG").map(|v| v == "1").unwrap_or(false) {
                     eprintln!(
                         "[auth-debug pid={}] attach connect to {} failed: {} (kind {:?})",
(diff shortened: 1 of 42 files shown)


psmux:
Thank you for this. I want to be straight with you about the outcome up front: I am asking you to split it rather than merging it as it stands. But I read all 8294 lines of it, I found real bugs in here that I am acting on, and the top item is going in with credit to you. So please read the rest of this as a plan rather than a rejection.

## The headline bug is real, and I reproduced it

I did not take this one on trust. Against unmodified master, in an isolated data root:

```
BEFORE: rsprobe: 2 windows (created Tue Sep  8 02:57:51 2026)

--- psmux respawn-pane   (NO -k, pane is LIVE => routine refusal) ---
client rc=0 out=[]

session alive AFTER routine refusal : False
AFTER : psmux: no server running on ...
```

A routine, expected refusal destroyed a two window session and the client reported success. The chain is exactly as you describe it: `window_ops.rs` returns `Err(..., "pane still active")`, the `?` in the RespawnPane arm of the event loop propagates it out of `run_server` (which is `-> io::Result<()>`), and the server exits. That is a serious bug and you found it.

I am landing that fix standalone, now, credited to you, rather than leaving it stuck behind 5000 lines of unrelated work. I am also having the same sweep done across every other `CtrlReq` arm, because respawn-pane is unlikely to be the only routine refusal wired to `?`, and I want the count.

## Why I cannot take the whole thing

Size and reviewability first. 5401 added, 2893 deleted, 42 files, in one change touching session lifecycle, IPC, auth, input, panes, paths and Win32 process handling. GitHub already reports it CONFLICTING and DIRTY. It is based on 05ade221, twelve commits behind master, and those twelve are the typing latency and startup readiness work that landed in `src/main.rs` and `src/session.rs` in the same new-session readiness region this rewrites with a net loss of 762 lines. Master now holds `timer_res::set_high(true)` across the readiness wait and the readiness poll backoff constants in exactly that area. Hand rebasing that is riskier than splitting.

Your own body says the full upstream suite was not run. For a change of this size touching session lifecycle, I cannot substitute targeted tests for that, and I would not ask you to run it on your own machine either. That is precisely why the pieces need to be small enough that I can run the suite on each one here.

## The part that needs its own argument, not a bundle

The reaper. `reap_orphaned_servers_in` becomes `fn reap_orphaned_servers_in(_psmux_dir: &Path) {}`, `select_orphan_pids` returns empty with every parameter underscored, and twelve assertions across three suites now pin the exact opposite of what they were written for:

- `test_issue448_orphan_reaper.rs`, four inverted, `assert_eq!(got, vec![1000], "untracked live server must be selected for reaping")` becomes `assert!(got.is_empty(), ...)`
- `test_session.rs`, six inverted, `assert!(!port_path.exists(), "invalid .port file should be removed")` becomes `assert!(port_path.exists(), ...)`
- `test_issue510_reaper_attribution.rs`, two inverted

To be clear, I do not think you hid this. Every inversion carries an honest rename, two files gained a header comment stating the policy change, no test was deleted, and your tradeoffs section discloses it. But the net effect is that all automatic cleanup of both stale registry files and orphaned processes is gone, and this repo leaks twenty to fifty live `__warm__` servers after a full test run. Making that permanent is a policy decision that needs to be argued on its own, in its own PR, where the argument is visible instead of being item nine in a list.

And here is my actual disagreement, which I think is the productive part: I do not believe you need to disable reaping to make it safe. Your own `terminate_server_pid_exact` is the fix. It opens with narrow rights, validates creation time, and terminates on the same retained handle so PID reuse cannot redirect the kill. That is exactly the primitive this repo has needed since it once managed to kill svchost and BSOD the machine. Reaping was unsafe because identity was image name matching. You replaced identity with PID plus creation time. So rebuild the reaper on your new primitive rather than deleting it. That gets your safety property and keeps the cleanup.

One more, `test_picker_namespace_filter.rs:37`: `session_namespace("dev____warm__")` flips from `None` to `Some("dev")` inside a test still named `warm_helpers_have_no_namespace`. The new behaviour may well be right, but an inverted assertion under an unchanged name is the one pattern I will always push back on, because it is indistinguishable from a regression slipping through. Please rename it.

## What I checked and found clean

Since you put this much work in, you should know what it survived. No dependency changes at all. AUTH genuinely tightens: the key is validated against CRLF and NUL injection before being sent, the ack is bounded and must be exactly `OK` rather than anything starting with it, and you deleted the debug line that printed the key. Process kill surface is strictly safer, one new TerminateProcess call site, gated on `verified_process_dead`. Three new unsafe blocks, all sound. Registry deletion is narrowed by a proper compare and swap. Path handling hex encodes anything outside `[A-Za-z0-9_-]` so `..` and separators cannot reach a filename. Roughly 25 `let _ = write_all(...)` became `write_all(...)?`, which is error surfacing, the opposite of swallowing. Your three audit scripts refuse to escape their target directory and clean up only what they created. No obfuscation anywhere.

## What I would like instead

In rough priority order, each rebased on current master:

1. The respawn refusal fix. I am landing this myself now with credit to you, so you can drop it.
2. `terminate_server_pid_exact` plus `process_identity`. Self contained, strictly safer than image name identity, and directly relevant to the BSOD history. I want this.
3. The client side AUTH tightening. About 15 lines, easy yes.
4. `registry::atomic_write`. Self contained, no caller behaviour change.
5. The paths.rs name encoding. Self contained and well tested, but it changes the on disk naming scheme, so its own PR and its own discussion.

Separately and each needing its own argument: the reaper policy (ideally as "reaper rebuilt on verified identity" rather than "reaper removed"), the pre auth 128 connection cap, and the bounded queue and writer rework, which overlaps what #542 is already proposing.

Three specific things to address whenever they land, all flagged as hazards rather than proven failures:

- `normalized_data_root` now calls `std::fs::canonicalize`, which fails when the data dir does not yet exist and falls back to the raw string. Two processes racing across directory creation could compute different root tags and therefore different mutex names, defeating the duplicate server guard. I have not proven this races in practice, but it wants a test.
- `acquire_session_mutex` now does `acquire_session_mutex_strict(name).ok().flatten()`, so an OS level `CreateMutexW` failure returns None, which callers read as "another owner holds it". Master documented that case as deliberately fail open so a legitimate start is never blocked. If that inversion is intentional, say so explicitly.
- `paths::validate_session_name` newly rejects names over 100 bytes and names containing `:`. Existing users with long session names would find them unusable after upgrade, so that needs a migration note.

I know splitting a change this size is genuinely annoying, and I would rather have your work than not. If it helps, send them one at a time in that order and I will review each promptly; you do not need to prepare all of them up front.

psmux:
Thank you for this. The volume of work is obvious and a real amount of it is good. I have read the whole diff rather than the description, built the branch, and run it, so what follows is specific.

I cannot take it as it stands, and the main reason is a reproduction rather than an opinion. Details below, along with the parts I do want and a proposal for how to land them.

## Credit first

The respawn failure you found is real and it was serious. The server was carrying a routine refusal out of the event loop on a bare `?`, so `run_server` returned `Err` and the process exited, destroying every window and pane in the session, while the client read no reply and exited 0. The session vanished and the caller was told it had succeeded.

That is already fixed on master in ca74224, and the commit message credits you by name. I also swept every other `?` in the event loop in 35654ad and pinned the exit status in 0c4f224. Finding that was worth the whole submission.

## The blocking problem: existing sessions become listed but unusable

`storage_base` in `src/paths.rs` only reproduces the current on disk spelling when the name passes `plain_component`, which permits `[A-Za-z0-9_-]` only, with no `__`, and no leading or trailing underscore. Every other name is rewritten to the `~psmux~` encoded form. Meanwhile `read_session_endpoint` in `src/session.rs` has no fallback to the legacy spelling, and all twenty `format!("{}__{}", ...)` call sites in master were replaced by `storage_base`.

I built your branch, created a session under an isolated data root, then renamed its files to the layout current master writes, leaving the same live server on the same port with the same key:

```
psmux list-sessions              ->  my.project: 1 windows (created ...)                        rc=0
psmux has-session  -t my.project ->  The system cannot find the file specified. (os error 2)    rc=1
psmux new-window   -t my.project ->  The system cannot find the file specified. (os error 2)    rc=1
psmux kill-session -t my.project ->  can't find session: my.project                             rc=1
```

So anyone upgrading with a session called `my.project`, `web app`, `proyecto`, `_scratch` or `build__dev` would see it sitting in `ls` and be unable to attach to it, and unable to kill it either. Not being able to clean it up is worse than losing it.

To be clear about what is and is not wrong here: the encoding itself is clean. `hex_name` emits only `[0-9a-f]`, the round trip is injective, `plain_component` refuses anything beginning with the `~psmux~` sentinel so a name cannot be forged into a collision, and `validate_registry_base` caps the result well under the 255 character component limit. I went looking for an injection path and there is not one. The defect is only that the read side has no compatibility fallback. If the encoding stays, the lookup needs to try the legacy spelling whenever the encoded base has no `.port`.

## The orphan reaper and the stale cleanup are turned off

`select_orphan_pids` now ignores all six of its arguments and returns an empty vector. `reap_orphaned_servers` has an empty body and `reap_orphaned_servers_in` is a no op, so `prune_stale_server_markers` is never called. `cleanup_stale_port_files_in_with` walks the directory, calls `probe`, then discards the result with `let _ =` and deletes nothing, which leaves the whole #530 machinery above it as dead code.

What made this hard to see is that the regression tests were not deleted, they were inverted. `orphan_with_no_registry_reference_is_reaped` became `..._is_preserved`, `wiped_registry_still_reaps_our_own_orphan` became `wiped_registry_does_not_authorize_killing_our_server`, `stale_cleanup_removes_invalid_port_and_key` became `..._preserves_invalid_registry_without_ownership`, and so on for nine tests across #448, #510 and #530. Each now asserts the opposite of what it was written to guard, so the suite stays green while the behaviour it pinned is gone.

I agree with your underlying principle. Missing or unreadable metadata is not proof that a process is disposable, and treating access denied as death is how you end up killing something you did not mean to. But the answer I want is to gate reaping on the stronger identity you built, not to stop reaping. Leaked servers accumulating is a problem I have hit on this machine, and dead entries that stay in `ls` forever after a reboot are their own bug. A change of that size to the reaping policy needs to be its own decision, argued in the open, not a side effect inside a reliability PR.

## Two smaller things

`open_authed` now half closes the write side after sending the command. I have a filtering network driver on this machine that drops one shot loopback replies once the write side is closed, and the symptom is empty output at exit 0 for every query. Master's `open_authed` does not do this, and the change would route every picker fetch, `session-info` and `list-tree` through that pattern. It needs to go, or be gated.

The control mode notification channel drops from 4096 to 128 while a full queue now disconnects the client rather than losing one message. I think that combination will fire during an ordinary output burst, and your own lifecycle script generates exactly that kind of flood.

## What I want to keep

Genuinely, several things.

The client side AUTH tightening is right and I checked it carefully. The server comparison is byte for byte what master already does, so the description slightly undersells where the change actually is, but the client half is a real improvement: the old reader would accept a payload from a peer that never acknowledged, and requiring an exact `OK`, plus a wall clock deadline, plus a size cap applied before the copy rather than after, is strictly better. No prefix comparison, no dispatch before auth. Good.

`process_identity`, `verified_process_dead` and `terminate_server_pid_exact` are the shape I want. Validating the creation time and then terminating through the same retained handle means there is no window between the check and the kill, and a recycled PID cannot redirect it. This project has a documented incident where psmux killed an unrelated system process, so that pattern is directly load bearing here. The `Alive` `Exited` `Unknown` tristate that stops treating access denied as death is the right model too.

`BoundedLines` closes a real unauthenticated memory growth path on the wire, with UTF-8 validation before dispatch and correct reassembly across read timeouts. `ConnectionPermit` is twenty lines and changes nothing until the server is under pressure. `atomic_write` in the registry, using a `create_new` temp plus a replacing rename, means a reader can never see a truncated registry file. Rejecting control characters and `:` in session names matches tmux and is right.

And the three audit scripts are careful work. Pinning the binary by SHA-256, isolating the data root, terminating only handles from your own `Process.Start` after validating creation time, and failing the run if the real registry or process list changed, is a better safety model than a good deal of my existing suite. I would take those regardless of what happens to the rest.

For completeness, since I was looking anyway: I checked every `unsafe` block in the diff and the preconditions hold in all of them, and neither PR touches `Cargo.toml`, `Cargo.lock` or any workflow file. Nothing here is doing anything it should not.

## What I would like to do

The branch forks 40 commits back and rewrites `main.rs`, `session.rs`, `server/mod.rs` and `input.rs` at a scale that will not rebase cleanly against the respawn fix, the #635 ingress work, the codepoint widths work and the #640 key table work that have all landed since. Rebasing it would be closer to re authoring it.

So rather than ask you to fight that, I would rather split it. Each of these would be a small PR I could review and take quickly:

1. The client AUTH tightening, without the half close.
2. `process_identity` and `terminate_server_pid_exact` as helpers, leaving the reaper in place.
3. `BoundedLines`.
4. `ConnectionPermit` plus the accept loop rejection.
5. `atomic_write`.
6. The three audit scripts.

The registry name encoding, and the discovery and reaping policy, are the two that need a design conversation first. I would rather have that on an issue where it can be argued properly than inside a five thousand line diff. Open one and I will engage with it seriously.

Leaving this open for now. Thank you again for the depth here, and particularly for the respawn find.

Separately, #636 is merged as 4a508df with your authorship kept.

Issue #504, opened 2026-07-29
Ctrl+Space prefix doesn't work in Alacritty


abdullah-albanna:
upon searching, I found that this a known issue of Alacritty in Windows.

And the proposed solution from [here](https://github.com/alacritty/alacritty/issues/7932) and [here](https://github.com/alacritty/alacritty/issues/1703) is to do

```toml
[keyboard]
bindings = [
  { key = "Space", mods = "Control", chars = "\u0000"}
]
```
but it still doesn't work

alacritty 0.16.0 (1208240)

psmux:
Thanks for the report, and especially for digging up the Alacritty issues first. That saved me a lot of time, and it turned out you were right on both counts: there really are two separate problems here, and one of them was mine.

I did not read any psmux code until I had measured what actually arrives. I wrote a console key dumper, ran it inside Alacritty on Windows, and drove it with real hardware keystrokes.

## Finding 1: Alacritty's native Ctrl+Space is genuinely undetectable

Here is what the console receives for Ctrl+Space:

```
alacritty   Ctrl+Space  ->  KEY down=1 vk=0x20 uchar=0x0020 ctrl=0x0
alacritty   plain Space ->  KEY down=1 vk=0x20 uchar=0x0020 ctrl=0x0
conhost     Ctrl+Space  ->  KEY down=1 vk=0x20 uchar=0x0020 ctrl=0x8
```

Alacritty's Ctrl+Space is byte for byte identical to an unmodified Space, with no Ctrl flag at all. At the raw PTY level it writes 0x20, the same byte as Space. Ctrl+B from the same Alacritty window correctly arrives as 0x02, so it is specific to this key.

There is nothing psmux can do about that one. If I made a bare Space arm the prefix, I would break the space bar. This is the upstream Alacritty bug you found, and conhost proves the same physical key is reported correctly elsewhere.

## Finding 2: your workaround was correct, and psmux was silently dropping it

This is the part that was my fault. When you configure Alacritty to send a literal NUL, that NUL **does** reach the console. ConPTY encodes it like this:

```
alacritty + NUL workaround -> KEY down=1 vk=0x32 uchar=0x0000 ctrl=0x18
```

[...]

Issue #396, opened 2026-06-16
Shift+Enter submits instead of inserting a newline in TUI apps over SSH (ESC+CR mis-decoded by the VT input parser)


sandersaares:
## Summary

When psmux runs over SSH on Windows, pressing **Shift+Enter** inside a TUI app (for example the GitHub Copilot CLI or Claude Code) submits the prompt instead of inserting a newline. Windows Terminal encodes Shift+Enter as the two bytes `ESC` + `CR` (`\x1b\r`), and psmux's SSH VT input parser mis-decodes that as a standalone `Escape` followed by an unmodified `Enter`. The inner app therefore receives a plain Enter and submits.

## Environment

- psmux: 3.3.6 (reproduces on `master` as well)
- Host OS: Windows, running the psmux session server under the Windows OpenSSH server (sshd ConPTY)
- Client: Windows Terminal connected over SSH
- Affected apps: any TUI that uses `ESC`+`CR` (`\x1b\r`) to mean "insert newline" — verified with the GitHub Copilot CLI; Claude Code behaves the same way

## Steps to reproduce

1. From Windows Terminal, SSH into a Windows host and attach a psmux session: `psmux new-session -A -s main`.
2. Inside that session, run a TUI app whose prompt treats Shift+Enter as "insert a newline" (e.g. the GitHub Copilot CLI).
3. Press **Shift+Enter**.

## Expected

A newline is inserted at the cursor and the prompt is **not** submitted.

## Actual

The prompt is submitted as if plain `Enter` was pressed (sometimes with a stray `Escape` side effect from the leading ESC).

## Root cause

Windows Terminal sends Shift+Enter as `ESC`+`CR` (`\x1b\r`). In `src/ssh_input.rs`, `on_escape()` only maps `ESC` followed by a *printable* character (`' '..='~'`) to `Alt`+char. `ESC` followed by a C0 control byte such as `CR` (`0x0D`) falls through to the catch-all arm, which emits a standalone `Esc` and then re-processes the `CR` via `on_ground()` as an unmodified `Enter`:

```rust
// src/ssh_input.rs — on_escape()
c if c >= ' ' && c <= '~' => {
    // Alt + printable character.
    emit(make_key(KeyCode::Char(c), KeyModifiers::ALT));
    self.state = PS::Ground;
}
c => {
    // Unknown after Esc — emit Esc then re-process char.
    emit(make_key(KeyCode::Esc, KeyModifiers::empty()));
    self.state = PS::Ground;
[...]

PR #643, opened 2026-09-10
fix(#642): the VT input path reads 0x0a as C-j, not as a plain Enter

### LINKED ISSUE #508: C-Space prefix broken under WezTerm on Windows (VT path)
## Summary

Under WezTerm on Windows, a `C-Space` prefix (leader) never arms, so the prefix is unusable. The same config works under Windows Terminal, and works under WezTerm if psmux is forced onto the native input path. `Ctrl`+letter keys work under WezTerm; only `Ctrl`+non-letter keys like `Ctrl+Space` are affected.

This is adjacent to #504 (Alacritty). psmux `9369e6d` already includes the `b6ca300` NUL→C-Space fold from that issue, but WezTerm on the VT path is still affected.

## Environment

- Windows 11
- WezTerm (Windows, ConPTY)
- psmux 3.3.7 (`9369e6d`) — includes #504 / `b6ca300`
- prefix: `set -g prefix C-Space` + `bind C-Space send-prefix`

## Symptom

- `C-Space` then e.g. `c` does nothing under WezTerm.
- The same config under **Windows Terminal** works.
- `Ctrl`+letters (e.g. a `vim-tmux-navigator`-style `C-h/j/k/l`) work under WezTerm; only `Ctrl+Space` (and presumably other `Ctrl`+non-letter) fail.

## What WezTerm sends

Per wezterm/wezterm#4055 (Windows), WezTerm reports `Ctrl+Space` as:

```
KeyEvent { key: Char(' '), modifiers: CTRL, ..., win32_uni_char: Some(' ') }
```

i.e. `VK_SPACE` + CTRL, but the unicode char is `0x20` (space), not NUL (`0x00`). The CTRL bit is present (conhost reports the same `0x20` char with the CTRL flag set).

## Path dependence

`needs_vt_input()` routes WezTerm (detected via `TERM_PROGRAM=WezTerm` / `WEZTERM_PANE`) to the VT input path. On that path the `C-Space` prefix is dropped. Clearing those two env vars before launching psmux (so `needs_vt_input()` is false → native `INPUT_RECORD` path) makes `C-Space` work — so the key reaches psmux, it just isn't folded to `C-Space` on the VT path. Windows Terminal, which stays on the native path, is unaffected.

## Not a workaround: `allow_win32_input_mode = false`

Setting WezTerm's `allow_win32_input_mode = false` does **not** help: the prefix is still dead, and it additionally breaks VT mouse. So legacy encoding is not a viable workaround.

## Ask

Could the VT input path fold WezTerm's `Ctrl+Space` (`VK_SPACE` + CTRL, `win32_uni_char 0x20`) onto `C-Space`, mirroring the `b6ca300` NUL fold from #504? I did not want to over-assert the internal mechanism — happy to run a console key-dumper under WezTerm and paste exactly what arrives if that helps.

## Current workaround

Launch psmux with `WEZTERM_PANE`/`TERM_PROGRAM` cleared so it uses the native input path (trade-off: loses psmux-level VT mouse).

### LINKED ISSUE #642: C-j prefix broken under WezTerm on Windows (VT path): 0x0a decoded as Enter, not C-j
## Summary

Under WezTerm on Windows, a `C-j` prefix never arms, so the prefix is unusable. The same config works under Windows Terminal. `C-b`, `C-Space` and `C-a` all work fine under WezTerm — only `C-j` is affected.

The VT input path decodes the byte `0x0a` as a plain `Enter` instead of `C-j`. `Ctrl`+letter and `Enter` are the same wire byte for exactly one letter — `Ctrl+J` *is* LF — and the VT path resolves that collision the opposite way from tmux, from psmux's own native input path, and from psmux's own output encoder.

This is adjacent to #508 and #504, but the `vk_nul_to_ctrl_space` fold that closed #508 does not cover it: the record here arrives intact and carries a character, it is just decoded as the wrong key.

## Environment

- Windows 11 26200
- WezTerm (Windows, ConPTY)
- psmux 3.3.8 (`66cf613`, winget release — unmodified)
- tmux 3.6a under WSL, for comparison

## Symptom

Tested with a minimal config, changing only the prefix, then pressing `prefix` + `c` to split:

| prefix | result under WezTerm |
|---|---|
| `C-b` (default, no config) | OK |
| `C-Space` | OK |
| `C-a` | OK |
| **`C-j`** | **prefix never arms** |

## What the trace shows

`PSMUX_SSH_DEBUG=1`, `~/.psmux/ssh_input.log`. The working and broken cases produce records of **identical shape** — the only difference is which C0 byte arrives.

`C-b` prefix, pressing `Ctrl+B`:

```text
KEY vk=0x0011 scan=0x001D u_char=0x0000 ctrl=0x00000008
KEY vk=0x0000 scan=0x0000 u_char=0x0002 ctrl=0x00000000
  → emit(char): Key(KeyEvent { code: Char('b'), modifiers: KeyModifiers(CONTROL), kind: Press, .. })
```

`C-j` prefix, pressing `Ctrl+J`:

```text
KEY vk=0x0011 scan=0x001D u_char=0x0000 ctrl=0x00000008
KEY vk=0x0000 scan=0x0000 u_char=0x000A ctrl=0x00000000
  → emit(char): Key(KeyEvent { code: Enter, modifiers: KeyModifiers(0x0), kind: Press, .. })
```

WezTerm and ConPTY both deliver the byte correctly. The key is lost inside psmux's VT parser, which reports it as an unmodified `Enter`, so no prefix match is possible.

## tmux comparison (measured, not assumed)

tmux 3.6a under WSL, same physical keyboard:

- `set -g prefix C-b` plus `bind -n C-j display-message "C-j reached tmux"` — pressing `Ctrl+J` shows the message, so tmux decodes `0x0a` as `C-j`.
- `set -g prefix C-j` — `Ctrl+J` then `c` creates a new window. The root-table message correctly does *not* fire, because the prefix takes precedence.

So `C-j` is a usable prefix in tmux.

## Path dependence

`needs_vt_input()` routes WezTerm (via `TERM_PROGRAM=WezTerm` / `WEZTERM_PANE`) to the VT input path. Windows Terminal stays on the native `INPUT_RECORD` path, where the same key works: conhost delivers `vk=VK_J` with `u_char=0x0a`, and that resolves to `Char('j')` + CONTROL. So psmux's two input paths disagree about the same keystroke.

## psmux's own output side also disagrees

- `send-key C-j` writes `0x0a` to the pane (`ctrl_char_send_keys_byte('j')`).
- `KeyCode::Enter` writes `0x0d`.

psmux therefore already treats `0x0a` and `0x0d` as different keys everywhere except this one decoder.

## Suspected cause

Offered separately from the measurements above, and not asserted — `src/ssh_input.rs`, `VtParser::on_ground`:

```rust
'\r' | '\n' => emit(make_key(KeyCode::Enter, KeyModifiers::empty())),
```

`0x0a` is consumed here before it can reach the `Ctrl+A..Ctrl+Z` arm a few lines below, which is what turns `0x02` into `C-b`. `git log -S` puts this line in the module's original commit (`44b81c8`, mouse over SSH), so it does not look like it was written to satisfy a reported case.

The neighbouring arms are not in question: `'\t'` → `Tab` and `'\r'` → `Enter` match tmux, where `C-i` and `Tab`, and `C-m` and `Enter`, are genuinely the same key. `0x0a` is the odd one out — it has no named key of its own, and tmux calls it `C-j`.

## One honest complication

`Ctrl+J` and `Ctrl+Enter` are indistinguishable on the wire: Windows Terminal sends `0x0a` for `Ctrl+Enter` too, which psmux itself relies on in the output direction (#409). So `0x0a` can only ever mean one of them, and today's mapping does not make `Ctrl+Enter` work either — it arrives as a plain `Enter` with the modifier already gone. Choosing `C-j` would match tmux, which has the same ambiguity and resolves it the same way.

## Ask

Could the VT input path decode `0x0a` as `C-j` rather than `Enter`, matching tmux, the native input path, and psmux's own `send-key` encoder? Happy to test a patch under WezTerm and paste the trace.

tanaeakihiko:
Fixes #642.

## What

`on_ground` in the VT input parser read `'\r' | '\n' => Enter`, so the byte `0x0a` was reported as an unmodified `Enter` and never reached the `Ctrl+A..Ctrl+Z` arm a few lines below. A `C-j` prefix was therefore dead on every client where `needs_vt_input()` is true — WezTerm, the JetBrains terminals, and everything over SSH — while the same config worked under Windows Terminal.

The change is one line:

```diff
-'\r' | '\n' => emit(make_key(KeyCode::Enter, KeyModifiers::empty())),
+'\r' => emit(make_key(KeyCode::Enter, KeyModifiers::empty())),
```

`0x0a` now falls through to the `Ctrl+A..Ctrl+Z` arm and becomes `C-j`. No existing test changes.

## Why `0x0a` is `C-j`

`Ctrl`+letter puts the letter's low five bits on the wire, so `Ctrl+J` **is** LF, the way `Ctrl+I` is Tab and `Ctrl+M` is CR. Those two collisions have a named key on the other side and tmux treats each pair as one key, which is why the `'\t'` and `'\r'` arms are correct and are untouched. `0x0a` has no such name — `Enter` is `0x0d` — so the only key it can be is `C-j`.

Three independent things already agree on that, and only this decoder disagreed:

| | `0x0a` |
|---|---|
| tmux 3.6a (WSL, measured) | `C-j` |
| psmux native input path (Windows Terminal) | `Char('j')` + CONTROL |
| psmux output encoder (`send-key C-j`) | writes `0x0a`; `Enter` writes `0x0d` |
| **psmux VT input path** | **`Enter`** |

## Measurements

Before, against the shipped 3.3.8 release (`66cf613`) under WezTerm on Windows 11 26200, changing only the prefix between runs:

| prefix | `prefix` + `c` |
|---|---|
| `C-b` | splits |
| `C-Space` | splits |
| `C-a` | splits |
| `C-j` | nothing — the prefix never arms |

`PSMUX_SSH_DEBUG=1` — the working and broken cases produce records of identical shape; only the C0 byte differs:

```text
Ctrl+B: KEY vk=0x0000 u_char=0x0002 ctrl=0x00000000
          → emit(char): Key(code: Char('b'), modifiers: CONTROL)
Ctrl+J: KEY vk=0x0000 u_char=0x000A ctrl=0x00000000
          → emit(char): Key(code: Enter,     modifiers: 0x0)
```

tmux 3.6a under WSL, same keyboard: with `prefix C-b` and `bind -n C-j display-message`, `Ctrl+J` fires the message; with `prefix C-j`, `Ctrl+J` then `c` creates a window and the root-table message correctly does not fire, because the prefix takes precedence.

After, WezTerm with `prefix C-j`:

- `prefix` + `c` splits.
- A byte logger in the pane shows `0a` for `Ctrl+J`, `0d` for `Enter`, and `1b 0d` for a WezTerm `SendString '\x1b\r'` binding — so **#396's Shift+Enter path is intact**.
- Other `Ctrl`+letters, `Enter` and mouse are unchanged.

## What is deliberately *not* changed

`on_escape` keeps `'\r' | '\n' => Alt+Enter`, so ESC+LF is still `Alt+Enter` — asymmetric with the bare byte, on purpose.

Decoding ESC+LF as `M-C-j` looked like the consistent thing to do, and a first cut of this fix did exactly that. Measuring it end to end ruled it out: a named `C-M-<letter>` never reaches `encode_key_event`, because `write_named_key_to_pane` injects a console record first and ConPTY drops the Alt from it.

| ESC+LF (`1b 0a`) in | bytes the pane's child receives |
|---|---|
| today | `1b 0d` — the LF is rewritten as a CR |
| as `M-C-j` | `0a` — the ESC is dropped |

Neither is what the terminal sent, so ESC+LF wants a fix in the **output** path, which is a different problem from #642. The reasoning and the measurement are recorded in the code and pinned by `esc_lf_is_left_as_alt_enter`, so the next reader does not repeat the attempt.

## The ambiguity that remains

`Ctrl+J` and `Ctrl+Enter` are the same byte — Windows Terminal sends `0x0a` for `Ctrl+Enter` too, which psmux relies on in the output direction (#409). `0x0a` can only ever mean one of them. It did not mean `Ctrl+Enter` before this change either: it arrived as a plain `Enter` with the modifier already gone.

## Relation to #508 / #504

Same class — a prefix key lost on the VT path, working on the native path — but a different mechanism. The `vk_nul_to_ctrl_space` fold that closed #508 catches a record that arrives with no character; here the record arrives intact and carries its byte, and is simply decoded as the wrong key.

## Tests

11 new tests in `tests-rs/test_issue642_vt_ctrl_j.rs`: the decoded key, that it matches a `C-j` prefix parsed from config, that LF and CR are no longer the same event, that `'\t'` and `'\r'` keep their named keys, that the other `Ctrl`+letters are unchanged, the byte round-trip, and guards on ESC+CR and ESC+LF. Full suite green (3560), `cargo clippy` clean on the change.

### DIFF
--- a/src/ssh_input.rs
+++ b/src/ssh_input.rs
@@ -1104,7 +1104,22 @@ impl VtParser {
             '\x1b' => {
                 self.state = PS::Escape;
             }
-            '\r' | '\n' => emit(make_key(KeyCode::Enter, KeyModifiers::empty())),
+            '\r' => emit(make_key(KeyCode::Enter, KeyModifiers::empty())),
+            // NOTE: 0x0a deliberately has NO arm here.  It falls through to the
+            // Ctrl+A..Ctrl+Z arm below, which turns it into C-j (issue #642).
+            //
+            // Ctrl+<letter> puts the letter's low five bits on the wire, so
+            // Ctrl+J IS the byte LF, the way Ctrl+I is Tab and Ctrl+M is CR.
+            // Those two collisions have a named key on the other side and tmux
+            // treats each pair as one key, which is why `'\t'` and `'\r'` keep
+            // their arms.  0x0a has no such name: Enter is 0x0d, so the only
+            // key 0x0a can be is C-j, and that is what tmux calls it (measured
+            // against tmux 3.6a: `bind -n C-j` fires on the byte, and a `C-j`
+            // prefix arms).  Sharing an arm with `'\r'` reported it as an
+            // unmodified Enter, so a `C-j` prefix was dead on every client
+            // where needs_vt_input() is true — WezTerm, the JetBrains
+            // terminals, and everything over SSH — while it worked under
+            // Windows Terminal, whose records take the native path.
             '\t' => emit(make_key(KeyCode::Tab, KeyModifiers::empty())),
             '\x7f' => emit(make_key(KeyCode::Backspace, KeyModifiers::empty())),
             // NOTE: 0x08 deliberately has NO arm here.  It falls through to the
@@ -1158,7 +1173,18 @@ impl VtParser {
                 // Windows Terminal sends ESC+CR for Shift+Enter; forwarding one
                 // \x1b\r (re-emitted by encode_key_event) lets TUI apps such as
                 // the Copilot and Claude CLIs insert a newline instead of
-                // submitting the prompt.
+                // submitting the prompt (issue #396).
+                //
+                // ESC+LF stays paired with ESC+CR here even though a BARE 0x0a
+                // is now C-j (see `on_ground`).  The asymmetry is deliberate:
+                // measured end to end, `\x1b\n` reaches a pane as `\x1b\r`
+                // today and would reach it as `0a` if this were decoded as
+                // M-C-j, because a named `C-M-<letter>` does not go through
+                // `encode_key_event` at all — `write_named_key_to_pane` injects
+                // a console record first and ConPTY drops the Alt from it.
+                // Neither is the `\x1b\n` the terminal sent, so making this
+                // right is a separate problem in the OUTPUT path and does not
+                // belong to #642.
                 emit(make_key(KeyCode::Enter, KeyModifiers::ALT));
                 self.state = PS::Ground;
             }
@@ -2239,6 +2265,10 @@ pub(crate) mod test_support {
 #[cfg(test)]
 #[path = "../tests-rs/test_ssh_vt_paste.rs"]
 mod tests;
+
+#[cfg(test)]
+#[path = "../tests-rs/test_issue642_vt_ctrl_j.rs"]
+mod tests_issue642_vt_ctrl_j;
 
 #[cfg(test)]
 #[path = "../tests-rs/test_issue457_ssh_mouse_build_gate.rs"]
--- a/tests-rs/test_issue642_vt_ctrl_j.rs
+++ b/tests-rs/test_issue642_vt_ctrl_j.rs
@@ -1,0 +1,217 @@
+//! Issue #642: a `C-j` prefix is dead on the VT input path.
+//!
+//! Measured first, against the shipped 3.3.8 release (`66cf613`) under WezTerm
+//! on Windows 11 26200, with only the prefix changed between runs:
+//!
+//! ```text
+//!   prefix C-b       prefix + c splits        OK
+//!   prefix C-Space   prefix + c splits        OK
+//!   prefix C-a       prefix + c splits        OK
+//!   prefix C-j       nothing happens          the prefix never arms
+//! ```
+//!
+//! `PSMUX_SSH_DEBUG=1` shows the working and broken cases producing records of
+//! identical shape — the only difference is which C0 byte arrives:
+//!
+//! ```text
+//!   Ctrl+B: KEY vk=0x0000 u_char=0x0002 ctrl=0x00000000
+//!             -> emit(char): Key(code: Char('b'), modifiers: CONTROL)
+//!   Ctrl+J: KEY vk=0x0000 u_char=0x000A ctrl=0x00000000
+//!             -> emit(char): Key(code: Enter,     modifiers: 0x0)
+//! ```
+//!
+//! WezTerm and ConPTY both deliver the byte. It was psmux that lost the key:
+//! `on_ground` read `'\r' | '\n' => Enter`, so 0x0a never reached the
+//! `Ctrl+A..Ctrl+Z` arm that turns 0x02 into `C-b`.
+//!
+//! ## Why 0x0a is C-j and not Enter
+//!
+//! Ctrl+<letter> puts the letter's low five bits on the wire, so Ctrl+J IS LF,
+//! the way Ctrl+I is Tab and Ctrl+M is CR. Those two collisions have a named
+//! key on the other side, and tmux treats each pair as one key — which is why
+//! the `'\t'` and `'\r'` arms are correct and are left alone here. 0x0a has no
+//! such name. Enter is 0x0d, so the only key 0x0a can be is C-j.
+//!
+//! Measured against tmux 3.6a under WSL, same keyboard: with `prefix C-b` and
+//! `bind -n C-j display-message`, pressing Ctrl+J fires the message; with
+//! `prefix C-j`, Ctrl+J then `c` creates a window and the root-table message
+//! correctly does not fire, because the prefix takes precedence. psmux's own
+//! native input path agrees (conhost reports `VK_J` + CONTROL, which is why the
+//! same prefix works under Windows Terminal), and so does psmux's own output
+//! encoder: `send-key C-j` writes 0x0a and `Enter` writes 0x0d.
+//!
+//! ## The one thing that cannot be fixed
+//!
+//! Ctrl+J and Ctrl+Enter are the same byte — Windows Terminal sends 0x0a for
+//! Ctrl+Enter too, which psmux relies on in the output direction (#409). So
+//! 0x0a can only ever mean one of them. It did not mean Ctrl+Enter before this
+//! change either: it arrived as a plain Enter with the modifier already gone.
+//!
+//! ## Scope
+//!
+//! Only the BARE byte moves. `on_escape` keeps `'\r' | '\n' => Alt+Enter`; see
+//! `esc_lf_is_left_as_alt_enter` for the measurement that ruled out changing it
+//! here.
+//!
+//! Verified after the change, WezTerm on the same machine, prefix `C-j`:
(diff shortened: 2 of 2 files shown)


psmux:
Merged, with your authorship intact, in 84728a2 on master (merge a8d694c). Thank you. This is exactly the kind of PR I like to receive: one line of behaviour, the reasoning written where the next reader will find it, and a measurement table instead of an argument.

I did not take it on trust. Before merging I rebuilt unmodified master from the worktree and reproduced it two ways, with your two line config (`set -g prefix C-j`, `bind c split-window`):

- A real WezTerm window, with the client's `WEZTERM_UNIX_SOCKET` recovered from the launcher so `wezterm cli send-text --no-paste` could put the raw byte `0x0a` and then `c` on the wire. Three runs on master: panes 1 to 1 every time. Three runs on your branch: 1 to 2 every time.
- A headless client on the VT path fed the exact record from your trace (`vk=0x0000 u_char=0x000A`) through `WriteConsoleInputW`. Same 0 of 3 before, 3 of 3 after.

The `ssh_input.log` lines match yours byte for byte on master (`code: Enter, modifiers: KeyModifiers(0x0)`) and read `Char('j'), modifiers: KeyModifiers(CONTROL)` after.

On the tmux side, `tty-keys.c:939` folds every C0 byte without a dedicated key into a Ctrl key and excludes exactly `C0_HT`, `C0_CR` and `C0_ESC`, so `0x0a | 0x40` is `J`, lowercased to `j | KEYC_CTRL`. And `key-string.c:79` names `C0_LF` only as `[LF]` while `C0_CR` gets `Enter`, which is tmux saying plainly that LF is not Enter. `input-keys.c:524` takes `j & 0x1f` back to `0x0a` on the way out, the same map `ctrl_char_send_keys_byte` ports, so your round trip claim holds.

One thing I added in ac45fcf. Every payload in the existing bracketed paste tests used `\r`, never `\n`, so nothing in the tree would have caught a later edit routing paste bytes through `on_ground` and turning every newline in a pasted block into a `C-j` keypress. The paste states never call the ground decoder (`on_paste` just pushes the char or watches for ESC), so it is structurally unreachable today, and two new guards now pin that with bare LF and CRLF payloads. There is also an end to end suite, `tests/test_issue642_ctrl_j_prefix.ps1`, that hosts a VT path client inside a ConPTY, drives a real WezTerm window when one is installed, and pins the native `INPUT_RECORD` path as unchanged. It scores 12 of 12 on the merged build and 8 of 12 with your one line reverted, the four failures being exactly the `C-j` checks.

Your reading of the ESC+LF asymmetry and the `Ctrl+Enter` collision both checked out, and I agree the output path question is a separate issue. Full suite is 3605 per binary, zero failures.

To pick it up before the next release:

```
cargo install --git https://github.com/psmux/psmux psmux
```

Issue #602, opened 2026-08-26
move-window ignores -s and swap/move relative targets are parsed as absolute/vector positions


idan-h:
## Summary

psmux has inconsistent window-target handling across the external CLI, attached-client/control path, and the server handlers.

Two live problems share this resolver gap:

1. `move-window -s SOURCE -t TARGET` ignores `-s` and always moves the active window.
2. `swap-window -t +N` / `-t -N` do not use tmux-relative semantics. `+N` is parsed as an absolute unsigned index, while a missing display index is incorrectly reinterpreted as a raw vector position.

All affected commands return exit code 0 without an error.

## Environment

- psmux 3.3.7 (`05cc5d4`, 2026-07-20)
- Windows 11 Pro 10.0.26200
- PowerShell 7+

## Reproduction: ignored move source

Before:

`0:lazygit 1:main* 2:claude`

Run:

`psmux move-window -s SESSION:2 -t SESSION:9`

Expected: `2:claude` becomes `9:claude`; `main` remains at 1.

Actual: the active `main` window moves from 1 to 9.

## Reproduction: relative target

Before, current window is 11:

`0:main 2:lazygit 9:claude 10:d 11:c* 12:b 13:a`

Run:

`swap-window -t +1`

Expected: swap 11 and 12.

Actual: swap display indexes 11 and 2.

`+2` produces the same target. `-1` is not resolved as the previous window.

## Root cause

`MoveWindow` carries only a destination and the handler explicitly operates on the active window:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/types.rs#L1528-L1530
- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/connection.rs#L2262-L2272
- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/mod.rs#L3922-L3933

The external CLI sends window targets through unsigned parsing without a relative-target branch:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/main.rs#L263-L279

The swap handler resolves a display index with `app.win_pos(target).unwrap_or(target)`:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/mod.rs#L3936-L3944

[...]

Issue #601, opened 2026-08-26
move-window/swap-window do not invalidate attached-client status; swap-window loses active window identity


idan-h:
## Summary

On psmux 3.3.7 (`05cc5d4`), successful `move-window` and `swap-window` mutations do not schedule a fresh frame or rebuild the cached window metadata. The server state changes, but an attached client's status bar keeps showing the old window list until `refresh-client -S`.

The same `swap-window` handler also leaves `active_idx` and `last_window_idx` at their old vector positions. If the active window is swapped, the active marker changes to whichever window entity lands at that position instead of following the originally active window as tmux does.

Both symptoms are post-mutation bookkeeping omissions in the same handlers.

## Environment

- psmux 3.3.7 (`05cc5d4`, 2026-07-20)
- Windows 11 Pro 10.0.26200
- PowerShell 7+

## Reproduction

1. Create a session with windows `0:main*`, `1:lazygit`, `2:claude`.
2. Attach a client.
3. From another shell, run `psmux swap-window -t SESSION:2`.
4. Run `psmux list-windows -t SESSION`.

The server-side list shows the swap, but the attached status bar remains unchanged. `psmux refresh-client -S -t SESSION` makes it update.

When the active window participates in the swap, the active index stays at the old vector position and points to the other window entity after the swap.

## Source

The mutation handlers change the window state but set neither `state_dirty` nor `meta_dirty`:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/mod.rs#L3922-L3945

The server pushes frames only when one of those flags is set, and rebuilds cached window metadata only for `meta_dirty`:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/mod.rs#L5392-L5408

The dump-state fast path returns `NC` while `state_dirty` is false:

- https://github.com/psmux/psmux/blob/05cc5d4cded047bd3f3d1955299fd0bd259f2d81/src/server/mod.rs#L1946-L1968

The swap itself is only `app.windows.swap(spos, tpos)`; it does not re-resolve the active or last window by stable window ID.

The same omissions remain on master `60e59fb`:

- https://github.com/psmux/psmux/blob/60e59fb353a9107edda9a4f1af65efad75e67f10/src/server/mod.rs#L4313-L4355

## Expected behavior

- A successful move/swap immediately updates every attached client's window list/status.
[...]

PR #652, opened 2026-09-12
fix(#651): bind prefix + . to a move-window prompt, and stop command-prompt leaking a flag's value

### LINKED ISSUE #651: No default binding for `prefix + .` (move-window), and `command-prompt` drops a value-taking flag's value
tmux has bound `.` by default for a long time. Measured with `list-keys -T prefix` on tmux 3.6a:

```
bind-key -T prefix . command-prompt -T target { move-window -t "%%" }
```

`prefix` `.` prompts for an index and moves the current window there. In psmux the key does nothing at all: no prompt, no message, and `list-keys` shows no binding for it.

```
$ psmux -L t list-keys | grep -E "prefix (\.|'|:) "
bind-key -T prefix ' select-window-index
bind-key -T prefix : command-prompt
```

`.` is simply absent from `PREFIX_DEFAULTS` (src/help.rs). From inside a session there is no way to renumber a window — you have to leave it and run `psmux move-window -t N` from another shell.

## Not just a missing line: the binding cannot work as written today

psmux has no `{}` command blocks (only the `if-shell 'cond' { ... }` config form), so the natural way to write tmux's command for psmux's parser is

```
command-prompt -T target "move-window -t '%%'"
```

which `parse_command_line` reduces to the same argv. Adding that does not fix the key, because the client's `command-prompt` argument parser consumes a value only for `-I` and `-p`, treats `-1`/`-N`/`-W` as booleans, and skips **the flag but not its value** for everything else (src/client.rs):

```rust
} else if tokens[i].starts_with('-') {
    i += 1; // skip unknown flags
} else {
    positional.push(tokens[i].clone());   // the -T value lands here
    i += 1;
}
```

So with `-T target "move-window -t '%%'"` the positional list becomes `["target", "move-window -t '%%'"]` and the template is

```
target move-window -t '%%'
```

Typing `3` and pressing Enter sends `target move-window -t '3'`, which is not a command. `-t` (target client) sits in the same hole.

The flag's shape is already recorded correctly — `ARGS_TEMPLATES` in src/cli.rs carries `command-prompt`'s template `1CbeFiklI:NPp:t:T:`, transcribed from tmux's `cmd_entry`, and one validator reads it (#635). The parser just never asked the table. No default binding uses `-T` or `-t` today, which is presumably why this has gone unnoticed.

## Could a user just bind it themselves? Measured: not with tmux's own line

This is the part worth being precise about, because `bind-key` **accepts** the binding and `list-keys` shows it — it fails only when the key is pressed, and the same parser serves user bindings and defaults alike (the dispatcher takes the command string out of the synced binding table without caring where it came from).

Measured on master (`ddd38bc`) by nesting psmux: the outer server hosts one pane whose child is a second psmux, attached, so `send-keys` into that pane is real key input for the inner client, and the inner session's window list read from outside is the witness. Inner config is the one line, `PSMUX_CONFIG_FILE` only.

| `.tmux.conf` line | `list-keys -T prefix` | prefix + `.` , then `5`, then Enter |
|---|---|---|
| `bind . command-prompt -T target "move-window -t '%%'"` | `bind-key -T prefix . command-prompt -T target "move-window -t '%%'"` | prompt opens (headed `command`); windows stay `0 1 2*` — nothing moves |
| `bind . command-prompt -p index "move-window -t '%%'"` | shown likewise | works: windows become `0 1 5*` |

So the workaround is to write the binding without `-T`. With `-T` — which is what tmux itself uses, and what anyone copying tmux's default would write — the key silently does nothing, and the reason is invisible: the binding is listed, the prompt appears, the typed index is accepted, and then the command that goes to the server is `target move-window -t '5'`.

## Relation to #601 and #602 (this is a different layer)

Worth saying explicitly, since both of those issues are still open: `move-window` **itself** is fine. The target resolver and the state invalidation landed in master as

* `f497d7b` fix(#602): resolve move-window/swap-window targets through one tmux-parity resolver
* `8986564` fix(#601): mark the state dirty so an attached client sees a move or a swap

and both are ancestors of current master. Measured on master (`ddd38bc`) in an isolated namespace:

| command | result | rc |
|---|---|---|
| windows `1 2 3*`, `move-window -t 5` | `1 2 5*` | 0 |
| `move-window -t 0` (free) | `0* 1 2` | 0 |
| `move-window -t 1` (occupied) | `ERROR: index in use: 1`, nothing moved | 1 |
| `move-window -t +3` (from 0) | `1 2 3*` — index arithmetic, tmux's `CMD_FIND_WINDOW_INDEX` | 0 |

This report is not a regression of those, and does not need any change to `move-window`: it is about the two things between the keyboard and that command — the missing key binding, and the flag parse that binding depends on. The fix I have builds *on* #601/#602: because `resolve_window_spec` is already complete, `+1`, `-1`, `$` and `{end}` typed at the prompt work, and an occupied index is refused rather than silently swapped.

(#601 and #602 may just be awaiting closing.)

## Environment

* psmux 3.3.8, built from `ddd38bc`
* Windows 11 Pro 26200, Intel Core Ultra 5 225U
* tmux side measured on 3.6a under WSL2 (Ubuntu), isolated `-L` sockets
* Nesting note, in case you reproduce it the same way: the inner psmux needs `PSMUX_SESSION` cleared ("sessions should be nested with care"), and a binary whose basename is not `psmux.exe` does not bring up a server at all, which cost me a while.
* Terminal-independent: the binding does not exist in any of psmux's input paths, which `list-keys` shows without pressing a key.

## What I would expect

`prefix` `.` opens a prompt and moves the current window to the index typed, matching tmux. `list-keys` would show the same command, in psmux's quoting rather than tmux's `{}` block.

I have a patch for both halves (the flag parse through `cli`'s existing flag table, plus the default binding) with unit tests, and can open a PR.

tanaeakihiko:
Fixes #651.

Fixes the two things that sit between the keyboard and `move-window`: tmux's default binding for `.` was missing, and the binding could not have worked anyway because the client's `command-prompt` parser dropped a value-taking flag's value into the template.

tmux 3.6a renders its own binding as

```
bind-key -T prefix . command-prompt -T target { move-window -t "%%" }
```

psmux has no `{}` command blocks, so `PREFIX_DEFAULTS` spells the same command the way psmux's parser reads it:

```
command-prompt -T target "move-window -t '%%'"
```

`parse_command_line` strips the quotes again before `move-window` sees the index, so the two are equivalent — `list-keys` prints the quoted form rather than tmux's braces.

## Why the binding alone was not enough

The parser consumed a value for `-I` and `-p`, skipped `-1`/`-N`/`-W` as booleans, and for every other flag skipped **the flag but not its value**. So `-T target "move-window -t '%%'"` built the template `target move-window -t '%%'`, and Enter sent `target move-window -t '3'` — not a command. `-t` was in the same hole.

This was not only about the missing default: a user's own `bind . command-prompt -T target "move-window -t '%%'"` hit the same parser, since the dispatcher takes the command string out of the synced binding table without caring where it came from. Measured on `ddd38bc` with nested psmux (outer pane hosting an attached inner psmux, `send-keys` as real key input, the inner window list as witness): the binding is stored and listed, the prompt opens, `5` + Enter moves nothing. The same config on this branch moves the window to 5. Written without `-T` (`-p index` instead) it worked on master too, which was the only workaround.

psmux already knows the shape of every flag: `ARGS_TEMPLATES` in `cli.rs` carries `command-prompt`'s `1CbeFiklI:NPp:t:T:` transcribed from tmux's `cmd_entry`, and one validator reads it (#635). The parser was not asking. `cli::flag_takes_value` exposes that same classification so a parser that has to step over a value cannot drift from the table, and the parse moves out of the event loop into `parse_command_prompt_args`, where it is testable.

Three shapes now follow tmux's `args_parse_flags` rather than the old code: the flags end at the first token that is not one (so a `-t` inside an unquoted template stays in the template), `--` ends them, and a value glued to its flag (`-pindex`) satisfies it. `-T` and `-t` are consumed and dropped — psmux has neither prompt-type completion nor per-client prompt routing.

## Not a change to move-window (re #601, #602)

`move-window` is untouched: no diff in `types.rs`, `server/`, `main.rs` or `commands.rs`. Its target resolution and state invalidation already landed in master as `f497d7b` (#602) and `8986564` (#601), and this change builds on them — because `resolve_window_spec` is complete, `+1`, `-1`, `$` and `{end}` typed at the prompt all work, and an occupied index is refused with `index in use: N` instead of silently swapping.

## Commits

1. **`fix(command-prompt): a value-taking flag consumes its value instead of leaking it into the template`** — `cli::flag_takes_value`, and the parse extracted into `parse_command_prompt_args`. No default binding uses `-T`/`-t` before this PR, so nothing changes for `:` or for a user's `command-prompt -I ... 'cmd "%%"'`.
2. **`feat(keys): bind prefix + . to move the current window to a typed index`** — the default binding, the same entry in the pre-sync fallback table (where `:` and `'` already live, so `.` is not dead for the moment before the first state dump), tests and docs. The fallback entry exposed a latent leak: the prompt's template was never cleared when the prompt closed, only overwritten by the next `command-prompt` binding, so a following `:` in that window would have inherited `move-window -t '%%'`. It is now cleared at all three close sites.
3. **`fix(command-prompt): with no -p, the prompt names the command it will run`** — tmux fills the prompt in from the template when the binding gives no `-p` (`cmd-command-prompt.c`: `xstrndup(template, strcspn(template, " ,"))` → `"(%s) "`). Measured rather than read off the source: tmux 3.6a nested inside tmux, `prefix .` on the inner server, the outer server capturing the pane — the inner status line reads `(move-window)`. psmux said `command` for every templated prompt, so `.` and `:` looked identical while one was about to renumber a window. The heading now reads `(move-window)`; a `-p` still wins. **Separate commit because it also changes the heading of a user's own templated `command-prompt` binding** (towards what tmux shows there) — drop it if you would rather not.

## Tests

`tests-rs/test_prefix_dot_move_window.rs`, 19 unit tests, no server started. They cover the binding surviving `populate_default_bindings` into the live prefix table; the `-T`/`-t` regression; booleans, `--`, glued values and unquoted templates; the unchanged `-I '#W' 'rename-window "%%"'` shape; the derived heading; and the substituted line all the way to the argv `parse_command_line` produces, including that it passes `validate_command_line_flags`. `move-window`'s own behaviour stays pinned by `test_issue601_602_move_swap_window.{rs,ps1}`.

Existing suites re-run green: `client::test_issue345_command_prompt_utf8`, `client::tests::paste_`, `commands::tests_flag_parity::command_prompt`, `types::tests_issue601_602_move_swap_window`, `commands::tests_issue210_gastown_fixes`, `config::tests_issue198`, `config::tests_issue209` (78 tests).

## Verified

Windows 11 Pro 26200, Intel Core Ultra 5 225U, psmux built from this branch on top of `ddd38bc`, isolated `-L` namespace.

Headless:

| | result |
|---|---|
| `list-keys` | `bind-key -T prefix . command-prompt -T target "move-window -t '%%'"` (tmux's command, psmux's quoting) |
| `move-window -t 5` on `1 2 3*` | `1 2 5*`, rc 0 |
| `move-window -t 1` (occupied) | `ERROR: index in use: 1`, nothing moved, rc 1 |
| `move-window -t +3` from `0*` | `1 2 3*`, rc 0 |

Attached, by hand: `prefix .` opens a prompt headed `(move-window)`; a typed index moves the current window and the status line follows it; an occupied index does nothing; Escape closes without running anything; `prefix '`, `prefix :` and `prefix ,` are unchanged.

https://claude.ai/code/session_01NEBUMwTtA5BFQGZNkHmsCg

### DIFF
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -88,6 +88,7 @@ rebind the prefix, the new key is bound to `send-prefix` for you.
 | `Prefix + &` | Kill current window (with confirmation) |
 | `Prefix + ,` | Rename current window |
 | `Prefix + '` | Prompt for window index (jump to any window) |
+| `Prefix + .` | Prompt for an index and move the current window there (`move-window`) |
 | `Prefix + 0-9` | Select window by number |
 
 **Gotcha: `select-window-index` is not a real command.** `Prefix + '` is bound to
--- a/docs/tmux_args_reference.md
+++ b/docs/tmux_args_reference.md
@@ -511,6 +511,7 @@ A psmux extension that creates a pane floating above the tiled layout.
 - Boolean: `-N` (numeric input only), `-W` (word input only)
 - Value: `-I` (initial value), `-p` (prompt list), `-T` (prompt type), `-t` (target)
 - Not accepted: `-1`, `-b`, `-e`, `-F`, `-i`, `-k`, `-l`
+- `-T` is parsed and its value consumed, but the prompt type itself is ignored: psmux has no target completion. `-t` is consumed and ignored the same way. The flags end at the first argument that is not a flag, so a template given unquoted (`command-prompt move-window -t %%`) keeps its own flags.
 
 **choose-tree**, **choose-window**, **choose-session**, **choose-client**, **choose-buffer**, **customize-mode**, **clock-mode**
 - No flags. These open client side overlays. While an overlay is open its keys are handled before any key table, so a bound key does not reach the tables until the overlay closes.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -316,6 +316,23 @@ pub fn validate_command_line_flags<S: AsRef<str>>(tokens: &[S]) -> Result<(), St
     validate_flag_arguments(command.as_ref(), &tokens[1..])
 }
 
+/// True when `-<flag>` takes a value for `command`, per [`ARGS_TEMPLATES`].
+///
+/// Same classification [`validate_flag_arguments`] applies, exposed so a parser
+/// that has to step OVER a flag's value cannot drift from the one flag table.
+/// Without it a hand-written parser has to keep its own list of value-taking
+/// flags, and the flag it forgets turns its value into a positional argument.
+pub fn flag_takes_value(command: &str, flag: char) -> bool {
+    let Some(template) = args_template(command) else {
+        return false;
+    };
+    let kind = match flag_value_kind(template, flag) {
+        FlagValue::None => flag_value_kind(psmux_extra_flags(command), flag),
+        kind => kind,
+    };
+    kind != FlagValue::None
+}
+
 /// Normalize `-x=VALUE` short-flag forms into `["-x", "VALUE"]`.
 ///
 /// tmux accepts both `-t VALUE` (space) and `-t=VALUE` (equals) for
--- a/src/client.rs
+++ b/src/client.rs
@@ -179,6 +179,73 @@ fn extract_confirm_command(args: &str) -> String {
     args.to_string()
 }
 
+/// What a `command-prompt` binding's flags asked the client to do: the text the
+/// prompt opens with (`-I`), the overlay heading (`-p`), and the command to run
+/// with `%%` replaced by whatever the user typed. A bare `command-prompt`
+/// leaves all three empty and the typed line is itself the command.
+#[derive(Debug, Default, PartialEq, Eq)]
+struct CommandPromptSpec {
+    initial: String,
+    label: Option<String>,
+    template: Option<String>,
+}
+
+/// Parse the arguments of a `command-prompt` binding — everything after the
+/// command name.
+///
+/// A value-taking flag consumes the token after it, classified through the one
+/// flag table in cli.rs (`cli::flag_takes_value`) rather than a list kept here.
+/// tmux's own `bind . command-prompt -T target "move-window -t '%%'"` is the
+/// case that needs it: skipping the flag but not its value left `target` as the
+/// first word of the template, so Enter ran `target move-window -t '3'`.
+///
+/// A value glued to its flag (`-pindex`) satisfies the flag on its own, the way
+/// tmux's `args_parse_flags` treats a template letter it finds a character
+/// after. `--` ends the flags, as it does there.
+fn parse_command_prompt_args(args: &str) -> CommandPromptSpec {
+    let tokens = crate::config::shell_words(args);
+    let mut spec = CommandPromptSpec::default();
+    let mut positional: Vec<String> = Vec::new();
+    let mut i = 0;
+    while i < tokens.len() {
+        let token = tokens[i].clone();
+        if token == "--" {
+            positional.extend(tokens[i + 1..].iter().cloned());
+            break;
+        }
+        // A bare `-`, and anything not starting with `-`, is a positional.
+        let flag = token.strip_prefix('-').and_then(|rest| {
+            let mut chars = rest.chars();
+            chars.next().map(|f| (f, chars.as_str().to_string()))
+        });
+        let Some((flag, glued)) = flag else {
+            positional.push(token);
+            i += 1;
+            continue;
+        };
+        let value = if !crate::cli::flag_takes_value("command-prompt", flag) {
+            None
+        } else if !glued.is_empty() {
+            Some(glued)
+        } else {
+            i += 1;
+            tokens.get(i).cloned()
+        };
+        match flag {
+            'I' => spec.initial = value.unwrap_or_default(),
+            'p' => spec.label = value,
+            // -t (target) and -T (prompt type) are consumed and dropped: psmux
+            // has no per-client prompt routing and no prompt-type completion.
+            _ => {}
+        }
+        i += 1;
+    }
+    if !positional.is_empty() {
+        spec.template = Some(positional.join(" "));
+    }
+    spec
+}
+
 /// One row of the choose-tree model: `(is_win, win_id, pane_id, label, session)`.
 /// A session header is `is_win == true` with `win_id == usize::MAX`.
 pub(crate) type TreeRow = (bool, usize, usize, String, String);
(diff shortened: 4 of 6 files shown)


psmux:
Merged to master in 96bb88c, with your three commits kept under your own authorship (5457334, 71becfa, 67872bb) behind the merge commit 717a5c7.

Before touching the diff I reproduced both halves on the unmodified cebc8cf build, three runs each, driving a real attached client with injected console keystrokes and reading the client's own screen back. `list-keys -T prefix` had 63 bindings and no `.`. With your `bind . command-prompt -T target "move-window -t '%%'"` line the prompt opened, the index was accepted, and the window list stayed `0 1 2*`. Naming a real command as the prompt type (`-T display-message`) made the leak visible on the status line: it printed `move-window`, which is exactly the `-T` value having become the command. The `-p index` control moved the window every time. So your diagnosis is right on every point, and `ARGS_TEMPLATES` for `command-prompt` matches tmux's `cmd_entry` character for character.

One gap I fixed on top rather than sending back (d5d1476): the new parser read only the first letter of a flag token, so a bundled form lost its tail. tmux writes its own defaults that way, `-pindex` for the index prompt and `-kpkey` for the key describer, and on the PR as submitted `-kpkey` arrived as a bare `-k` with `pkey` dropped. The parser now walks the characters of one token the way `arguments.c` does, with unit assertions for `-kpkey`, `-1N` and `-1p index`.

I kept `-T target`. tmux HEAD dropped the `target` and `window-target` prompt types after 3.7b (7b02cb6d, "Fix . and ; bindings"), but every released tmux still binds it that way and psmux consumes and ignores the flag either way, so a copied config keeps working. The reasoning is recorded next to the `PREFIX_DEFAULTS` entry so nobody has to rediscover it.

Also landed with it: your 19 unit tests plus one of mine (20 of 20), and a new `tests/test_issue651_prefix_dot_move_window.ps1` that drives prefix `.` into a live client for the default binding, the `-T target` form, `-p index`, `-t <client>`, the bundled booleans, `+1` and `$` targets, an occupied index refused with tmux's `index in use` message, Escape, and the `(move-window)` heading. It scores 19 of 19 on the merge and 8 of 19 on cebc8cf, with the eight controls passing on both. `cargo test --bin psmux` is 3702 passed, 0 failed, and the 27 prompt, bind, rename and move window suites are 344 passed, 0 failed.

Thank you for the measured report and the clean patch. If you like psmux, please consider sharing it with someone or supporting the project through sponsorship.

Issue #611, opened 2026-08-27
Shift+Enter degrades to plain Enter under host load: augment_enter_shift polls key state at processing time, not event time


AminDhouib:
## Shift+Enter degrades to plain Enter on a loaded host — `augment_enter_shift` polls key state at processing time, not event time

psmux 3.3.6, Windows 11 26200, Windows Terminal, local (non-SSH) panes. Shift+Enter in Claude Code intermittently-to-usually submits instead of inserting a newline. Same terminal without psmux: works. This is the local sibling of #396/#397 (which fixed the SSH decode path in 3.3.7) — the local path still rides #121's `augment_enter_shift` hack.

## Why it races

`augment_enter_shift` compensates for VT's inability to encode modified Enter by polling `GetAsyncKeyState(VK_SHIFT)` when the client processes the Enter KeyEvent. That reads the keyboard's state **now**, not at the time the key event was generated. Whenever event processing lags the keystroke — and on a host running a heavy compute fleet it lags by design (see #608: client and server are windowless Normal-priority processes with no foreground boost) — the user has already released Shift by the time the poll runs. Result: the harder the box is working, the more often Shift+Enter silently becomes Enter, which in Claude Code means "submit the prompt" instead of "newline". Fast typists lose the race even on an idle box, since Shift-release can precede the client's event-loop turn.

A state poll can never be made reliable here: the information needed (modifiers at event time) existed in the original KEY_EVENT_RECORD and was discarded upstream of the poll. #159 carried the real fix (win32-input-mode end-to-end, modifiers preserved at the source) but was closed for inactivity without merging.

## Ask

Same shape as #610: land the narrow slice — preserve Enter's modifiers from the actual event (crossterm on a real Windows console does surface them via ReadConsoleInput; the SSH path already decodes ESC+CR since #397) and stop consulting `GetAsyncKeyState` when the event itself carries the answer. The existing ESC+CR delivery to the child is already right (libuv drops SHIFT on VK_RETURN — nodejs/node#2890 — so the Alt-style `\x1b\r` is what Node apps actually recognize); it's only the detection side that races.

Same fleet as #608/#609/#610; happy to test a build under real load.

AminDhouib:
[...]

PR #653, opened 2026-09-12
fix(#654): support CSI u extended keys on input

### LINKED ISSUE #654: `CSI u` key sequences are mishandled
psmux does not handle `CSI u` key sequences on input. `CSI <code> ; <modifiers> u` (fixterms / modifyOtherKeys) is how a terminal reports a key that legacy VT cannot encode; tmux has parsed it since 3.2 (`tty-keys.c`). psmux parses it nowhere.

Because psmux has two separate input paths, that one gap surfaces as two different failures, in opposite directions:

| Path | Terminals | Symptom on `493d721` |
|---|---|---|
| Console records | Windows Terminal, conhost, VS Code | the sequence is delivered to the pane **as literal text** |
| VT bytes | SSH, WezTerm, JetBrains | the key is **silently discarded** |

Neither reads as a missing feature from the user's seat: one types characters nobody pressed, the other swallows a key that was pressed.

## Already fixed? No — and this is worth settling first

Three issues in this area have landed fixes, and every one of them is about a key arriving in some *other* shape:

| | what landed | mechanism | covers |
|---|---|---|---|
| #409 (closed) | `1bf846a` | Ctrl+Enter's character payload set to LF instead of CR, `VK_RETURN` + Ctrl metadata kept | Ctrl+Enter, console path |
| #610 (open, fix landed) | `8da3313`, `8afd754` | `modified_key_name()` extended to Backspace; VT path stops mapping byte `0x08` to an unmodified Backspace | Ctrl+Backspace, Ctrl+H |
| #611 (open, fix landed) | `9b5cb6a` | Shift+Enter read from the key record (win32 input mode negotiates the modifier), `ESC CR` forwarded atomically | Shift+Enter, console path |

None of them decodes a `CSI u` sequence. Each carries one key, and each needed its own mechanism, because psmux has no general encoding for a modified key on input. #159 proposed the general fix — win32-input-mode, "replace per-key hacks with uniform modifier encoding" — and was closed on 2026-06-14 without merging: `src/win32_input.rs` does not exist on `493d721`, and `git grep 9001 -- src/` is empty.

`CSI u` is the other general route, and the one tmux itself uses — `tty-keys.c` has parsed it since 3.2. It is not a competitor to win32-input-mode: that is psmux asking a *Windows* terminal for a Windows encoding, while `CSI u` is any terminal telling psmux, in the cross-platform encoding, what was pressed.

Two things those three fixes cannot reach:

* **The console-path half is a corruption bug on its own terms.** Whatever one concludes about who ought to send `CSI u`, psmux typing `[13;2u` into the user's shell is wrong — six characters nobody pressed land in the pane.
* **On the VT path, `CSI u` is the only encoding available for a whole class of keys.** #610's VT-side fix works by special-casing the incoming byte `0x08`; that technique exists only for keys that *have* a legacy byte. `C-Tab` has none. Neither does `C-S-Tab`, nor `C-i` as distinct from `Tab`.

## Why it matters — keys with no other representation

CR is CR whether or not Shift is down, and C0 collapses whole families of Ctrl combinations onto one byte. These have no other encoding:

* `S-Enter`, `C-Enter`, `C-S-Enter`
* `C-Tab`, `C-S-Tab` — a common want for window switching
* `C-i` distinct from `Tab`, `C-m` distinct from `Enter` (both `0x09` / `0x0d` otherwise)
* `C-;`, `C-'`, `C-,`, `C-.`
* `C-S-<letter>` distinct from `C-<letter>`

psmux already carries a tmux-parity remap table for exactly this collision — `ctrl_char_send_keys_byte` (`src/input.rs:1927`), called live from `src/commands.rs` and `src/server/mod.rs` for `send-keys`. That table exists on the *output* side because these keys collide in C0. On the input side there is nothing that can tell them apart.

Once these arrive as real key events they also become reachable by `bind-key`, which today they are not.

## Reproduction

Both need a terminal deliberately configured to emit the sequence, because psmux never asks a terminal for this encoding. That is part of the report rather than an excuse for it: psmux has no way to ask, and no way to accept it when a terminal does send it — including a terminal the user has switched to `CSI u` globally, or an SSH client from one that emits it.

**Console path** — Windows Terminal, in `settings.json`:

```json
{ "command": { "action": "sendInput", "input": "\u001b[13;2u" }, "keys": "shift+enter" }
```

Inside a psmux pane, press Shift+Enter: the text `[13;2u` appears in the pane. Note the missing `ESC` — see the measurement below for why.

**VT path** — WezTerm, `~/.wezterm.lua`:

```lua
config.keys = {
  { key = 'Enter', mods = 'SHIFT', action = wezterm.action.SendString '\x1b[13;2u' },
}
```

Inside a psmux pane, press Shift+Enter: nothing happens at all.

## The measurement — console path

conhost's input parser does not know `CSI u`, so it flushes the unrecognised sequence into the console input buffer one record per byte. Reading `INPUT_RECORD`s directly under Windows Terminal, one press of Shift+Enter delivers seven records in a **single** `ReadConsoleInputW`:

```text
DOWN vk=0x00 scan=0x00 uChar=0x001b        the ESC
DOWN vk=0x00 scan=0x00 uChar=0x005b  '['
DOWN vk=0x00 scan=0x00 uChar=0x0031  '1'
DOWN vk=0x00 scan=0x00 uChar=0x0033  '3'
DOWN vk=0x00 scan=0x00 uChar=0x003b  ';'
DOWN vk=0x00 scan=0x00 uChar=0x0032  '2'
DOWN vk=0x00 scan=0x00 uChar=0x0075  'u'
```

Two facts in there decide what any fix can do.

**The ESC does not survive.** With no virtual key code to name it and a `uChar` inside the control range, crossterm hands that record to `ToUnicodeEx`, which answers nothing for a synthesised record, and drops the event (`event/sys/windows/parse.rs`). psmux is left with six bare characters. The paste heuristic then sees three or more ASCII characters inside 20 ms (`src/client.rs:2980-2984`), calls it a paste, and the child receives the text. So `[` is the only introducer a fix has left to key off.

**They arrive in one batch.** Every remaining byte is already queued when the `[` is handed over.

## The VT path

`needs_vt_input()` routes SSH sessions, WezTerm and the JetBrains terminals through `src/ssh_input.rs` instead. That parser assembles the sequence correctly and then has nowhere to put it: `dispatch_csi` (`src/ssh_input.rs:1462`) has arms for `A B C D H F P Q R S Z I O t ~` and no `'u'`. The sequence is parsed and dropped.

This is the silent half of the bug, and it is only visible from inside one of those terminals.

## tmux parity, measured — parsing is unconditional and no option is involved

In tmux, `extended-keys` governs what tmux **writes** to a pane, never what it accepts from the terminal, so parsing is unconditional there. That is the claim this request rests on, so it is measured rather than read off the source.

tmux 3.6a, nested: the outer server hosts a pane whose child is an inner tmux, attached, so `send-keys -H 1b 5b 31 33 3b 32 75` into that pane is `CSI 13;2u` arriving on the inner client's stdin exactly as a terminal would send it. The inner pane's child is `sh -c 'stty raw -echo; cat -v'`, so whatever the inner tmux writes to the pane is printed with its control bytes visible. Inner config is only the line shown — no `terminal-features`, and no `-T` capability claim from the fake terminal.

| inner tmux config | what the pane's child received |
|---|---|
| *(none — `extended-keys off`, the default)* | `^M` |
| `set -s extended-keys on` | `^M` |
| `set -s extended-keys always` | `^[[27;2;13~` |
| `set -s extended-keys always` + `extended-keys-format csi-u` | `^[[13;2u` |

The pane never sees the seven literal bytes in any row, so tmux consumed and decoded the sequence in all four — with no option set, and with nothing in `terminal-features`. The option only chooses what tmux re-emits: the legacy `CR` when it cannot or may not use an extended encoding, and otherwise the form `extended-keys-format` names (`xterm`'s `CSI 27 ; mods ; code ~` by default, `CSI code ; mods u` with `csi-u`).

Two things follow for psmux. Accepting `CSI u` on input needs **no new option**. And `set -s extended-keys always` in a shared `.tmux.conf` should keep doing nothing here until the output half exists.

## What this is not

Re-emitting an extended key to a pane, and `extended-keys` / `extended-keys-format` as options, are a separate and still-unimplemented feature. psmux writes the legacy byte, which is the `^M` row above. This report is the input half only.

## Environment

Windows 11 Pro 26200, Intel Core Ultra 5 225U.

Code references above were checked on `493d721` (current master). The hardware measurements were taken on `54aa541`; `git log 54aa541..493d721 -- src/ssh_input.rs src/client.rs src/platform.rs src/input.rs` is empty, so no input-path file changed between the two and the measurements describe current master.

Terminals: Windows Terminal (console path), WezTerm 20240203-110809-5046fc22 (VT path). tmux measurements on 3.6a under WSL, isolated `-L` namespaces. **Not checked:** the JetBrains terminals — `needs_vt_input()` routes them through the same reader and parser as WezTerm and SSH, and that path is exercised above, but not on that host.

## What I would expect

A `CSI u` sequence arriving from the terminal is decoded into the key it names, on both paths, through one shared decoder so the two cannot disagree about what `CSI 9;5u` is — and with nothing to turn on, matching tmux. The key then reaches the pane's child in whatever encoding psmux already uses for it, and becomes bindable.

A fix for both paths, with 21 unit tests, is in #653.

tanaeakihiko:
Fixes #654.

## What

psmux does not understand `CSI u` key sequences on input. Both of its input paths mishandle them, in opposite directions:

| Path | Terminals | Symptom today |
|---|---|---|
| Console records | Windows Terminal, conhost, VS Code | the sequence is delivered to the pane **as literal text** |
| VT bytes | SSH, WezTerm, JetBrains | the key is **silently discarded** |

Neither is a missing feature — one types characters the user never pressed, the other swallows a key that was pressed. psmux never asks a terminal for this encoding, so a `CSI u` sequence only ever arrives because the user deliberately configured their terminal to send it.

## Why it matters

`CSI <code> ; <modifiers> u` (modifyOtherKeys / fixterms) is the only way a terminal can report keys that legacy VT cannot encode. CR is CR whether or not Shift is down, and C0 collapses whole families of Ctrl combinations onto the same byte. These keys have no other representation:

- `S-Enter`, `C-Enter`, `C-S-Enter`
- `C-Tab`, `C-S-Tab` — a common want for window switching
- `C-i` distinct from `Tab`, `C-m` distinct from `Enter` (both `0x09` / `0x0d` otherwise)
- `C-;`, `C-'`, `C-,`, `C-.` — psmux already carries a tmux-parity remap table (`ctrl_char_send_keys_byte`) precisely because these collide in C0
- `C-S-<letter>` distinct from `C-<letter>`

Once these arrive as real key events they can also be reached by `bind-key`, which today they cannot.

This is tmux parity: `tty-keys.c` has parsed `CSI u` since 3.2. Note that in tmux `extended-keys` governs what tmux **writes** to a pane, never what it accepts from the terminal — parsing is unconditional there, and is unconditional here. **No new option is added.**

## The measurement

conhost's input parser does not know `CSI u`, so it flushes the unrecognised sequence into the console input buffer one record per byte. Reading `INPUT_RECORD`s directly under Windows Terminal on Windows 11 26200, one press of Shift+Enter delivers seven records in a **single** `ReadConsoleInputW`:

```text
DOWN vk=0x00 scan=0x00 uChar=0x001b        the ESC
DOWN vk=0x00 scan=0x00 uChar=0x005b  '['
DOWN vk=0x00 scan=0x00 uChar=0x0031  '1'
DOWN vk=0x00 scan=0x00 uChar=0x0033  '3'
DOWN vk=0x00 scan=0x00 uChar=0x003b  ';'
DOWN vk=0x00 scan=0x00 uChar=0x0032  '2'
DOWN vk=0x00 scan=0x00 uChar=0x0075  'u'
```

Two facts in there decide the design.

**The ESC does not survive.** With no virtual key code to name it and a `uChar` inside the control range, crossterm hands that record to `ToUnicodeEx`, which answers nothing for a synthesised record, and drops the event (`event/sys/windows/parse.rs`). psmux is left with six bare characters; the paste heuristic sees three or more ASCII characters inside 20 ms, calls it a paste, and the child receives the text `[13;2u`. So `[` is the only introducer left.

**They arrive in one batch.** Every byte is already queued when the `[` is handed over, which is what makes keying off `[` safe.

## The change

`EscCoalesce` — which already holds a bare Escape for this class of reason (#611) — gains a look-ahead. `feed_with(ev, now, pull)` takes a `pull` that returns only what the terminal has **already** delivered and `None` the instant it has not.

- A `[` that opens a sequence: every remaining byte is queued, the pulls succeed, `recover_csi_u` decodes the key.
- A `[` a person typed: nothing behind it, the first pull comes back empty, the character goes straight out. **No hold, no added latency** on a character that opens every array subscript. A timer-based approach was tried first and rejected for exactly this reason.
- `Ctrl+[` / `Alt+[` never open a sequence — `Ctrl+[` *is* Escape on a real keyboard.

Nothing pulled is ever dropped: a run that turns out not to be a sequence is handed back in arrival order, because the caller cannot re-read what the pull consumed. The tests fix that contract.

The VT path needed twelve lines: it already assembled the sequence correctly and only lacked a `'u'` arm in `dispatch_csi`. Both paths share one decoder, so they cannot disagree about what `CSI 9;5u` is.

A Kitty protocol release report (`CSI 13;2:3u`) is consumed rather than forwarded, because a release of Enter is what the client's WezTerm workaround promotes back into a press.

`feed()` is kept as a thin `feed_with(.., || None)` wrapper, so the existing #611 tests are unchanged.

### Real terminals, real application: before and after

Windows 11 26200. **F10** was bound to send `ESC [13;2u` (the Shift+Enter sequence) in both Windows Terminal and WezTerm, then Claude Code was started inside psmux and F10 was pressed in its prompt.

Windows Terminal, `settings.json`:

```json
"actions": [
    { "command": { "action": "sendInput", "input": "\u001b[13;2u" }, "id": "User.sendInput.999999" }
],
"keybindings": [
    { "id": "User.sendInput.999999", "keys": "f10" }
]
```

WezTerm, `~/.wezterm.lua`:

```lua
config.keys = {
  { key = 'F10', action = wezterm.action.SendString '\x1b[13;2u' },
}
```

| Terminal | Input path | Released psmux | This branch |
|---|---|---|---|
| Windows Terminal | console records | the text `[13;2u` is typed into the prompt | newline inserted, prompt **not** submitted |
| WezTerm | VT bytes | nothing happens | newline inserted, prompt **not** submitted |

Two details make this result conclusive:

- **F10 has no newline of its own.** Nothing but the bound sequence can produce one, so any newline must come from psmux decoding `CSI u`. A binding on Shift+Enter itself could not rule out a plain Enter getting through some other way.
- **The prompt was not submitted.** In Claude Code, Enter submits and Shift+Enter inserts a newline. A newline without submission means the Shift modifier survived all the way to the application. psmux delivered Shift+Enter, not a stripped Enter.

The "before" column reproduces, in a real terminal with a real application, each of the two failures described above, one per input path. The "after" column shows both fixed.

### Other checks

- **VT path under Windows Terminal**, with `{ "keys": "shift+enter", "command": { "action": "sendInput", "input": "\u001b[13;2u" } }` and `TERM_PROGRAM=WezTerm` (the variable `needs_vt_input()` checks). `PSMUX_SSH_DEBUG=1` confirmed the VT reader was the active path. Shift+Enter inserts a newline in the pane's child.
- Ordinary `[`, Escape, and Ctrl+V paste unchanged; mouse unchanged on the VT path.

21 new tests in `tests-rs/test_extended_keys_csi_u.rs`, covering both paths, the replay ordering, and the no-latency contract for a typed `[`. `cargo clippy` clean on the changed code.

**Not verified:** the JetBrains terminals.

https://claude.ai/code/session_01V72ESMLHrKbNUW5p6foW9A

### DIFF
--- a/examples/csi_u_diag.rs
+++ b/examples/csi_u_diag.rs
@@ -1,0 +1,166 @@
+//! Ground truth for what a console delivers when a terminal sends an extended
+//! key (`CSI 13;2u`, the Shift+Enter of `set -s extended-keys`).
+//!
+//! `key_diag` shows what crossterm MADE of the input; this shows what arrived,
+//! one `INPUT_RECORD` at a time, with the fields crossterm decides on:
+//! `wVirtualKeyCode`, `UnicodeChar` and `dwControlKeyState`.  The distinction
+//! matters because crossterm can drop a record entirely — a record whose
+//! virtual key code is not one it names and whose `UnicodeChar` is a control
+//! code goes to `ToUnicodeEx`, which answers nothing for a synthesised record,
+//! and the event is discarded (crossterm 0.29 `event/sys/windows/parse.rs`).
+//!
+//! Run it in the terminal under test, NOT inside psmux, then press the key:
+//!
+//! ```text
+//!   cargo run --example csi_u_diag
+//! ```
+//!
+//! Press `q` (or Ctrl+C) to quit.  Lines also go to `$TEMP/psmux_csi_u_diag.log`.
+#![cfg(windows)]
+
+use std::ffi::c_void;
+use std::io::Write;
+use std::path::PathBuf;
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct KeyEventRecord {
+    key_down: i32,
+    repeat_count: u16,
+    virtual_key_code: u16,
+    virtual_scan_code: u16,
+    u_char: u16,
+    control_key_state: u32,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct InputRecord {
+    event_type: u16,
+    _pad: u16,
+    data: [u8; 16],
+}
+
+#[link(name = "kernel32")]
+extern "system" {
+    fn GetStdHandle(n: u32) -> *mut c_void;
+    fn GetConsoleMode(h: *mut c_void, mode: *mut u32) -> i32;
+    fn SetConsoleMode(h: *mut c_void, mode: u32) -> i32;
+    fn ReadConsoleInputW(h: *mut c_void, buf: *mut c_void, len: u32, read: *mut u32) -> i32;
+}
+
+const STD_INPUT_HANDLE: u32 = -10i32 as u32;
+const KEY_EVENT: u16 = 0x0001;
+const ENABLE_PROCESSED_INPUT: u32 = 0x0001;
+const ENABLE_LINE_INPUT: u32 = 0x0002;
+const ENABLE_ECHO_INPUT: u32 = 0x0004;
+const ENABLE_WINDOW_INPUT: u32 = 0x0008;
+const ENABLE_VIRTUAL_TERMINAL_INPUT: u32 = 0x0200;
+
+fn log_path() -> PathBuf {
+    let dir = std::env::var("TEMP")
+        .or_else(|_| std::env::var("TMP"))
+        .unwrap_or_else(|_| ".".into());
+    PathBuf::from(dir).join("psmux_csi_u_diag.log")
+}
+
+fn emit(line: &str) {
+    print!("{}\r\n", line);
+    let _ = std::io::stdout().flush();
+    if let Ok(mut f) = std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(log_path())
+    {
+        let _ = writeln!(f, "{}", line);
+    }
+}
+
+/// How crossterm 0.29 resolves this record, so the log says outright whether
+/// the key survives the trip.  Mirrors `parse_key_event_record`: the named
+/// virtual key codes win, and everything else falls back to `UnicodeChar` —
+/// where a control code is handed to `ToUnicodeEx` and, for a record conhost
+/// synthesised while flushing a sequence it did not recognise, comes back
+/// empty.
+fn crossterm_verdict(k: &KeyEventRecord) -> &'static str {
+    match k.virtual_key_code {
+        0x10..=0x12 => "dropped (bare modifier)",
+        0x08 => "KeyCode::Backspace",
+        0x1b => "KeyCode::Esc",
+        0x0d => "KeyCode::Enter",
+        0x09 => "KeyCode::Tab",
+        0x25..=0x28 => "an arrow KeyCode",
+        _ => match k.u_char {
+            0x00..=0x1f => "ToUnicodeEx fallback -> DROPPED unless the layout answers",
+            _ => "KeyCode::Char(u_char)",
+        },
+    }
+}
+
+fn main() {
+    let _ = std::fs::write(log_path(), "");
+    let h = unsafe { GetStdHandle(STD_INPUT_HANDLE) };
+    let mut original: u32 = 0;
+    if unsafe { GetConsoleMode(h, &mut original) } == 0 {
+        eprintln!("stdin is not a console");
+        return;
+    }
+    // Raw, and with VTI on: the mode psmux's client runs its console in.
+    let raw = (original & !(ENABLE_PROCESSED_INPUT | ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT))
+        | ENABLE_WINDOW_INPUT
+        | ENABLE_VIRTUAL_TERMINAL_INPUT;
+    unsafe { SetConsoleMode(h, raw) };
+
+    emit(&format!(
+        "=== csi_u_diag: console mode {:#06x} -> {:#06x} (VTI {}), log {:?} ===",
+        original,
+        raw,
+        if raw & ENABLE_VIRTUAL_TERMINAL_INPUT != 0 { "on" } else { "off" },
+        log_path()
+    ));
+    emit("Press the key under test (Shift+Enter). Press q to quit.");
+
+    let mut buf = [InputRecord { event_type: 0, _pad: 0, data: [0u8; 16] }; 32];
+    loop {
+        let mut read: u32 = 0;
+        let ok = unsafe {
+            ReadConsoleInputW(h, buf.as_mut_ptr() as *mut c_void, buf.len() as u32, &mut read)
+        };
+        if ok == 0 {
+            emit("ReadConsoleInputW failed");
+            break;
+        }
+        let mut quit = false;
+        emit(&format!("--- batch of {} record(s) ---", read));
+        for rec in buf.iter().take(read as usize) {
+            if rec.event_type != KEY_EVENT {
+                emit(&format!("  other event type {:#06x}", rec.event_type));
+                continue;
+            }
+            let k: KeyEventRecord = unsafe { std::ptr::read(rec.data.as_ptr() as *const _) };
+            let ch = char::from_u32(k.u_char as u32)
+                .filter(|c| !c.is_control())
+                .map(|c| format!("{:?}", c))
+                .unwrap_or_else(|| "-".into());
+            emit(&format!(
+                "  {} vk={:#04x} scan={:#04x} uChar={:#06x} {:>4} ctrl={:#06x}  crossterm: {}",
+                if k.key_down != 0 { "DOWN" } else { "UP  " },
+                k.virtual_key_code,
+                k.virtual_scan_code,
+                k.u_char,
+                ch,
+                k.control_key_state,
+                crossterm_verdict(&k),
+            ));
+            if k.key_down != 0 && (k.u_char == b'q' as u16 || k.u_char == 0x03) {
+                quit = true;
+            }
+        }
+        if quit {
+            break;
+        }
+    }
+
+    unsafe { SetConsoleMode(h, original) };
+    emit("=== csi_u_diag done ===");
+}
--- a/src/ssh_input.rs
+++ b/src/ssh_input.rs
@@ -548,6 +548,11 @@ pub fn keepalive_reasserts_mouse_input() -> bool {
 /// has used since #397.
 pub(crate) const ESC_COALESCE_MS: u64 = 50;
 
+/// Ceiling on the parameter bytes collected for one `CSI` sequence.  A real
+/// key sequence is a handful of bytes; anything longer is a stream that merely
+/// starts like one, and must not keep the held Escape alive indefinitely.
+pub(crate) const CSI_MAX_PARAM_LEN: usize = 32;
+
 /// Folds a bare Escape that is immediately followed by Enter into one
 /// `Alt+Enter` event (issue #611).
 ///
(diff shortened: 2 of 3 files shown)


psmux:
Merged to master in 67b9e18, with your commit a7bc298 kept under your own authorship behind the merge commit f393f61, and four follow ups of mine on top.

Before reading the diff I reproduced both halves on the unmodified cebc8cf build with a byte logger in the pane (`stty raw -echo; cat -v`, so every byte psmux writes is visible). Console path: injecting your seven `INPUT_RECORD`s in one `WriteConsoleInputW` made the child receive the literal text `[13;2u`, three runs of three, with the ESC gone exactly as you measured. VT path: with the client routed onto the byte parser inside an outer psmux pane, `send-keys -H 1b 5b 31 33 3b 32 75` produced nothing at all, three of three, while a control of `41 0d 42` arrived as `A^MB`. After the merge both paths deliver `^[^M` (psmux's Shift+Enter encoding) and `CSI 13;2u`, `CSI 9;5u` and `CSI 105;5u` fire `bind -n S-Enter`, `C-Tab` and `C-i`.

One defect I found in the PR and fixed on top rather than sending back (50c02a2): a clipboard paste reaches a console as the same one shot burst of character records, so pasted text containing `[13;2u` was decoded into a key. A paste of `PA[13;2uXY` reached the child as `PA CR XY` on the PR build, three of three, and verbatim on master. The recovery is now gated on a 20 ms text burst window, the same window the client's paste detector already uses: a real sequence opens its own burst because the console dropped the ESC, while inside a paste the bracket follows text by microseconds. A decoded key does not count as text, so a held repeat keeps decoding. Sequence still decodes three of three, paste verbatim three of three, a typed `[ab` verbatim three of three.

Two smaller ones: a Kitty key release report (`code;mods:3u`) is dropped on the VT path too instead of becoming a second press (f8906ad), and modifier bit 8 (Meta) folds into Alt the way `tty_keys_extended_key` does (8fcc439), so `9` is Alt and `13` is Ctrl+Alt. On the parity point you raised, I read `tty-keys.c` again: tmux parses both extended forms from the raw byte stream in `tty_keys_next` with no option and no terminal feature gate, and `extended-keys` only governs what it writes to a pane. psmux now matches that: nothing to turn on, and `set -s extended-keys always` in a shared config keeps doing nothing until the output half exists.

Landed with it: your 21 unit tests plus mine (25 in the module, the modifier table covers params 1, 2, 3, 5, 6, 7, 8, 9, 13), a record injector `tests/csiinject654.cs`, and `tests/test_issue654_csi_u_input.ps1` which drives both paths with real records and real VT bytes and scores 9 of 9 on the fix and 3 of 9 on the old build, the six failures being exactly the #654 assertions. `cargo test --bin psmux` is 3707 passed, 0 failed.

Thank you for a report that measured everything down to the input records. If you like psmux, please consider sharing it with someone or supporting the project through sponsorship.

tanaeakihiko:
Thank you for merging both, and for the follow-ups. The paste case in 50c02a2 is one
I should have tested: a burst of character records is exactly the shape I was keying
off, and I only checked typed input. The same goes for the bundled flags on #652, where
I matched tmux's glued value form but never read far enough into args_parse_flags to
see that one token can carry several flags.

One small follow-up from testing after the merge: under Git Bash's mintty the client runs
behind Cygwin's pseudo console, and that conhost silently drops an unrecognised CSI u
when VTI is off, so neither path sees the sequence there. I am treating that as a
limitation rather than something to fix.

psmux:
Thank you for coming back with the mintty result, that is exactly the kind of after merge observation that saves the next reporter a day.

Your read on it matches what I have measured on this side: under Cygwin's pseudo console the conhost in the middle drops a CSI u it does not recognise when VTI is off, so the sequence never reaches either of our paths. I agree it is a host limitation rather than something psmux should paper over, and I have noted it in the keyboard docs as a known mintty boundary so nobody files it as a psmux regression.

The bundled flags and the paste burst were both cases where reading one more function in tmux would have told us, so no need to be hard on yourself about it. The CSI u work was solid and it is in the product now. I appreciate the care you put into both PRs.

Issue #659, opened 2026-09-15
warm server claimed by a new session keeps the spawner's environment (plugins then cannot find psmux.exe)


tscodeplus:
### What happens

A psmux server started by a process whose `PATH` does not contain the psmux install directory keeps that `PATH` for its whole life — including after a later session claims it. Every `run-shell` child inherits it, so plugin scripts and key bindings that shell out to `psmux` fail with

```
psmux : The term 'psmux' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

psmux surfaces that output in a modal popup that has to be dismissed with `Esc`, and the plugin action simply does nothing. Every shipped plugin (`psmux-resurrect`, `psmux-continuum`, `ppm`, …) resolves the binary via `Get-Command psmux`, so all of them are affected.

### Why the server has the wrong environment

psmux keeps a warm server per namespace. It is spawned by whichever client needs one first, and claiming it later only renames the session — the process environment is never refreshed. So:

* a warm server spawned from a shell that had the psmux directory on `PATH` is fine (the normal case);
* a warm server spawned from anything else (WSL interop, a service, a stripped `PATH`, a terminal that was started before psmux was installed) poisons every later session in that namespace, permanently.

`set -g warm off` avoids the pool entirely, but nothing warns the user that the inherited environment is stale.

### Reproduction (Windows)

```powershell
# 1. Start a server from a process whose PATH lacks the psmux directory.
$env:PATH = (($env:PATH -split ';' | Where-Object { $_ -notlike 'D:\psmux*' }) -join ';')
& D:\psmux\psmux.exe new-session -d -s poisoned

# 2. From a normal shell, run any run-shell in that session
psmux run-shell 'pwsh -NoProfile -Command "Get-Command psmux"'
```

The child reports `NOT_FOUND`, and the output shows up as a modal popup. Measured with 3.3.8 and master `e70323c`: the child's `PATH` contains **0** entries for the psmux directory. With a warm server that was spawned from a good environment the same probe reports the directory once.

The trigger that led me here: `psmux` was first launched from WSL via interop, which spawned `__warm__` with the WSL-side `PATH` (no `D:\psmux`). The Windows session that claimed it afterwards could not run any plugin.

### Workaround

* `set -g warm off` — measured on this machine: cold start 0.70–0.75 s vs claiming a warm server 0.56–1.37 s, so warm buys little here anyway;
* `psmux kill-server -a` and relaunch from a shell that has psmux on `PATH`.

[...]

PR #660, opened 2026-09-15
fix(run-shell): make the psmux binary resolvable for run-shell children

tscodeplus:
## What

`run-shell` children now get a `PATH` on which the psmux binary is resolvable, instead of trusting whatever environment the server happens to have inherited.

## Why

A `run-shell` child inherits the **server's** environment, and the server is not necessarily started by something whose `PATH` contains the psmux install directory — a parked warm server keeps the environment of whatever spawned it, even after a later session claims it (#659). Plugin scripts and key bindings that shell out to `psmux` then fail with

```
psmux : The term 'psmux' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

and psmux surfaces that in a modal popup, so the plugin action silently does nothing. Every shipped plugin resolves the binary with `Get-Command psmux`, so all of them break in such a session.

I hit this for real: psmux was first launched from WSL interop, which spawned `__warm__` with a `PATH` that had no `D:\psmux`. The Windows session which then claimed it could not run a single plugin.

## How

`build_run_shell_command` becomes a thin wrapper around the previous body (now `build_run_shell_command_inner`) and prepends `current_exe()`'s directory to the child's `PATH`. `psmux.exe`, `pmux.exe` and `tmux.exe` always live in the same directory, so one entry covers every name a child may use. An inherited `PATH` that already contains the directory is left exactly as the caller set it up — no duplicate entry, no reordering.

`prepend_if_absent` and `apply_own_dir_to_path` take the inherited `PATH` as a parameter, so the decision is unit-testable without mutating the process environment.

Every run-shell entry point benefits automatically, because they all funnel through `build_run_shell_command`: hooks, key bindings, `run` config lines, client-forwarded commands and the CLI fallback.

## Tests

* `tests-rs/test_run_shell_resolve.rs` — five unit tests: prepend when missing, empty `PATH`, keep an existing entry, case/trailing-separator insensitivity, and the poisoned-`PATH` / no-duplicate behaviour of the wrapper.
* `tests/test_run_shell_child_path.ps1` — end-to-end: starts a server whose `PATH` deliberately excludes the psmux directory and proves the `run-shell` child still resolves `psmux` to that directory, with exactly one entry. On stock 3.3.8 the same probe reports `resolved=NOT_FOUND count=0`, so the test does catch the regression.

```
$ cargo test --release --bin psmux prepend_if_absent   # 4 passed
$ cargo test --release --bin psmux run_shell_child     # 2 passed
$ pwsh -File tests/test_run_shell_child_path.ps1       # RESULTS Passed: 3  Failed: 0
```

## Scope

This is a stopgap for the psmux directory specifically; the underlying problem — a claimed server keeps a stale environment — is tracked in #659, where a proper fix would cover every other tool the inherited environment is missing, not just psmux. It is Windows-only because that is where the failure was reproduced; happy to generalise with `std::env::split_paths`/`join_paths` if you prefer that.

Refs #659


psmux:
Merged with your authorship preserved as 547ff50, on master under the root cause fix for #659 (69d6588 and 9215ced).

I read the whole diff, built it, and ran your three tests plus the hook and claim probes from #659 against it. Nothing to change: `prepend_if_absent` is careful about case and trailing separators, an inherited `PATH` that already knows the directory is left exactly as the caller set it, and routing every `run-shell` through the one wrapper means hooks, bindings, config `run` lines and forwarded commands all get it for free.

On scope, you called it correctly in the PR body. The claim now refreshes the server's environment from the claiming client, so the warm pool case is fixed at the root, and your prepend still earns its place for the case no claim can reach: a server cold started from a shell that genuinely has no psmux directory on `PATH`. Measured on that scenario with your change, the hook child resolves `psmux` from the running binary's directory while the rest of the inherited environment is still the poisoned one, which is exactly the split you described.

Thank you for a PR that came with its own end to end proof; `resolved=NOT_FOUND count=0` on stock 3.3.8 versus one entry on the fix is the kind of before and after that makes merging easy. Details of what I measured and what changed are in my comment on #659.

Issue #604, opened 2026-08-26
Mouse support seems to conflicts in some scenarios


fekir:
I've noticed an inconsistency when using the mouse inside psmux (all test done in Windows Terminal)

~~~~
powershell -> psmux ->        nvim.exe -u NONE -> write something -> use mouse and click to change cursor position
powershell ->          wsl -> nvim     -u NONE -> write something -> use mouse and click to change cursor position
powershell -> psmux -> wsl -> nvim     -u NONE -> write something -> use mouse and click to change cursor position
~~~~

In the first two cases, the cursor position changes successfully.

In the third case (using nvim of wsl inside psmux) the cursor position does not change.

I've also noticed that the cursor position flickers when moving the mouse (without clicking), this happens in following cases

~~~~
powershell -> psmux ->        nvim.exe -u NONE
powershell -> psmux -> wsl -> nvim     -u NONE
~~~~

and does not happen when not using psmux:

~~~~
powershell ->        nvim.exe -u NONE  
powershell -> wsl -> nvim     -u NONE
~~~~

psmux:
Thanks for the careful matrix, it turned out to be two separate defects and both reproduced here with a real attached client, real nvim (Windows and inside WSL) and real injected mouse events, using nvim's own mappings as the oracle for what it received.

## A. Click in nvim under WSL did nothing

```
psmux -> nvim.exe -u NONE            4 of 4 clicks reached nvim
wsl -> nvim -u NONE (no psmux)       4 of 4
psmux -> wsl -> nvim -u NONE         0 of 4
```

Root cause: for a VT bridge child like `wsl.exe`, psmux injects the SGR mouse report into the child's console with `WriteConsoleInput`, and the helper that did so wrapped the write in a `SetConsoleMode` set/restore pair that also toggled `ENABLE_QUICK_EDIT_MODE`. conhost derives "is this client tracking the mouse" from the console input mode, so it mirrored every toggle back up the ConPTY into the pane's own output as `ESC[?1003;1006h` immediately followed by `ESC[?1003;1006l`, one pair per injected event. psmux's terminal parser applied both, and since DECRST 1003 clears the mouse protocol outright (tmux does the same in `input.c`), the trailing reset wiped the DECSET 1002 that nvim had really asked for. The pane's mouse mode went from ButtonMotion to None across the very first forwarded event, and every click after it was suppressed as "the app never asked for the mouse". `PSMUX_PANE_RAW=1` shows the pairs in the raw stream.

[...]

PR #663, opened 2026-09-16
fix(window-size): make "latest" follow the client the user is active in

tscodeplus:
## What

`window-size latest` picked its geometry from whichever client resized **last**, not from the client the user is actually using.

`latest_size_client_id` was only ever written by `CtrlReq::ClientSize` (a real resize). Typing, clicking, scrolling and switching panes did not refresh it, so:

1. a phone client (SSH/Termius) attaches and narrows the window,
2. the user goes back to the desktop and starts working there,
3. the window stays at the phone's size — `latest` still pointed at the phone, and nothing short of another real resize could take it back.

tmux keeps the most recently active client at the head of `clients->latest` (`window-size latest` sizes from `clients->latest->sx/sy`); psmux had no equivalent, so "latest" effectively meant "last to resize".

## How

- `CtrlReq::ClientActivity(u64)` — sent by the server's per-client read loop for every request **except** bare pointer motion (`is_bare_motion_cmd`, so hovering still cannot steal the size, see #604).
- `resize_window::note_client_activity(app, id)` — records the client as the latest one *and* as the latest-sized client, then recomputes the dynamic geometry and returns true only when it actually changed (the caller then resizes the panes). Clients that never reported a size — one-shot CLI clients (`psmux ls`) and CONTROL clients, which send requests constantly — are ignored, so they can never drive the geometry.
- `largest` / `smallest` keep their existing behaviour: activity is a no-op for them.

## Tests

5 new cases in `tests-rs/test_resize_window_compat.rs`:

| test | pins |
| --- | --- |
| `latest_window_size_follows_the_active_client` | desktop activity takes the window back from the phone (200x50 over 80x24) |
| `repeated_activity_from_the_same_client_is_a_no_op` | no resize storms on every keystroke |
| `activity_from_a_client_without_a_size_is_ignored` | `psmux ls` / CONTROL clients cannot drive geometry |
| `largest_and_smallest_ignore_which_client_is_active` | only `latest` changes |
| `phone_going_away_without_a_resize_still_restores_the_desktop` | detaching the phone + typing on the desktop restores the desktop size |

```
cargo test --bin psmux test_resize_window_compat   # 20 passed
cargo test --bin psmux resize_window               # 23 passed
```


psmux:
Merged at ebc53e2, with one follow up of mine at 19bfbb9, and thank you: the model is exactly tmux's (`server_client_update_latest` moving `w->latest` on every key except focus out and on resize, CONTROL clients skipped), and ignoring clients without a size entry is the right way to cover both tmux's CONTROL exclusion and psmux's one shot CLI clients. Keeping bare pointer motion out of it is a deliberate divergence from tmux and I agree with it; hovering must not steal the window (#604).

What the follow up fixes: the activity ping was only sent from the read at the bottom of the command loop in `src/server/connection.rs`, which runs with a 10 ms timeout right after a command was dispatched. Real typing is spaced further apart than that, so that read timed out, and the next command was picked up by the read at the top of the loop, which sent nothing. End to end the feature did not fire. Your unit tests pass because they call `note_client_activity` directly. Measured with two persistent clients and `window-size latest`:

```
before 19bfbb9
after desktop client-size 200x50 : 200x50
after phone   client-size  80x24 : 80x24
after desktop TYPES 'a'         : 80x24     <- stayed at the phone

after 19bfbb9
after desktop TYPES 'a'         : 200x50    <- latest followed the active client
after phone   TYPES 'b'         : 80x24     (tmux parity: latest follows keys)
after desktop bare mouse-move   : 80x24     (must stay, #604)
after one-shot 'psmux ls'       : 80x24     (a sizeless client cannot drive it)
```

One thing on the record for later: tmux keeps `latest` per window (`w->latest`) while psmux keeps a single `latest_client_id` per server. That is pre existing structure, not this PR, but it is the next divergence if someone wants per window behaviour.

`cargo test --bin psmux resize_window` 23 passed.

PR #664, opened 2026-09-16
feat(mouse): add mouse-drag-enter-copy-mode (tmux's MouseDragStart)

tscodeplus:
## What

Dragging with the left button over a pane that does not track the mouse only painted psmux's client-side selection overlay. tmux treats that same gesture as `MouseDragStart` and enters copy mode (`copy-mode -M`), so the selection is a *real* copy-mode selection: the mode keys stay available while dragging, the selection is server-side (visible to every client), and the yank goes through the same path as any other copy-mode yank.

The client-side overlay stays the default — this is opt-in, because both behaviours are legitimate.

## How

- New option `mouse-drag-enter-copy-mode` (Bool, default `off`), plumbed through `types.rs`, `config.rs`, `server/options.rs` (`get_option_value` / `apply_set_option` / boolean classification / missing-value toggle), `commands.rs` (`show-options`) and `format.rs` (both lookup tables), plus the client handshake JSON and its two argument lists in `server/mod.rs`. Documented in `docs/configuration.md`.
- `client.rs`: with the option on, a left-button drag in a pane that does not track the mouse sends `copy-enter`, then replays the button-down press at the cell it happened in and forwards the drag/release as copy-mode mouse events, undoing the client-side selection that was armed a moment earlier.
- The drag's first replayed event is a press, so the selection anchors at the **press cell** rather than at the pane's cursor. Release yanks and leaves copy mode, exactly like the existing copy-mode drag path (#62).

## Tests

- `tests-rs/test_mouse_drag_enter_copy_mode.rs` — `defaults_to_off`, `config_can_turn_it_on`, `config_can_turn_it_off_again`, `show_options_reports_it`, `apply_set_option_sets_it`, `is_a_boolean_option`.
- inline `window_ops::tests::client_entered_copy_mode_drag_anchors_at_press_and_yanks` — `copy-enter` → press at (col 5, row 2) → drag to (col 9, row 3) must anchor at `(2, 5)`; release must yank into the paste buffer and return to `Mode::Passthrough`.

```
cargo test --bin psmux mouse_drag_enter_copy_mode   # 6 passed
cargo test --bin psmux client_entered_copy_mode_drag # 1 passed
cargo test --bin psmux copy                          # 148 passed
```


psmux:
Merged at dff6d13, with a follow up of mine at 873970a. The gate matches tmux's `MouseDrag1Pane` binding (`if -F '#{||:#{pane_in_mode},#{mouse_any_flag}}' { send -M } { copy-mode -M }`), the release matches `copy-pipe-and-cancel`, and replaying the press so the selection anchors at the press cell is what `copy-mode -M` does. Keeping it off by default so nobody's overlay changes under them was the right call.

Two things the follow up adds, both found by probing rather than reading:

1. The option was not in `src/server/option_catalog.rs`, so `set -gu mouse-drag-enter-copy-mode` was a silent rc 0 no op. Against the sibling `scroll-enter-copy-mode` in the same probe:
   ```
   after set on          : [mouse-drag-enter-copy-mode on]
   after set -gu         : [mouse-drag-enter-copy-mode on]     <- did not reset
   sibling after set -gu : [scroll-enter-copy-mode on]         <- reset to default
   ```
   After 873970a, `set -gu` returns it to `off`, and a catalog default test now pins it in your test file.
2. The bulk `show-options -g` listing in `src/server/mod.rs` never printed it; only `generate_show_options` in `commands.rs` had been updated, so `show-options -g mouse-drag-enter-copy-mode` worked but the full listing omitted it.

Verified end to end over the control port with the exact verbs the patched client emits: `#{pane_in_mode}` 0 before the drag, 1 during, 0 after release, the release yanked a real selection into the paste buffer, and a plain click without a drag does not enter copy mode. `pwsh-mouse-selection` can be on alongside it and the new option wins, since the press clears the rsel state and the drag branch consumes the pending copy mode entry.

One line worth adding to the docs when you get a moment: the gate sits inside the client side selection block, so the option only takes effect while `mouse-selection` is on. With `mouse-selection off` it is inert. Defensible, since it replaces the overlay, but a reader would not guess it.

`cargo test --bin psmux mouse_drag_enter_copy_mode` 7 passed.

PR #665, opened 2026-09-16
fix(copy-mode): make g/G vi-only and add M-< / M-> for history top/bottom

tscodeplus:
## What

tmux binds `history-top` / `history-bottom` to `g` / `G` in **`copy-mode-vi`** and to `M-<` / `M->` in **`copy-mode`** (`key-bindings.c`).

psmux answered a bare `g` / `G` in *both* tables and never bound `M-<` / `M->`. In emacs mode — the default — that meant any stray `g` threw the view to the very top of the scrollback until the user pressed `Esc`. It is easy to hit in practice: a thumb scroll on a phone is a wheel event that enters copy mode, and the next prompt letter the user types can be a `g`, which copy mode then consumes.

## How

- `handle_key`: `KeyCode::Char('g')` / `Char('G')` are now gated on `app.mode_keys != "emacs"`, and `Alt+<` / `Alt+>` are bound to `scroll_to_top` / `scroll_to_bottom` — which is exactly the pair tmux has in its emacs table and psmux was missing.
- `handle_copy_mode_char` (the path a client's printable characters really take): same gating, so a `g` that arrives as pasted/injected text or from a client that sends characters rather than keys behaves the same.
- `docs/keybindings.md`: the emacs-vs-vi table gains the `g` / `G` row plus a paragraph explaining the split.

Nothing changes for `mode-keys vi`: `g` / `G` still reach history-top / history-bottom, and `M-<` / `M->` are a new extra spelling there (tmux's emacs table keys are always active in vi mode in psmux, as the existing docs note for the other keys).

## Tests

3 inline cases in `src/window_ops.rs`:

| test | pins |
| --- | --- |
| `emacs_mode_bare_g_does_not_jump_to_history_top` | after a scroll into copy mode, `send_text_to_active(app, "g")` leaves the offset untouched and stays in copy mode |
| `vi_mode_bare_g_still_jumps_to_history_top` | vi mode is unchanged (offset > 20) |
| `emacs_mode_alt_less_than_jumps_to_history_top` | `M-<` reaches history-top in emacs mode |

```
cargo test --bin psmux mode_bare_g                    # 2 passed
cargo test --bin psmux emacs_mode_alt_less_than        # 1 passed
cargo test --bin psmux copy                            # 141 passed
```


psmux:
Merged at edf7dd5, with a follow up of mine at 195fe3d. The tables are right against `key-bindings.c`: `M-<` and `M->` live in `copy-mode` (lines 619 and 620), `g` and `G` in `copy-mode-vi` only (694 and 678). Since psmux defaults to `mode-keys emacs`, this is a user visible change (a bare `g` no longer jumps to history top) and it goes in the release notes as such.

The follow up: `M-<` and `M->` were added to `handle_key`'s `KeyCode::Char('<') + ALT` arm, but Alt keys reach copy mode as named keys through `send_key_to_active` in `src/input.rs`, the same reason `M-x` has its own arm there, and the comment beside it says so. The new binding was therefore dead. Measured on a 380 line scrollback with `mode-keys emacs`:

```
send-keys 'M-v'           : pos 1 -> 11    (a pre existing Alt binding works)
send-keys 'M-<'           : pos 1 -> 1     <- dead
send-keys -X history-top  : pos 1 -> 380   (the server command is fine)
```

After 195fe3d, both mode-keys tables:

```
emacs 'g' left the view alone   (1 -> 1)
M-< reached history top         (1 -> 380)
M-> reached history bottom      (380 -> 0)
vi 'g' reached history top      (1 -> 380)
vi 'G' reached history bottom   (-> 0)
M-< works in vi mode too        (1 -> 380)
```

That last line is a small additive divergence: tmux binds `M-<` in `copy-mode` only, not in `copy-mode-vi`. Nothing is displaced by it, so I left it. Also cosmetic: the gate is `mode_keys != "emacs"` rather than `== "vi"`, so a nonsense `mode-keys foo` keeps `g` live. Not worth a change on its own.

`cargo test --bin psmux history_top` 4 passed, plus a new regression test that drives the named key path in both tables.

PR #666, opened 2026-09-16
fix(copy-mode): re-anchor the view across a pane resize

tscodeplus:
## What

A pane resize while copy mode is active could strand the view at the very top of the scrollback, with no way back except `Esc`.

`copy_scroll_offset` counts lines **above the live bottom**. A resize reflows the vt100 screen and can drop retained scrollback (`scrollback_filled()` shrinking), and `set_scrollback` clamps to the retained depth. An offset that was valid before the resize therefore ends up pointing past the oldest retained line: the pane renders that oldest line, and the freeze/dump sync writes the clamped offset straight back, so the view stays pinned there.

Reproduced live with a phone client whose on-screen keyboard toggles the pane height (100 ms polling of `#{pane_in_mode}|#{scroll_position}|#{history_size}`, QPC-timestamped):

```
mode=1 scroll=3771 hist=4359     <- user reading history
mode=1 scroll=4390 hist=508      <- resize (soft keyboard) reflowed retained history
mode=1 scroll=508  hist=508      <- pinned at the top (scroll == history_size)
```

## How

- `copy_mode::offset_after_trim(offset_before, filled_before, filled_after) -> Option<usize>` — pure function. Shifts the offset by the number of lines the trim dropped, so the view keeps showing the same content; returns `None` when the offset can no longer be anchored to anything retained.
- `copy_mode::reanchor_after_resize(app, offset_before, filled_before)` — measures the retained depth *after* the resize, applies the shift, and when there is nothing left to anchor to it leaves copy mode instead of leaving the pane showing its oldest line.
- `tree::resize_window_panes` captures `(copy_scroll_offset, scrollback_filled())` before `resize_node` (only while in `CopyMode` / `CopySearch`) and calls the re-anchor once the panes have their new size.

## Tests

`tests-rs/test_copy_mode_resize_reanchor.rs` (5):

| test | pins |
| --- | --- |
| `no_trim_keeps_the_offset` | a resize that retains history does not move the view |
| `a_trim_shifts_the_offset_by_the_dropped_lines` | 100 → 90 retained shifts 30 → 20 (same content) |
| `a_trim_below_the_offset_clamps_to_the_live_bottom` | 5 → 0 instead of running past the end |
| `an_offset_beyond_the_old_history_can_no_longer_be_anchored` | the live case: `offset_after_trim(4390, 4366, 508) == None` |
| `landing_exactly_on_the_top_of_the_new_history_is_still_valid` | the boundary is kept, not rejected |

```
cargo test --bin psmux test_copy_mode_resize_reanchor   # 5 passed
cargo test --bin psmux copy                            # 146 passed
cargo test --bin psmux resize                          # 60 passed
```


tscodeplus:
The `Test Windows x64` failure is a pre-existing race in `pane::tests_issue615_warm_claim_cwd_hint::a_reading_that_moved_wins_over_the_request`:

```
reported "C:\\Users", expected the live reading "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\"
```

It is unrelated to this change:

- the same job passed on #663, #664, #665 and #667, and on `master`
- the test passes 20/20 locally here
- the reading it depends on is a PEB walk, which is racy for a process that was just created - the header of that very test file measures the same effect on a booting warm spare (peb_own=Some("C:\\"), then Some("C:\\Users") three seconds later). On a loaded runner the test asked "has the reading moved?" before the move was observable, and the product correctly answered "not yet".

I opened #670 to deflake it (wait for the raw reading, 15s deadline, assertions unchanged). Re-running this job now.

psmux:
Merged at 81687e5, with a follow up of mine at 82a04e4. The arithmetic is tmux's: `window_copy_resize` converts the offset to an absolute line before the resize (`cy = gd->hsize + data->cy - data->oy`) and recomputes `oy` against the new hsize, which is algebraically `offset - (filled_before - filled_after)`.

Two defects the follow up fixes:

1. A stolen `#[cfg(test)]`. The new module declaration was inserted between an existing `#[cfg(test)]` and the `mod tests_copy_cancel_stale_state;` it belonged to, so the new module carried two attributes and `tests_copy_cancel_stale_state` carried none and was compiled into the release binary.
2. Wrong window. `reanchor_after_resize` addressed `app.active_idx`, but `resize_window_panes` runs per window from `resize_all_panes`, so a background window's resize measured its own pane's retained depth and applied that shift to the pane the user was looking at. The snapshot is now taken only when `window_index == app.active_idx`.

Verified live: in copy mode at scroll 240 of 563, a 120x40 to 60x20 resize and a grow to 140x45 both left the top visible line on the identical content line (`resize scrollback line 322`) and the view still paged afterwards. An honest caveat: on this machine psmux's vt100 did not actually trim retained history on resize in either run (`history_size` identical before and after), so the trim branch is covered by your unit tests, not end to end. Your phone keyboard reproduction is the real one.

One divergence to consider for a later PR: when the shifted offset can no longer be anchored, this exits copy mode. tmux never leaves copy mode on a resize; it clamps `oy` and lands at the live bottom with copy mode still up. Since #671 removes the corruption this recovers from at the root, it is not urgent, but the `None` arm clamping to the live bottom would be strict parity.

`cargo test --bin psmux copy_mode_resize_reanchor` 5 passed.

psmux:
Follow up on the one divergence I noted above: the `None` arm now behaves as tmux's `window_copy_resize` does, at fabfd2d.

tmux pre clamps `oy` to the history size before the reflow and re derives it afterwards, so an offset that can no longer be reached lands on the oldest retained line and copy mode stays up; it never leaves copy mode on a resize. `offset_after_trim` now returns the clamped offset instead of `None`, and `reanchor_after_resize` never calls `exit_copy_mode`. Your five unit cases are unchanged apart from the one that asserted `None`, which now asserts the clamp (`offset_after_trim(4390, 4366, 508) == 508`), and a sixth drives the real function on a real pane and asserts the mode survives. The pinned view your phone reproduction showed is gone at the root with #671's snapshot, so the resize no longer needs to bail out to recover from it.

`cargo test --bin psmux copy_mode_resize_reanchor` 6 passed, `copy` 169, `resize` 68, and the copy mode and resize E2E suites are clean on the installed build.

Issue #91, opened 2026-03-08
Japanese IME input delayed by ~300ms due to paste-detection heuristic


dy-KingDiamond:
## Summary

When typing Japanese (or other CJK) text via IME, every keystroke is delayed by ~300ms. The paste-detection buffering system in the Windows client misidentifies rapid IME character bursts as clipboard pastes.

## Root Cause

The paste-detection logic in `src/client.rs` buffers all printable characters on Windows and uses a two-stage heuristic:

1. **Stage 1 (0–20ms)**: Accumulate characters in `paste_pend`
2. **Stage 2 (20–300ms)**: If ≥3 characters arrived within 20ms, assume it's a paste and wait up to 300ms for a Ctrl+V Release event

IME composition routinely generates 3+ characters within 20ms (e.g. confirming a kanji conversion), which triggers Stage 2. Since no Ctrl+V Release ever arrives, every IME keystroke incurs the full 300ms timeout.

## Environment

- OS: Windows 11
- Terminal: Windows Terminal
- Shell: PowerShell 7
- psmux: 0.4.10
- IME: Microsoft Japanese IME

## Reproduction

1. Start a psmux session
2. Switch to Japanese IME
3. Type any Japanese text (e.g. `nihongo` → 日本語)
4. Observe noticeable lag (~300ms) on each confirmed conversion

## Expected

IME input should be sent immediately with no perceptible delay, matching the behavior of typing directly in Windows Terminal without psmux.

## Suggested Fix

When the 20ms window expires with ≥3 characters, check whether the buffer contains any non-ASCII characters. If it does, skip Stage 2 and flush immediately as normal text input. Rationale:

- **Windows Terminal Ctrl+V**: paste content is injected directly into the ConPTY pipe (bypassing `ReadConsoleInputW`), so `paste_pend` is typically empty — unaffected by this change
- **IME input**: always produces non-ASCII characters — correctly identified and flushed immediately
- **Legacy console + non-ASCII paste**: the only scenario affected — bracketed paste wrapping would be lost, but this is a rare edge case

## Related

- #74 — the paste-detection heuristic was introduced to fix nvim autoindent issues during paste. This fix preserves that behavior for ASCII pastes.

psmux:
Hey @dy-KingDiamond, thanks for the detailed write-up — the root cause analysis and suggested fix were spot-on. Really appreciate the thoroughness.

[...]

PR #667, opened 2026-09-16
fix(client): stop the Ctrl+V read-back from pasting a CJK burst twice

tscodeplus:
## What

Pasting text from another Windows application into a psmux pane could deliver it **twice**. Reported with a Chinese word: selecting `恭喜通关` in Notepad and pasting into a psmux shell prompt produced `恭喜通关恭喜通关`.

## Why

Paste detection (`paste-detection`, on by default) buffers the characters the console host injects for a paste and decides how to forward them. Non-ASCII bursts are deliberately **not** treated as pastes: the fix for #91 (Japanese IME input delayed ~300 ms) makes a burst of ≥3 characters that contains non-ASCII flush immediately as individual text, so IME composition is never held back. A short CJK clipboard paste takes exactly that path — four characters, flushed as one `send-text` per character.

Those characters are not the only source of the paste, though. On Windows the client has a second path for the same Ctrl+V: when the Release arrives with nothing buffered, it reads the clipboard itself and sends `send-paste` (a fallback that exists for terminals which do not inject the clipboard as character events). The "flush as normal text" branch did not record that it had just delivered a paste, so the fallback read the clipboard and sent the same string again — the pane received it twice.

ASCII pastes are unaffected: a ≥3 character ASCII burst enters the stage-2 path, which opens the 200 ms `paste_suppress_until` window and suppresses the fallback.

## How

- The client remembers the text of the last paste-shaped burst it delivered (`paste_burst_delivered`), and both read-back sites — the Ctrl+V Release fallback and a late `Event::Paste` (Windows crossterm can emit `Event::Paste` *and* the per-character key events for a single paste) — drop a delivery that repeats the **same string** within 300 ms.
- Matching the content, not just the timing, is what keeps this safe: a *different* clipboard payload is never dropped, and deliberately pasting the same text twice still delivers both copies, because the second paste arrives through its own burst first — only the redundant read-back is skipped.
- Every place that forwards `paste_pend` records what it sent, including the interrupt flush (`flush_paste_pend_as_text`) and the zero-latency typing flush, so a 1-2 character paste (`ab`) is covered as well.
- The two pastes the user asks for explicitly — `Ctrl+Shift+V` in `pwsh-mouse-selection` mode and right-click paste — only *record* their delivery. They are separate gestures, so they are never suppressed.

## Tests

`tests-rs/test_client.rs` gains six cases:

| test | pins |
| --- | --- |
| `clipboard_read_back_of_a_just_delivered_burst_is_a_duplicate` | the reported case: the read-back of `恭喜通关` is recognised |
| `a_different_clipboard_payload_is_never_dropped` | `恭喜通关！` / `abc` are delivered normally |
| `a_read_back_after_the_window_is_not_a_duplicate` | 400 ms later is a fresh paste |
| `nothing_delivered_yet_is_not_a_duplicate` | no burst recorded |
| `an_empty_burst_never_matches` | empty strings never match |
| `the_interrupt_flush_remembers_the_cjk_burst_it_delivered` | the sequence from the report: the flush still sends `send-text` per character (issue #91 intact) and the burst is remembered |

Red/green: with `record_paste_delivery` stubbed out, exactly `the_interrupt_flush_remembers_the_cjk_burst_it_delivered` fails (47 passed / 1 failed); restored, the client suite is green again.

```
cargo test --bin psmux client::tests   # 48 passed
cargo test --bin psmux paste           # 59 passed
```


psmux:
Merged as is at 17d7942. Clean and well scoped: recording the text of the last paste shaped burst at every delivery site and dropping a read back only when the content matches inside 300 ms is the right design, and content comparison rather than bare timing is what makes it safe.

I checked the one question that matters for a guard like this, whether it can ever eat a legitimately repeated paste. With a real attached client and `Ctrl+V` written into its console input buffer, each case with a distinct payload:

```
CJK   '甲乙丙丁'  two Ctrl+V 120 ms apart -> 2 occurrences on the pane
CJK   '戊己庚辛'  two Ctrl+V 250 ms apart -> 2
CJK   '壬癸子丑'  two Ctrl+V 600 ms apart -> 2
ASCII 'ZQA'      two Ctrl+V 120 ms apart -> 2
ASCII 'ZQB'      two Ctrl+V 250 ms apart -> 2
ASCII 'ZQC'      two Ctrl+V 600 ms apart -> 2
```

Never, because the text always reaches the pane through the burst, which is never suppressed; only the redundant read back is dropped. And the reported case itself, `恭喜通关` on the clipboard, one `Ctrl+V`: exactly one occurrence on the pane, with a deliberate second `Ctrl+V` giving two.

`cargo test --bin psmux client::tests` 48 passed, `paste` 59 passed. Thank you for the six unit cases; the interrupt flush one is the one I would have missed.

tscodeplus:
Follow-up pushed: the first version compared the forwarded **text** with the clipboard, which the reporter found insufficient. Two more cases still pasted twice:

- `C2单元格应显示` — an ASCII prefix followed by CJK
- `=(B3-B2)/B2` — a formula, so a leading `=` and punctuation

Both are the same mistake. One paste can be delivered as **several bursts**, and what was forwarded is then only a fragment of the clipboard:

| burst | who forwards it |
| --- | --- |
| `C2` / `=` | the zero-latency path (1-2 characters are flushed immediately as typing) |
| `单元格应显示` | the IME heuristic of #91 (≥3 characters containing non-ASCII go out as text) |
| `(B3-B2)/B2` | stage 2 / the Release fallback, depending on timing |

So `duplicates_recent_paste()` compared `单元格应显示` (a fragment) with `C2单元格应显示` (what the clipboard holds), said "different", and the read-back delivered the whole string a second time.

The client now tracks the **gesture** rather than the text: a Ctrl+V Press opens one, every burst forwarded on its behalf is recorded, and both read-back sites ask `PasteGesture::blocks()` — which answers "already delivered" for a fragmented paste as well. A dropped read-back also arms the same 200 ms suppression window, so a phantom Release cannot read the clipboard afterwards. The explicit gestures (Ctrl+Shift+V, right-click paste) keep recording without ever being suppressed.

Red/green: making the decision content-only again fails exactly these three cases:

```
a_fragmented_paste_gesture_blocks_the_read_back_of_the_whole_text ... FAILED
a_fragmented_ascii_formula_blocks_the_read_back_too   ... FAILED
a_gesture_that_forwarded_characters_blocks_the_read_back ... FAILED
test result: FAILED. 49 passed; 3 failed
```

restored: `client::tests` 52 passed, `paste` 60 passed.

tscodeplus:
Thanks for merging this as `17d7942`.

One follow-up, since the reporter hit two more strings that this version alone did not fix (`C2单元格应显示`, `=(B3-B2)/B2`): comparing the forwarded **text** misses a paste the host delivers in several bursts — `C2` goes out as typing through the zero-latency path and the CJK part through the #91 IME heuristic, so the forwarded text is only a fragment of the clipboard and the read-back looked like a fresh paste.

The fix now tracks the paste **gesture** (a Ctrl+V Press opens one, every burst forwarded on its behalf is recorded, and both read-back sites ask it) and is in **#676**, on top of this merge. Red/green included: with the decision made content-only again, exactly the three fragmented cases fail.

psmux:
Thanks for chasing the fragmented cases down rather than stopping at the first green. The gesture tracking in #676 is the right shape and it is merged at dc9ea31, with a real client test of mine on top at 2cd4cb3 that pastes the mixed ASCII plus CJK string and the spreadsheet formula through an attached client and asserts each lands exactly once. Both are in the installed build now, so this thread is complete from my side.

Issue #615, opened 2026-08-28
pane_current_path seems to be not updated correctly in wsl


fekir:
In `psmux.conf`

~~~~
set -g default-shell "powershell -NoLogo"
bind '"' split-window -c "#{pane_current_path}" -e CHERE_INVOKING=1
~~~~

~~~~
cd C:/
psmux
wsl
# current path is C:/
cd Users
# current path is C:/Users
Ctrl+B "
# opens new instance of powershell in C:/ instead of C:/Users
~~~~

the path is tracked correctly in a cygwin (or git-bash) shell

~~~~
cd C:/
psmux
$env:CHERE_INVOKING = "1"; C:/cygwin64/bin/bash.exe --login
# current path is C:/
cd Users
# current path is C:/Users
Ctrl+B "
# opens new instance of powershell in C:/Users
~~~~

I'm not sure if this is an error in `psmux`, or in `wsl`; if possible, I would like to have a consistent behavior.

psmux:
Thanks @fekir, and you asked exactly the right question, because the answer is that it is neither a psmux bug nor a wsl bug in the usual sense, but there is now a consistent behaviour on master (ab0d836, with 6003ef8 and 753b289 for the tests).

**What I measured.** I wrote a small probe that reads the Win32 working directory (the `CurrentDirectory` in each process's PEB, which is what `GetCurrentDirectory` returns inside it) for every process in a pane's tree, and ran your two recipes side by side.

Cygwin bash, after `cd /cygdrive/c/Users`:

```
27080 bash     CWD=C:\Users\
```

WSL, after `cd /mnt/c/Users`, while `capture-pane` shows the prompt sitting at `/mnt/c/Users`:

```
28936 pwsh     CWD=C:\
12740 wsl      CWD=C:\
18820 wsl      CWD=C:\
44204 wslhost  CWD=C:\
```

Cygwin (and msys2 / Git Bash) implement `cd` by also calling `SetCurrentDirectory`, so the Win32 cwd of the bash process follows the Unix one and psmux reads the right answer. Under `wsl` the shell is a Linux process inside the VM; the only Windows processes in the pane are `wsl.exe` and `wslhost.exe`, and their working directory is fixed at the moment they start and never moves again. So the lookup did not fail, it succeeded with a stale value: the directory you were in before you typed `wsl`. That is why `split-window -c "#{pane_current_path}"` opened in `C:\`.

[...]

PR #670, opened 2026-09-17
test(#615): wait for the parked process's cwd reading instead of assuming it

tscodeplus:
## What

`pane::tests_issue615_warm_claim_cwd_hint::a_reading_that_moved_wins_over_the_request` failed in CI once:

```
reported "C:\\Users", expected the live reading "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\"
```

on [this run](https://github.com/psmux/psmux/actions/runs/35047377371/job/104640052647) — the same job passed on four other PRs and on `master`, and the test passed 20/20 locally. It is a race in the test fixture, not in the product.

## Why

The test parks `ping.exe` in a directory (`current_dir(dir)`) and then asserts what `#{pane_current_path}` answers once "the reading has moved off the pool directory". That premise is only true once the reading has *observed* the move — and the PEB walk the reading is built on is racy for a process that has just been created. The header of `test_issue615_warm_claim_cwd_hint.rs` documents exactly this effect on a booting warm spare:

```
pane=2 child_pid=Some(47444) announced=None peb_fg=Some("C:\\")
       peb_own=Some("C:\\") server_cwd=C:\        <- reported C:\
pane=2 child_pid=Some(47444) announced=None peb_fg=Some("C:\\Users")
       peb_own=Some("C:\\Users") server_cwd=C:\   <- same pane, 3s later
```

On a loaded runner `ping.exe` can still report the directory it was created in for a moment after `current_dir` was applied. The test then asks the product to decide "has it moved?", the product correctly answers "not yet" and reports the requested directory, and the assertion fails while everything behaved as designed.

## How

`park_process_in` now waits until `process_info::get_process_cwd(pid)` agrees with the directory the child was created in, with a 15-second deadline that still fails loudly (and names the last reading) if it never does. The waiting is done on the raw reading, so the assertions in the tests stay exactly as they are and keep testing the product rather than the fixture.

```
cargo test --bin psmux tests_issue615_warm_claim_cwd_hint   # 10 passed, 20 consecutive runs
```

## Notes

- Test-side only: the product's behaviour inside that window (report the requested directory until the reading is seen to move) is the fix for #615 and is unchanged.
- If you would rather have the reading be robust at the source, that is a bigger change — the PEB walk would have to retry internally — and I did not want to fold it into a deflake. Happy to open that separately if you want it.


psmux:
Merged as is at 89fdb09. Test only, one helper, bounded poll with a loud failure if the reading never arrives, no product code touched. Ran the #615 tests ten times in a row on the merged tree: 26 passed every time, 0 failed runs.

On your closing question: I would rather not make the PEB walk retry internally. The product's rule of reporting the requested directory until the reading is seen to move is the #615 fix, and a retry inside the reader would blur when a reading is "seen". If a real user hits the window you describe on a loaded machine, the symptom is a briefly stale `#{pane_current_path}`, and that is the honest answer for that moment. Thank you for tracing the CI failure to the fixture rather than papering over it.

Issue #621, opened 2026-09-01
Mouse-wheel scroll stops working while a program is blocking on a stdin read


0xmadumere:
# Problem
when i open up a psmux session(`psmux new -s test`) and then run a script, program or any tool in general that reads input from stdin, WHILE that program is blocked waiting for input, I am completely unable to scroll with the mouse wheel inside that psmux session, the scrolling completely stops working, it instead begins to cycle the prompt history and flicker the cursor, if I kill the program(`Ctrl + C`) the scroll immediately begins to work again. I have mouse mode enabled, scrolling works fine, this only happens when a program ran in that psmux session/window is blocking/waiting for input from stdin. I have tried different input methods using different languages and its all the same issue

# How to reproduce
- save this as test.py:
```python
# prints to force the terminal to scroll downwards
for i in range(40):
    print(i)

input("Try and scroll back up with your mouse, while this is blocked here: ")
```
- open up a psmux session, then once inside that psmux session, run the python program(`python test.py`) and then while the program is blocked waiting for input, try and scroll back up with your mouse, it wont work.
- kill the script (ctrl+c) — scrolling immediately works again
- keyboard-based copy mode entry (prefix + [) does NOT have this problem, only mouse wheel

# Environment:
psmux version: tmux 3.3.8, psmux 3.3.8 (66cf613 2026-08-18)
OS: Windows 11
Terminal: Windows Terminal 1.24.11911.0
Shell inside psmux: Windows PowerShell 7.6.5 
mouse option: on

psmux:
Thank you for the clear report and the tidy repro script. You found a real bug, and I have good news: it is already fixed on master, and your build is the last release that still has it.

What you were seeing, confirmed by reproducing it on exactly your version (3.3.8, 66cf613): that build decided wheel behavior from the pane's foreground process identity. When `python` (or `Read-Host`, or `cmd /c pause`) was the foreground instead of your shell, each wheel notch was translated into three `ESC[A` arrow key writes into the pty rather than opening copy mode. Windows canonical reads answer Up arrow with prompt history recall, which is precisely the history cycling and cursor flicker you described. `Ctrl+C` made pwsh the foreground again, so the wheel appeared to recover.

That foreground process heuristic was removed in PR #548, merged on 2026-08-21, three days after the 66cf613 release build you are running. Wheel forwarding is now decided purely from the pane's terminal state (alternate screen, or a mouse protocol the application itself enabled), which matches how tmux gates its `WheelUpPane` handling.

[...]

Issue #612, opened 2026-08-28
Copy mode's search only works across visible part of a buffer


kirill146:
All of `?`, `/`, `Ctrl+s`, `Ctrl+r` can't see past the on-screen area of the buffer.
Tested on win11, psmux 3.3.8, NUL config, Windows Terminal and Alacritty

psmux:
Thanks @kirill146, you were right on all four, and this one was a single root cause wearing four hats.

**Reproduced** on an 80x24 pane holding `LINE_1` .. `LINE_400`. Searching for a term still on screen worked; the same key with a term that had scrolled into history did nothing at all and cleared `#{search_match}`:

```
after-visible   cy=6 cline=[LINE_385] scroll=0 match=[LINE_385] present=1
after-history   cy=6 cline=[LINE_385] scroll=0 match=[]         present=1
```

**Root cause:** `search_copy_mode` in `src/copy_mode.rs` scanned `for r in 0..last_rows`, which is only the rows the vt100 screen currently frames, and stored every hit as a *visible row index*. The scrollback was never read, and since a match could only ever be a visible row, nothing could move the viewport to bring a hit into view. The `?` and `/` prompts, `Ctrl+s` / `Ctrl+r` in emacs mode, and `n` / `N` all sat on top of that same list, so every opener hit the same wall.

**Fix in 22fa31c** (test tightened in e0748b7, both on master):

* Search now reads scrollback plus screen as one buffer indexed by absolute line, exactly as tmux `window_copy_search_jump` walks `0 .. hsize + sy - 1`.
* Matches are ordered along the search direction starting from the cursor, so `?` stops at the nearest match *above* the cursor rather than the bottom most visible one.
* Landing a match is a port of tmux `window_copy_scroll_to`: a match already on screen leaves the viewport alone, an off screen match is parked a quarter screen up from the bottom. `n` / `N` scroll the same way.
* Bonus: `send-keys -X search-backward <term>` and `search-forward <term>` now honour the optional term argument like tmux instead of silently dropping it and opening the prompt.

**Parity check** against real tmux 3.4 on the same 400 line 80x24 geometry: `search-backward LINE_5` from the live bottom lands on `LINE_59` with the cursor at row 18 (rows minus rows/4) in both. `search-forward LINE_390` from history top gives `scroll_position=7` in both.

[...]

Issue #607, opened 2026-08-27
Copy Mode on Newly Created Panes/Windows


CalebMostyn:
When creating a new pane/window, if the current pane or window is in copy mode, the newly created pane will be opened in copy mode. This behavior is different than `tmux`, wherein a newly created pane does not appear to ever be in copy mode. `psmux` also appears to only show the copy mode identifier on the active pane, but that may be an intentional design change.

This still occurs with a blank configuration (`psmux -f NUL new-session`).

Versions:
```
psmux --version
tmux 3.3.7
psmux 3.3.7 (05cc5d4 2026-07-20)

 [System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      26200  0
```

Logs:
```
===== autorename.log =====
[10:53:57.416] pid=10748 fg_child=None (BFS found nothing)
[10:53:57.416] pid=10748 no_foreground_child
[10:53:58.472] root=28952 snapshot_entries=368
[10:53:58.472] root=28952 no_direct_children
[10:53:58.472] pid=28952 fg_child=None (BFS found nothing)
[10:53:58.473] pid=28952 no_foreground_child
[10:53:58.495] root=10748 snapshot_entries=368
[10:53:58.495] root=10748 no_direct_children
[10:53:58.495] pid=10748 fg_child=None (BFS found nothing)
[10:53:58.496] pid=10748 no_foreground_child
[10:53:59.470] root=28952 snapshot_entries=368
[10:53:59.470] root=28952 no_direct_children
[10:53:59.470] pid=28952 fg_child=None (BFS found nothing)
[10:53:59.471] pid=28952 no_foreground_child
[10:53:59.519] root=10748 snapshot_entries=368
[10:53:59.519] root=10748 no_direct_children
[10:53:59.519] pid=10748 fg_child=None (BFS found nothing)
[10:53:59.520] pid=10748 no_foreground_child
[10:54:00.473] root=28952 snapshot_entries=368
[10:54:00.473] root=28952 no_direct_children
[10:54:00.473] pid=28952 fg_child=None (BFS found nothing)
[10:54:00.473] pid=28952 no_foreground_child
[10:54:00.520] root=10748 snapshot_entries=368
[10:54:00.521] root=10748 no_direct_children
[10:54:00.521] pid=10748 fg_child=None (BFS found nothing)
[10:54:00.521] pid=10748 no_foreground_child
[10:54:00.597] root=30176 snapshot_entries=368
[...]

PR #671, opened 2026-09-17
fix(copy-mode): give copy mode a snapshot of the pane grid, and keep capture-pane live

tscodeplus:
## What

The pane in copy mode showed the pane's **live** grid. So output that kept arriving while the user was reading moved the bottom out from under `copy_scroll_offset`, and once enough history had churned the anchor was evicted: the pane rendered its oldest retained line — `scroll_position == history_size` — until the user pressed `Esc`.

Measured on a machine running a streaming TUI, 100 ms polling of `#{pane_in_mode}|#{scroll_position}|#{history_size}`:

```
64 mode=1 scroll=3
91 mode=1 scroll=30    hist 5334 -> 1163    (reflow while output continued)
93 mode=1 scroll=3476  hist=3476            <- pinned at the oldest retained line
```

and a second capture of the same signature: `826 scroll=61 hist=696` -> `828 scroll=698 hist=698`.

The same frozen view was what `capture-pane` returned. In tmux `capture-pane` reads `wp->base`, the live screen, no matter what copy mode is showing — which is what plugins (continuum-style auto-save) and scripts rely on.

## How

tmux's copy mode shows a *copy* of the grid and leaves `wp->base` alone. This does the same thing:

- `vt100-psmux::Parser::snapshot()` clones the screen into a fresh parser (the parser keeps no handshake state of its own).
- `Pane` gains `live_term: Option<Arc<Mutex<vt100::Parser>>>`. `enter_copy_snapshot()` parks the live parser there and leaves a snapshot in `term`; `leave_copy_snapshot()` restores it and reattaches the live screen at the bottom; both are idempotent. The PTY reader thread holds its own `Arc` from spawn time, so swapping `term` never disturbs what feeds the pane.
- `copy_mode::sync_copy_snapshot()` reconciles on every frame dump — the active window's copy-mode pane holds a snapshot and every other pane releases it. `Mode::Passthrough` is assigned directly in many places, so a per-frame reconciliation is what cannot be missed.
- `tree` resize and `warm_pane_sync` update both parsers.
- `Pane::live_parser()` returns the parked live parser when there is one. `capture-pane` (plain, `-S`/`-E`, styled, and its text helper) reads through it, while the view, selection, search and rendering keep reading the snapshot.

While copy mode is active, `#{history_size}` reports the frozen view's depth rather than the live screen's — deliberate, so the status bar's `scroll/history` pair describes what the user is looking at. Say the word if you would rather have it track the live pane and I will change it.

## Tests

`tests-rs/test_copy_mode_snapshot.rs` (6):

| test | pins |
| --- | --- |
| `entering_copy_mode_installs_a_snapshot_and_parks_the_live_screen` | the two parsers exist and are distinct |
| `pane_output_while_in_copy_mode_cannot_strand_the_view` | the reported defect: output during copy mode leaves the offset where it was |
| `leaving_copy_mode_shows_everything_that_arrived` | nothing is lost on exit |
| `sync_copy_snapshot_reconciles_both_directions` | a pane that leaves copy mode by another route releases its snapshot |
| `resizing_while_in_copy_mode_resizes_both_screens` | both parsers get the new size |
| `capture_pane_reads_the_live_screen_while_copy_mode_shows_a_snapshot` | `capture-pane` answers with the live screen |

Red/green:

- disabling the snapshot install in `enter_copy_mode` fails **4 of 6**, including `pane_output_while_in_copy_mode_cannot_strand_the_view`
- making the capture helper read the view again fails **exactly** the capture test (5 passed / 1 failed)
- restored: `6 passed`

```
cargo test --bin psmux test_copy_mode_snapshot   # 6 passed
cargo test --bin psmux copy                      # 147 passed
cargo test --bin psmux capture                   # 59 passed
cargo test --bin psmux warm                      # 119 passed
cargo test --bin psmux resize                    # 56 passed
```

## Notes

- #666 re-anchors the copy-mode offset when a resize trims retained history — the resize variant of the same symptom. It is a separate, independent fix; I merged the two branches locally and they do not conflict in either order.


psmux:
Merged at a2aa40c. This is the right design and it is `window_copy_init` in spirit: copy mode on a copy of the grid, the reader thread feeding the live parser it already holds, `capture-pane` reading through `live_parser()` the way `cmd-capture-pane.c` reads `wp->base`, and a per frame reconciliation so no `Mode::Passthrough` assignment can leave a snapshot behind. Verified live with a pane streaming output while sitting in copy mode scrolled up, polled every 100 ms for four seconds: `1|200|531` was the only distinct `pane_in_mode|scroll_position|history_size` sample across all forty, `capture-pane` advanced from `stream line 188` to `stream line 510` over the same window, and leaving copy mode jumped history from 531 to 968 with nothing lost.

Four follow ups of mine landed with it, and one decision went against the PR body:

1. c1daa53: `clear-history` stopped working in copy mode. `CtrlReq::ClearHistory` replaced the parser behind `pane.term`, which is now the snapshot, so it wiped the screen the user was reading and left the live scrollback untouched:
   ```
   in copy mode:              1|60|502
   after clear-history:       1|60|0      <- the snapshot was cleared
   after leaving copy mode:   0|0|502     <- the live history is back
   ```
   tmux does `window_pane_reset_mode_all(wp)` then `grid_clear_history(wp->base.grid)` in that order (cmd-capture-pane.c:418): leave copy mode first, then clear the live grid. `window_ops::clear_active_pane_history` now does exactly that, with a regression test in your test file.
2. a929002: `snapshot()` was inserted between `pub fn process` and its doc comment, so the vt100 crate warned on every release build. Restored.
3. 088fbad: two shipped suites (#612 and #621) watched the copy mode viewport by diffing a plain `capture-pane -p`, which only ever worked because psmux used to scroll the live parser in copy mode. Those assertions now ask for the view's region with `-S -<scroll_position>`. I am ratifying this consciously: `capture-pane` no longer follows a scrolled copy mode view for any reason, not only when output arrives, and that is tmux's behaviour (`oy` lives in the mode's own data, `wp->base` does not move). It goes in the release notes because user scripts that scraped the copy mode view will need the same `-S` change.
4. 82796c0: `#{history_size}` and `#{history_bytes}` read the live grid during copy mode. tmux's `format_cb_history_size` answers from `ft->wp->base.grid->hsize` whatever mode the pane is in, and a tmux script that polls it expects history to keep growing, so parity won over the frozen pair. The docs line is updated with it.

Two things measured that you should know about, neither blocking:

- Memory. A 200 column pane with 49952 retained lines costs 180 MB private for the snapshot while copy mode is open (190.7 MB before, 365.3 MB in, 193.9 MB after one cycle). Repeated enter and leave cycles retain about 59 MB and plateau there from cycle 10 through 30, allocator retention rather than a leak. tmux pays the same class of cost by design, but its grid is far more compact than the vt100 cell array. A copy on write share of the scrollback rows would remove most of it; I have opened a follow up issue for that.
- Scope. `sync_copy_snapshot` gates on the active pane of the active window, and `enter_copy_mode` only snapshots that pane, while psmux allows more than one pane in copy mode at once (#607). A non active pane in copy mode gets no snapshot and keeps the old behaviour. Not a regression, and the active pane is the case that matters, but the docs sentence "the view is a snapshot" overstates it a little. Same follow up issue.

Merged tree: `cargo test --bin psmux` 3884 passed, and 19 copy, capture and resize E2E suites clean (358 assertions, 0 failures). Thank you for the red and green table in the PR body; it made the review honest work rather than guesswork.

PR #676, opened 2026-09-18
fix(client): remember the whole paste gesture, not just the last burst

tscodeplus:
## What

Follow-up to #667 (merged as `17d7942`), from a report that two more strings still pasted twice after that fix:

- `C2单元格应显示` — an ASCII prefix followed by CJK
- `=(B3-B2)/B2` — a formula, leading `=` and punctuation

## Why #667 alone was not enough

#667 compared the forwarded **text** with the clipboard. That misses any paste the host delivers as **several bursts**, because what was forwarded is then only a fragment of the clipboard:

| burst | who forwards it |
| --- | --- |
| `C2` / `=` | the zero-latency path (1-2 characters flush immediately as typing) |
| `单元格应显示` | the IME heuristic of #91 (≥3 characters containing non-ASCII go out as text) |
| `(B3-B2)/B2` | stage 2 or the Release fallback, depending on timing |

So `duplicates_recent_paste()` compared `单元格应显示` (the fragment) with `C2单元格应显示` (what the clipboard holds), concluded "different", and the read-back delivered the whole string a second time. Pure CJK and pure ASCII strings are not split, which is why they were already fine.

## How

Track the **gesture** instead of the text:

- a Ctrl+V Press opens one (`PasteGesture::start()`)
- every burst forwarded on its behalf is recorded (`record()`), whether it goes out as `send-text`, `send-paste` or through `Event::Paste`
- both read-back sites — the Ctrl+V Release fallback and a late `Event::Paste` — ask `PasteGesture::blocks()`, which answers "already delivered" for a fragmented paste too
- a dropped read-back also arms the same 200 ms suppression window, so a phantom Release cannot read the clipboard afterwards
- the explicit gestures (`Ctrl+Shift+V` in `pwsh-mouse-selection` mode, right-click paste) keep recording without ever being suppressed

The content comparison is kept as the second half of the decision, so a clipboard payload that was never injected still goes through.

## Tests

`tests-rs/test_client.rs` — the three fragmented cases are new, the rest pin the gesture lifecycle:

| test | pins |
| --- | --- |
| `a_fragmented_paste_gesture_blocks_the_read_back_of_the_whole_text` | the CJK report: `C2` + `单元格应显示` blocks the read-back of `C2单元格应显示` |
| `a_fragmented_ascii_formula_blocks_the_read_back_too` | the formula report |
| `a_gesture_that_forwarded_characters_blocks_the_read_back` | anything forwarded for this gesture suppresses the read-back |
| `a_fresh_gesture_starts_clean` / `finishing_a_gesture_forgets_what_it_delivered` | lifecycle |
| `nothing_delivered_is_never_a_duplicate` / `an_empty_delivery_is_not_a_delivery` | no false positives |
| `the_same_text_after_the_window_does_not_block_a_read_back` | the content check still expires |
| `the_interrupt_flush_records_on_the_gesture` | the `Backspace`/`Esc` interrupt path records too |

Red/green: making the decision content-only again fails exactly the three fragmented cases (`49 passed; 3 failed`); restored, `client::tests` is 52 passed and `paste` is 60 passed.

```
cargo test --bin psmux client::tests   # 52 passed
cargo test --bin psmux paste           # 60 passed
```


psmux:
Merged at dc9ea31. Tracking the gesture instead of the text is the right shape: `blocks()` keeps the content comparison as its second half, so a clipboard payload that was never injected still goes through, and the explicit `Ctrl+Shift+V` and right click gestures only record.

I verified the red and green myself on the merged tree: with `blocks()` reduced to the content comparison, exactly `a_fragmented_paste_gesture_blocks_the_read_back_of_the_whole_text`, `a_fragmented_ascii_formula_blocks_the_read_back_too` and `a_gesture_that_forwarded_characters_blocks_the_read_back` fail (49 passed, 3 failed); restored, `client::tests` is 52 passed and `paste` 60.

One honest note for the record. I tried to reproduce the reporter's duplicate on the build before this PR (fabfd2d) two ways, and could not, on Windows 11 26200 with the current Windows Terminal:

- a real attached client in Windows Terminal, hardware `Ctrl+V` through `keybd_event` with a 90 ms hold, clipboard set to `恭喜通关`, `C2单元格应显示`, `=(B3-B2)/B2` and `ZQ7`: each landed once
- the same client driven at the console input buffer with the characters injected as key records and only the V release forwarded, at 30, 80, 150 and 300 ms between the characters and the release: each landed once

So on this host the fragmented paste was already arriving once before #676, which means the reporter's host splits the paste differently from mine. That does not weaken the fix, the unit cases pin the exact sequence you traced, but if you still have the reporter's Windows Terminal version and psmux `-V` line it would be worth noting on the issue for the next person.

What landed with the merge: `tests/paste_host_injector.cs` and `tests/test_pr676_paste_gesture.ps1`, which drive a real attached client through both routes above for the four strings and assert one landing per paste, a second deliberate paste landing too, and a plain `Ctrl+V` (a host that does not inject) still pasting once through the read back. 11 of 11 on the merged build, and the copy mode and paste suites are unchanged.

PR #678, opened 2026-09-19
copy mode: the yank matches the highlight in every direction and at every edge

tscodeplus:
# copy mode: the yank now matches the highlight, in every direction and at every edge

Fixes the reported "copied text has one character more than the highlighted
selection", including the left-to-right and multi-row variants, and the variant
where the last selected cell simply *looked* unselected.

Four separate defects were involved. Each one is fixed, pinned by tests, and was
reproduced on a real client: a psmux client driven with injected Windows console
mouse records, so the cells it actually painted could be read back with
`ReadConsoleOutputAttribute` and compared with the paste buffer.

## 1. The release must never move the endpoint (`c3ce75c`)

The release handlers assigned `copy_pos` from the release cell. A terminal
reports the last motion and the button-up together, so the release cell is often
one past the cell that was painted: the yank picked up a character the user
never saw highlighted.

Red/green on a real client (`E:\Code\bin`, drag to cell 6, release at 7):

| build | painted highlight | yank |
|---|---|---|
| before | `E:\Code` | `E:\Code\` (6) |
| after | `E:\Code` | `E:\Code` (5) |

## 2. The client re-reports the cell it painted (`45e0b01`)

Fixing (1) is not enough on its own. When the slip and the release arrive in the
*same* input burst the client has not painted the slip's frame yet, so the
server's newest drag is still a cell that never reached the screen. Only the
client knows what it painted, so a copy-mode drag release re-reports the
endpoint it last painted as a drag, immediately before the release. The server
keeps treating the newest drag as the endpoint and the release stays a no-op.

The raw `mouse-down/drag/up` protocol has no client that can speak, so
`remote_mouse_up` keeps dropping a same-batch slip against the endpoint of the
last frame.

A/B on a real client (`aaa ~\.psmux\client_debug bbb`, press on `~`, drag to
`u`, then a slip to `g` plus the release in one burst):

| build | painted highlight | yank |
|---|---|---|
| without the re-pin | cells 4..23 `~\.psmux\client_debu` | `~\.psmux\client_debug` (21) |
| with the re-pin | cells 4..23 `~\.psmux\client_debu` | `~\.psmux\client_debu` (20) |

## 3. Pair each column with its own row, and carry each end's scroll offset (`b957aa6`)

* The frame paired rows and columns independently (a min/max of each) while the
  yank pairs every column with the row it was dragged over, the way tmux's
  `screen_select_cell` does. A drag whose row direction differed from its column
  direction therefore painted one range and copied another.
* Only the anchor recorded the scroll offset it was captured at; the endpoint
  was resolved against the *current* offset. Dragging onto the first or last
  content row auto-scrolls the view, which moved the endpoint onto a different
  content line than the one that had been painted: extra lines at the bottom
  edge, an empty yank at the top edge.

Both ends now carry their own offset (`copy_pos_scroll_offset`) and the frame
pairs cells exactly the way the yank does.

## 4. Don't reverse the copy cursor cell while a selection is on screen (`4853ddd`)

This is the one the reporter could *see*, and the reason the copy looked wrong
even after (1)-(3). The copy-mode cursor was drawn as a `REVERSED` cell with the
host terminal's cursor parked on it. During a mouse drag the cursor sits on a
selection endpoint, so the last selected cell rendered as "text colour on the
default background" -- indistinguishable from an unselected cell. That is
exactly how it was reported: *"the last g is yellow text, not a yellow
background"*. The selection and the copy were both correct; the display was
lying.

Read back from a live client, the endpoint cell carried
`COMMON_LVB_REVERSE_VIDEO` (`0x4000`). The fix skips both the `REVERSED`
modifier and the cursor move when the cursor cell is inside the selection
(skipping `set_cursor_position` also keeps ratatui's `?25l`, so the host cursor
stays hidden for that frame and every selected cell keeps the selection style).
Keyboard copy mode has no selection and is unchanged: the cursor is still drawn
and reversed.

| build | selection cells | endpoint cell | yank |
|---|---|---|---|
| before | 4..24 | `0x4060` (reverse video) | `~\.psmux\client_debug` |
| after | 4..24 | selection style | `~\.psmux\client_debug` |

## Tests

`window_ops_tests` 24, `copy` 179, `mouse` 142, `select` 105, `paste` 60,
`client::` 276, `input::` 282, `layout::` 27, `copy_mode::` 84;
`cargo check --all-targets` clean.

New or updated:

* `a_release_after_a_client_repin_yanks_the_painted_cell`,
  `a_release_without_a_repin_follows_the_last_drag`,
  `a_legacy_mouse_up_slip_no_frame_carried_must_not_widen_the_yank`
* `a_direction_mismatched_drag_paints_and_yanks_the_same_range`, and
  `copy_drag_release_after_edge_scroll_yanks_and_exits` updated to the range the
  frame actually paints (which is the whole point of the fix)
* `copy_release_repin_tests` (5) pin the client's re-pin ordering: before the
  release, never on a click, nothing painted means no re-pin
* `copy_cursor_in_selection` (4) pin the cursor/selection overlap for char,
  rectangle and line modes

Note that (4) is a client-side change: clients need the new binary, not just the
server.


tscodeplus:
Closing this for now, please do not merge. The release-endpoint fix itself is correct and verified with a real client (injected Windows mouse records), but the reporter still sees a selection/copy mismatch on real hardware that I have not fully explained yet. I will reopen with a re-verified branch once the remaining case is understood.

psmux:
@tscodeplus understood, I will hold this one and not merge it until you say it is ready.

For what it is worth, the way you split it reads right to me: the release never moving the endpoint, the client reporting the cell it actually painted, each end carrying its own scroll offset, and the reversed cursor cell hiding the last selected cell. Reading the painted cells back with `ReadConsoleOutputAttribute` and comparing them with the paste buffer is exactly the kind of proof I like to see, so thank you for being honest that one real hardware case still does not fit rather than letting it land half explained.

Two things that may help with the remaining case, since the reporter sees it on real hardware and the injected records do not:

1. A real terminal can coalesce or reorder motion and button up differently from injected records, and Windows Terminal reports motion per cell while some hosts report per pixel move. If the reporter can run the client with `PSMUX_MOUSE_DEBUG=1` and send the log for one bad drag, the exact sequence of drag and release cells will show whether the endpoint or the paint is the side that is off.
2. Wide characters and a selection that ends on the second half of a double width cell are worth a look, given the CJK reports you have been fixing next door in #667 and #676.

Take the time it needs. Reopen or push here whenever it is ready and I will test it against tmux's `screen_select_cell` pairing on my side before it goes in.

tscodeplus:
Thanks - and that hold is on me, not on anything outstanding in the branch.

I closed this myself on 2026-09-20 01:57 UTC with "please do not merge", because at that
moment the reporter still saw a selection/copy mismatch on real hardware that I had not
explained. I found it 53 minutes later - it was a *rendering* artifact, not a yank bug: the
copy cursor cell was painted with COMMON_LVB_REVERSE_VIDEO, so the last selected cell looked
unselected while the buffer was already correct - pushed it, and reopened this PR at 02:50 UTC.
Your reply landed 33 h after that reopen, which is why it reads as if it were still on hold.
It is ready now.

Both of your suggestions were already covered; I have a fresh measurement for the second one.

1. The real-hardware case. I stopped relying on injected records alone: the harness drives a
   real console client (`FreeConsole`/`AttachConsole` + `WriteConsoleInputW` with MOUSE_EVENT
   records, started in its own console with `PSMUX_SESSION`/`TMUX` cleared) and then reads the
   cells that were actually painted with `ReadConsoleOutputAttribute`, comparing them with
   `show-buffer`. Before the last commit the endpoint cell came back `0x4060`
   (REVERSE_VIDEO); after it, painted range and buffer agree exactly. That A/B is also what
   killed my earlier idea of dropping one of the commits: with the slip and the release in the
   same input burst, the 3-commit build yanks 21 characters where the 4-commit build yanks the
   20 the frame actually paints.

2. Wide characters. Just re-ran the endpoint on a double-width line
   (`另外别忘了那 3.9G end`), ending the selection on the *second* half of a double-width cell:

   | drag end cell | release cell | selection_end_x | yanked | code points |
   |---|---|---|---|---|
   | 1 (2nd half of 另) | 1 | 1 | 另 | U+53E6 |
   | 2 (1st half of 外) | 2 | 2 | 另外 | U+53E6 U+5916 |
   | 3 (2nd half of 外) | 3 | 3 | 另外 | U+53E6 U+5916 |
   | 1 | 3 | 1 | 另 | U+53E6 |

   A double-width pair is never split, in either direction. The last row is the intended rule:
   the endpoint follows drag/motion events and never the release - the client re-pins the cell
   it actually painted just before the release, so a release the terminal coalesced cannot
   extend the selection.

No rush from my side. If you would rather land the three server-side fixes first, say so and I
will split (4) out into its own PR - it is the only client-side one, so it needs new client
binaries rather than just a server restart.

tscodeplus:
Rebase check on the current master, since #682 landed in copy mode after this branch was opened.

Cherry-picking the four commits onto `0bcc421` is clean, no conflicts (matching the
`mergeable_state=clean` GitHub reports), and the result passes everything I have:

* `tests/test_pr679_idle_poll_not_activity.ps1` - your E2E, including the focus-out cell: 5/5.
* copy-mode self-test, all five cases: 5/5. This includes the same-row slip that a plain
  master build still fails, with the original symptom: painted `[W5 AAAA]`, yanked
  `[W5 AAAA ]` - the "one character more than the highlight" report.
* rendering check: no selected cell comes back inverted.
* flicker byte check: the clients write nothing after the first second.

So the branch and #682 do not interact - page size following the pane height, the `C-v`
rectangle toggle and the numeric prefix repeats all touch the same files, and nothing
regressed. The check is `git cherry-pick c3ce75c 45e0b01 b957aa6 4853ddd` onto master, if
you want it in your own tree before merging.

Note for the record, since it caused confusion on my side earlier: `copy_pos_published`
and `copy_cursor_in_selection` are in this branch and not in master, so a build straight
from master drops the last two fixes. That is expected while this is open, not a sign that
the branch is stale.

psmux:
@tscodeplus merged, thank you, and thank you for how carefully this is put together. Merge commit 75bbd4c, with one follow up of mine on top that I want to explain, because it changes one line of yours.

I took your four commits onto `0bcc421` (clean, as you said) and went at them with a real attached client, real `MOUSE_EVENT` records written straight into the client's console input, and the frame the server published read back just before the button goes up, then `show-buffer` after the release. Seven cases, three runs each: left to right, right to left, upward multi row, downward multi row, ending on the last column, the slip (motion and release in one `WriteConsoleInput` batch), and a CJK line ending on the second half of a wide cell.

Three of the four commits land exactly as you describe. The one I most wanted to see is the slip: on master the client paints `PR678-L1-` and the buffer comes back `PR678-L1-A`, nine cells highlighted and ten copied, which is the reporter's "one character more than the highlight" word for word. On your branch it is nine and nine. The pairing fix is right too; I checked it against `screen_check_selection` in tmux's `screen.c`, which pairs `sx` with `sy` and `ex` with `ey` exactly the way your `top_col` / `bot_col` do. The edge auto scroll case behaves the way `window_copy_drag_update` plus `window_copy_cursor_up` do: the row under the pointer stays selected and the selection grows one content line per scroll. And you are right that tmux never reverses the copy cursor cell; there is no cursor special case anywhere on `screen_select_cell`'s path.

The one thing I had to change, and I am fairly sure it is a naming trap rather than a thinking one: `active_copy_sel_end` reads `sel_end_row` / `sel_end_col`, but after your own b957aa6 those are the painted rectangle's top and bottom ordered by content line, so on a drag that moved UP the bottom end is the anchor. The client then re pins the endpoint onto the press cell, the two ends become the same cell, and the release yanks nothing at all. An upward three row drag on my box painted 59 characters and copied 0, three runs out of three, where master copied the 59. I isolated it to the re pin alone by replaying the exact `pane-mouse` command the client sends: without it the buffer is 61 bytes, with it the buffer is empty.

The end you want is `copy_cursor_row` / `copy_cursor_col`, which is `copy_pos` itself and so is always the end the drag is moving. `sel_end_row` is still read, but only to tell a real selection apart from keyboard copy mode, which has a cursor and no selection and must not re pin. That is 5b36347, with two unit tests. With it all seven cases pass three runs out of three, including your slip case and the CJK line, and your own `copy_release_repin_tests` and `a_release_after_a_client_repin_yanks_the_painted_cell` stay green. tmux avoids this trap by naming: its `selection_end_x` / `selection_end_y` formats are `endselx` / `endsely`, the end being dragged, never the bottom of the rectangle.

Two small things I tidied in the same merge: a stray reformat that had joined two lines in `border_cell_touches_rect`, and an extra blank line in `yank_selection`.

Smaller things I noticed and left for a later PR, none blocking: the raw protocol's `copy_pos_published` adopts the current scroll offset rather than the one it was published at, so a raw client releasing right after an edge auto scroll can still resolve that row against the wrong line; `mark_active` clamps only the row for an end that is off screen, where `window_copy_adjust_selection` also pulls the column to 0 or to the last column, and tmux hides the selection outright when both ends are off screen on the same side; and `render_layout_json` still carries an inline copy of the predicate that `copy_cursor_in_selection` now expresses, which is the kind of pair that drifts. If you feel like picking any of those up, I would take them gladly.

The proof lives in `tests/test_pr678_selection_matches_highlight.ps1` with its `pr678_mouse_probe.cs`. One note on method: on Windows 11 26200 `ReadConsoleOutputAttribute` returns `0x0007` for every cell a VT sequence coloured, even the status line, under both the default host and a classic conhost, so I read the highlight from the published frame instead of the attribute plane. I am not disputing your `0x4060` reading, only recording that it does not reproduce here.

Thank you again for the measurements, the wide character table, and for saying plainly when something was still unexplained. That is exactly how I like to receive a change.

tanaeakihiko:
This modification appears to have caused a misalignment in the copy position specification within copy mode. I'm currently investigating the issue, so please wait a while for the 3.3.9 release.
P.S.) I've submitted PR #688 regarding this matter.

PR #679, opened 2026-09-20
server: a client's idle frame poll is not user activity (window-size latest flicker)

tscodeplus:
# server: a client's idle frame poll is not user activity

Fixes the pane flicker reported with two clients attached: a full-screen program
(pi) repainted continuously while it was idle, on the SSH/Termius client *and*
on the local one, and it stopped the moment the phone disconnected.

## Root cause

`server/connection.rs` pinged `ClientActivity` for **every** command line a
persistent client sent, with only bare pointer motion excluded (#604):

```rust
if !crate::client::is_bare_motion_cmd(&line) {
    let _ = tx.send(CtrlReq::ClientActivity(client_id));
}
```

An idle client polls `dump-state` once a second (`IDLE_FLOOR_MS` in
`client.rs`), so with two clients of different sizes attached each poll moved
`window-size latest` to the *other* client:

* `resize_window.rs::note_client_activity` — "window-size latest follows the
  client the user is actually using" — recomputes the geometry,
* `refresh_dynamic_window_sizes` returns a changed size,
* `resize_all_panes` resizes every pane, which reaches the pane's PTY,
* so the program inside repaints its whole screen and every attached client
  redraws.

With a single client its own poll left `latest_client_id` where it already was,
`note_client_activity` returned early and nothing resized — which is why the
flicker needed the second (SSH) client to be attached.

## Fix

`is_client_poll_cmd`, next to the existing #604 exclusion, and both activity
pings consult it. A poll is not user intent; typing, clicking, scrolling and a
real resize still move the window to that client.

## Measurements

Real pipe-mode clients (the SSH path, `PSMUX_PIPE_VT=1`), two different sizes,
nothing but their own one-second polls touching the server. A flicker is a
repaint, and a repaint is bytes on the terminal, so this is what the user sees:

| build | bytes written in 10s of idle polling | per second |
|---|---|---|
| before | 2716 (settle 762) | `411 81 259 81 297 43 81 81 43` — a repaint every second |
| after | 932 (settle 762) | `784` then **silent** |

The window geometry tells the same story, sampled once a second with the same
two clients:

| build | geometry |
|---|---|
| before | `42x35 80x23 42x35 80x23 42x35 80x23 …` — flips on every poll |
| after | `80x23` — stable |

And the behaviour the feature exists for is unchanged (two clients at
42x35 / 80x23): typing in the desktop client takes the window to `80x23`, typing
in the phone takes it back to `42x35`.

## Tests

`tests-rs/test_resize_window_compat.rs`:

* `the_frame_poll_is_not_user_activity` — `dump-state` is a poll;
* `real_user_requests_still_count_as_activity` — keys, clicks, drags, a real
  `client-size`, `copy-mode` and `refresh-client` are not.

`resize_window_compat` 22 passed, `client::` 267 passed, `input` 295 passed,
`cargo check --all-targets` clean.

Note: `server::option_catalog::...::every_catalog_default_matches_a_fresh_appstate`
fails on this machine before and after the change (`cursor-style`: catalog
`bar` vs fresh `AppState` `default`) — it is unrelated to this fix.

Server-side only, so it needs a server restart to take effect.


psmux:
@tscodeplus thank you for this one. Genuinely good catch, and the measurements in the description made it quick to verify.

I reproduced it on master before reading the diff, with two persistent clients of different sizes attached to one session under `window-size latest`, doing nothing but their own once a second `dump-state`. Sampling the geometry after each client's poll:

```
before (b053bb7)   120x40 60x20 120x40 60x20 120x40 60x20 ...
after  (merged)    60x20 60x20 60x20 60x20 ...
```

Before, 20 of 20 samples alternate, and the server answered every single poll with a full frame instead of NC, about 1.37 MB over 10 seconds of pure idle. After, 20 of 20 identical and about 0.36 MB over the same 10 seconds. Typing still hands the window over both ways. So your diagnosis is exactly right, including why one client alone could never show it.

On parity I went through the tmux tree to be sure this is not merely convenient. tmux moves `w->latest` only through `server_client_update_latest`, which has exactly three callers: re election when the latest client dies (`server-client.c:385`), the `out:` label of the key callback (`server-client.c:1654`), and `MSG_RESIZE` (`server-client.c:2590`). More to the point, a tmux client sends nothing at all on a timer. There is no evtimer anywhere in `client.c`, and after identify the only messages it ever sends are `MSG_RESIZE` on SIGWINCH, `MSG_EXITING`, `MSG_WAKEUP` and `MSG_UNLOCK`. tmux has no poll to exclude, so excluding ours is the only way to match it.

That reading turned up one more case while I was in there. tmux guards the key callback caller with `key != KEYC_FOCUS_OUT`, so focus in counts as activity and focus out is the single key it deliberately filters. psmux reports both as ordinary command lines, so `focus-out` was still pinging `ClientActivity` after your fix. Measured on your branch with the same two clients: typing in the 120x40 client sized the window 120x40, then a bare `focus-out` from the idle 60x20 client pulled it to 60x20 and resized every pane. The same repaint you removed from the poll, reached by alt tabbing away from the second terminal. I added `is_focus_loss_cmd` beside your `is_client_poll_cmd` and both pings consult it, in 5a95f79. Focus in still counts.

I also checked every other line a client puts on the wire without the user asking, in case there was a third. `client-size` is sent only when the size actually changes, which is tmux `MSG_RESIZE` and correct. The `prefix-end` repeat expiry and the copy mode drag auto scroll both run on timers but can only fire on the client that is already latest, so they cost nothing. The pipe mode size query every second goes to the terminal, not to the server. Your poll was the only real offender.

On the catalog test: it fails for you and passes here, and the difference is your shell. `cursor-style` and `cursor-blink` are not in `AppState` at all, they are read straight out of `PSMUX_CURSOR_STYLE` and `PSMUX_CURSOR_BLINK` (`server/options.rs:429`), so the test was measuring your machine rather than the catalog. With `PSMUX_CURSOR_STYLE=default` exported I get your exact failure on an otherwise clean tree:

```
1 of 87 catalog defaults disagree with a fresh AppState.
  cursor-style
      catalog default : "bar"
      fresh AppState  : "default"
```

That is a test isolation defect and nothing you caused. Fixed in 4407d99: the parity tests now take the shared env lock and read those two with the variables removed, and the three cursor config tests move off their file local mutex onto the same lock, since `set -g cursor-style` calls `env::set_var` and could poison the comparison from inside the binary too.

As landed: `resize_window_compat` 25 passed, `client::` 267 passed, `option_default_parity` 7 passed, `config_exhaustive` 324 passed, `connection` 60 passed. I added a PowerShell E2E guard in 5dde965, `tests/test_pr679_idle_poll_not_activity.ps1`, with three cells: geometry stability under two idle clients, the typing handover both ways so a fix that simply froze the geometry cannot pass, and the focus rule. It fails 2 of 5 on b053bb7 and passes 5 of 5 on the merge, rerun against the installed binary after merging.

Your commit 47f6842 is on master with your authorship, merged at ce51830. Thank you again, this was careful work and the write up saved me a lot of time.

